PR #20, opened 2017-05-27
Added OPTIONAL or REQUIRED remarks to certain fields

chhsia0:
Added OPTIONAL/REQUIRED remarks and fixed some description inconsistencies.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -383,7 +383,7 @@ message CreateVolumeRequest {
   CapacityRange capacity_range = 3;
 
   // The capabilities that the provisioned volume MUST have: the Plugin
-  // MUST provision a volume that could satisfy ANY ONE of the
+  // MUST provision a volume that could satisfy ALL of the
   // capabilities specified in this list. The Plugin MUST assume that
   // the CO MAY use the  provisioned volume later with ANY of the
   // capabilities specified in this list. This also enables the CO to do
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -603,7 +603,7 @@ message NodeID {
   // Information about a node in the form of key-value pairs. This
   // information is opaque to the CO. Given this information will be
   // passed around by the CO, it is RECOMMENDED that each Plugin keeps
-  // this information as small as possible.
+  // this information as small as possible. This field is REQUIRED.
```
jdef: I think I'm ok with this change. @jieyu ?

spec.md
```diff
@@ -677,12 +677,12 @@ message ValidateVolumeCapabilitiesRequest {
...
 
   // The information about the volume to check. This is a REQUIRED
-  // field. 
+  // field.
   VolumeInfo volume_info = 2;
 
   // The capabilities that the CO wants to check for the volume. This
   // call SHALL return “supported” only if all the volume capabilities
-  // specified below are supported.
+  // specified below are supported. This field is REQUIRED.
```
jdef: Without this change, is there value in calling this RPC? For example, if a CO wants to validate the existence of a volume, is it valid to invoke `ValidateVolumeCapabilities` with an empty capabilities list? If so, we should document that expectation. Otherwise mark this as `REQUIRED` as suggested here. @jieyu @saad-ali thoughts?

jdef:
p0 because it's blocking #21, which is blocking other things

chhsia0:
@jieyu Removed the style fixes. I'll create a separated PR for it.

jdef:
Thanks!

PR #21, opened 2017-05-27
Suggested error codes and recovery behaviors

chhsia0:
This PR depends on #20 .

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1317,7 +1317,7 @@ message Error {
       // information to detach the volume from the other node. Caller
       // SHOULD ensure the specified volume is not attached to any other
       // node before retrying with exponential back off.
-      VOLUME_ALREADY_ATTACHED = 6;
+      VOLUME_ALREADY_PUBLISHED = 6;
 
       // Indicates that a node corresponding to the specified `NodeID`
       // does not exist.
@@ -1339,6 +1339,15 @@ message Error {
       // information to detach the volume from one other node before
       // retrying with exponential backoff.
       MAX_ATTACHED_NODES = 8;
+
+      // Indicates that the specified `NodeID` is not allowed or
+      // understood by the Plugin, or the Plugin does not support the
+      // operation without a `NodeID`. More human-readable information
+      // MAY be provided in the `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
+      INVALID_NODE_ID = 9;
     }
     
     // Machine parsable error code.
@@ -1422,8 +1431,21 @@ message Error {
       // terminated or deleted before retrying.
       NODE_DOES_NOT_EXIST = 6;
 
+      // Indicates that the specified `NodeID` is not allowed or
+      // understood by the Plugin. More human-readable information MAY
+      // be provided in the `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
       INVALID_NODE_ID = 7;
+
       VOLUME_NOT_ATTACHED_TO_SPECIFIED_NODE = 8;
+
+      // Indicates that the Plugin does not support the operation
+      // without a `NodeID`.
+      //
+      // Recovery behavior: Caller MUST specify the `NodeID` before
+      // retrying.
       NODE_ID_REQUIRED = 9;
     }
     
@@ -1444,10 +1466,25 @@ message Error {
       // to match Plugin CSI version.
       UNKNOWN = 0;
 
+      // Indicates that a volume corresponding to the specified
+      // `VolumeInfo` does not exist.
+      //
+      // Recovery behavior: Caller SHOULD verify that the `VolumeInfo`
+      // is correct and that the volume is accessable and has not been
+      // deleted before retrying.
       VOLUME_DOES_NOT_EXIST = 1;
-      UNSUPPORTED_MOUNT_OPTION = 2;
+
+      UNSUPPORTED_MOUNT_FLAGS = 2;
       UNSUPPORTED_VOLUME_TYPE = 3;
       UNSUPPORTED_FS_TYPE = 4;
+
+      // Indicates that the specified `VolumeInfo` is not allowed or
+      // understood by the Plugin. More human-readable information MAY
+      // be provided in the `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the `VolumeInfo` before
+      // retrying.
+      INVALID_VOLUME_INFO = 5;
     }
     
     ValidateVolumeCapabilitiesErrorCode error_code = 1;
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1339,6 +1338,9 @@ message Error {
       // information to detach the volume from one other node before
       // retrying with exponential backoff.
       MAX_ATTACHED_NODES = 8;
+
+      INVALID_NODE_ID = 9;
```
jdef: Please suggest a description and recovery semantics

spec.md
```diff
@@ -676,9 +676,8 @@ message ValidateVolumeCapabilitiesRequest {
   // The API version assumed by the CO. This is a REQUIRED field.
   Version version = 1;
 
-  // The information about the volume to check. This is a REQUIRED
-  // field.
-  VolumeInfo volume_info = 2;
+  // The ID of the volume to check. This is a REQUIRED field.
+  VolumeID volume_id = 2;
```
jdef: I **think** I'm OK with this. /cc @jieyu @saad-ali for their input

spec.md
chhsia0: Done.

spec.md
jdef: on second thought, why just the ID here? one of the reasons that we return `VolumeInfo` in `ListVolumesResponse` is to facilitate stateless plugin backends. By only supplying the ID here instead of the Info, I think that actually hurts stateless support.

What's the reasoning behind using an ID here instead of Info?

chhsia0: It seems to me that the other fields in `VolumeInfo` except for `id` and `metadata` are not necessary to determine `VolumeCapability`. As of the stateless property, most of the other API functions ask for two parameters: `VolumeID` and `VolumeMetadata`, instead of a single `VolumeInfo`. For consistency, I suggest that we use two parameters: `VolumeID` and `VolumeMetadata` here, or we could change the other functions to accept a `VolumeInfo` parameter. Thoughts?

jdef: Yes, good catch. We should be consistent and use ID and metadata

jdef: Discussed with @jieyu: for now `VolumeInfo` lets a plugin check access mode, which is a requirement. Let's not make this change here for now.

chhsia0: Undone.

spec.md
chhsia0: Also suggested the recovery semantics for other error codes. The recovery semantics of the following errors are still missing since I'm not sure what behaviors are appropriate:

- ValidateVolumeCapabilitiesError
-- UNSUPPORTED_MOUNT_FLAGS
-- UNSUPPORTED_VOLUME_TYPE
-- UNSUPPORTED_FS_TYPE
- NodePublishVolumeError
-- UNSUPPORTED_MOUNT_FLAGS
-- UNSUPPORTED_VOLUME_TYPE
-- UNSUPPORTED_FS_TYPE
-- MOUNT_ERROR
- NodeUnpublishVolume
-- UNMOUNT_ERROR
- ProbeNodeError
-- BAD_PLUGIN_CONFIG
-- MISSING_REQUIRED_HOST_DEPENDENCY
- GetNodeIDError
-- BAD_PLUGIN_CONFIG
-- MISSING_REQUIRED_HOST_DEPENDENCY

jdef: thanks, I documented the hack of recovery documentation in #56

spec.md
```diff
@@ -1422,8 +1437,21 @@ message Error {
...
+      // be provided in the `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
       INVALID_NODE_ID = 7;
+
       VOLUME_NOT_ATTACHED_TO_SPECIFIED_NODE = 8;
+
+      // Indicates that the `NodeID` is not set to match that returned
+      // by `GetNodeID`.
```
jieyu: This is not quite precise. There is another case where the CO just want to unpublish the volume no matter which node the volume is attached to. I think this error code is for those plugins that do not support this feature? What will be the logic of an CO when receiving this error code? What if the node ID has been lost? Thoughts? @saad-ali @jdef

jdef: Ah, good catch. Correct, this error code is only for those plugins which actually require the node ID. I think we decided that if a plugin requires node ID but the CO lost it, then administrator intervention is required. Does that sound right?

chhsia0: Fixed.

spec.md
```diff
@@ -1339,6 +1339,21 @@ message Error {
...
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
+      INVALID_NODE_ID = 9;
+
+      // Indicates that the `NodeID` is not set to match that returned
+      // by `GetNodeID`.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
+      NODE_ID_REQUIRED = 10;
```
jdef: actually, why do we need `NODE_ID_REQUIRED` here? The recovery semantics are the same (as for INVALID_NODE_ID) and there's no special meaning for the `ControllerPublishVolumeRequest.NodeID` being unset. Can we remove this error code?

spec.md
```diff
@@ -1422,8 +1437,22 @@ message Error {
...
+      // understood by the Plugin. More human-readable information MAY
+      // be provided in the `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
       INVALID_NODE_ID = 7;
+
       VOLUME_NOT_ATTACHED_TO_SPECIFIED_NODE = 8;
+
+      // Indicates that the `NodeID` is not set to match that returned
```
jdef: * If the `NodeID` is set, but it doesn't match `GetNodeID` then `INVALID_NODE_ID` should be returned.
* If `NodeID` is NOT set but this plugin RPC requires it, then `NODE_ID_REQUIRED` should be returned. This is meaningful because an empty `NodeID` in this case carries with it specific intentions.

.. so the first sentence of this description should be simply: "The plugin does not support the operation without a `NodeID`".

spec.md
chhsia0: Hmm... make sense. I'll remove it and explain in the comments of `INVALID_NODE_ID` that it includes the case that the `NodeID` is missing.

spec.md
chhsia0: Done.

spec.md
```diff
@@ -1422,8 +1431,21 @@ message Error {
...
+      // Recovery behavior: Caller MUST fix the `NodeID` before
+      // retrying.
       INVALID_NODE_ID = 7;
+
       VOLUME_NOT_ATTACHED_TO_SPECIFIED_NODE = 8;
+
+      // Indicates that the Plugin does not support the operation
+      // without a `NodeID`.
+      //
+      // Recovery behavior: Caller MUST fix the `NodeID` before
```
jdef: nit: maybe "Caller MUST supply a `NodeID` when retrying." instead?

chhsia0: Done. I used the same wording as in `ControllerPublishVolumeError::VOLUME_ALREADY_PUBLISHED`.

PR #41, opened 2017-06-16
Updated build process

akutz:
This patch updates the build process so that the protobuf generation is handled via a Makefile instead of directly in the Travis-CI build script. This makes it easier to generate the protobuf locally. The Makefile includes "csi.mk", which contains the generation logic and can be reused by other projects.

Additionally the Travis-CI build script was updated to explicitly declare its expected package location within Go's package structure. This is useful for forks of this repository that set up their own Travis-CI builds.
    
Finally the Travis-CI build script has been updated to push the generated protobuf and associated Go source file to a Gist in order to make them available after a build.


jdef:
Thanks for the PR.
I'm probably missing something obvious, but where do I find the uploaded gist?

akutz:
Hi @jdef,

You can find it right [here](https://travis-ci.org/container-storage-interface/spec/builds/243789020#L242) in the Travis-CI log.

akutz:
Hi @jdef,

FYI, I changed the logic to `printf` instead of `sed`'s `p` operation since the latter isn't supported by `sed` on Darwin systems. It's the same result, but more portable. I also updated the package name to `csi` to make it consumable by others since `spec` is a poor package name for generated code. I'd be happy to take on a more active role in CSI's repo structure and build management.

akutz:
Hi @jdef,

You know, before this is merged I'm going to update it once more to have it insert the commit ID into the generated protobuf file in order to track from which `spec.md` the protobuf was generated.

akutz:
Hi @jdef,

The gist is now failing since Github API access is not configured for your Travis-CI builds. I've updated the `.travis.yaml` file to place the contents of the environment variable `GITHUB_API_TOKEN` into the file `$HOME/.gist`, the file the `gem` command uses to grab the OAuth token for authenticated Github API requests.

```yaml
after_success:
  - printf "%s" "$GITHUB_API_TOKEN" > "${HOME}/.gist"
  - gist -d "https://travis-ci.org/container-storage-interface/spec/jobs/${TRAVIS_BUILD_ID}" *.{proto,go}
```

If you, or someone, could create a secure Travis-CI environment variable named `GITHUB_API_TOKEN` and place an OAuth token into it then the build will always succeed in uploading the Gist. Thanks!

jdef:
Thanks. Want to try incorporating changes from 0eba188? If that doesn't work then I probably did something wrong setting up the envvar and/or the GH token

akutz:
Hi @jdef,

Instead of adding the secure env var directly to the `.travis.yml` file I was suggesting you add it as a secure env var via the Travis-CI's build settings. For example:

image

image

See what I mean?

akutz:
Hi @jdef,

I went ahead and rebased onto 0eba1880979af6c931ee17493b9fa342eec1c8a4, but I'd prefer to undo that if you would add the env var via the settings instead of directly to the travis config file.

jdef:
OK, I've added a new token via the travis UI and revoked the old token. You
should be able to undo the rebase and pick up the envvar via the UI setting

akutz:
Hi @jdef,

I updated the PR. Please note that the PR build [won't actually have access](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions) to that variable due, but as long as it is set up correctly it should work.

jdef:
LGTM

jdef:
Jie, Saad: would like one of you to sign off on this too

Issue #59, opened 2017-07-13
Proposal: GetCapacity MAY report capacity w/ respect to capabilities + parameters


jdef:
The current form of `GetCapacity` returns the total "raw" available capacity of storage backing the plugin. The capacity actually used by `CreateVolume` requests is expected to alter the available capacity of the backing storage platform, and it's also expected that such alteration is not necessarily 1:1.

For example, a storage platform might support several variations of RAID-style redundancy. The initial "raw" physical capacity may be reported as 1000GB. A `CreateVolume` request which includes a "raid1" parameter and allocates 100GB of RAID1 (mirrored) storage actually consumes 200GB of raw physical capacity. Likewise, a `CreateVolume` request could attempt to allocate 1000GB of raid1 storage and subsequently fail because the 2000GB if required storage aren't actually available. Without the CO understanding the specific semantics of "raid1" there no way for it to predict such a failure. And it's undesirable to require CO's to understand the impact of volume parameters - such parameters are opaque and beyond the scope of CO comprehension.

In order to allow CO's to more accurately estimate capacity changes w/ respect to a set of capabilities and parameters, I'd like to propose that `GetCapacityRequest` includes OPTIONAL `capabilities` and `volume_parameter` fields:

```protobuf
message GetCapacityRequest {
  // The API version assumed by the CO. This is a REQUIRED field.
  Version version = 1;

  // New fields:
  map<string,string> parameters = 2; // OPTIONAL
  VolumeCapabilities volume_capabilities = 3; // OPTIONAL
}

message GetCapacityResponse {
  message Result {
    // The total estimated "raw" capacity (available + used) of the storage pool from
    // which the controller provisions volumes. This is a REQUIRED
    // field.
    uint64 total_capacity = 1;

    // New field: `adjusted_capacity` MAY differ from `total_capacity` in the presence
    // of either `parameters` or `volume_capabilities` in the related GetCapacityRequest.
    // When neither `parameters` or `volume_capabilities` have been specified in the
    // request, then `adjusted_capacity` is expected to equal `total_capacity`.
    // This is a REQUIRED field.
    uint64 adjusted_capacity = 2;
  }
...
```

We could also introduce a `GetCapacityError` with a specific sub-codes for unsupported parameters, unsupported capabilities.

jdef:
fixed by #85

PR #61, opened 2017-07-19
CreateVolumeRequest: add requested AccessMode

jsafrane:
User should be able to specify (via CO) what kind of volume should be created.

Some applications running on CO may require a `SINGLE_NODE_WRITER` volume for a database, while others may want `MULTI_NODE_MULTI_WRITER` for say sharing of pictures. The Plugin should 1) see what the application needs  and 2) be able to raise an error if it's not able to provision a volume with requested access mode.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -391,11 +391,20 @@ message CreateVolumeRequest {
   // not supported by the Plugin, the call SHALL fail. This field is
   // REQUIRED.
   repeated VolumeCapability volume_capabilities = 4;
-    
+
   // Plugin specific parameters passed in as opaque key-value pairs.
   // This field is OPTIONAL. The Plugin is responsible for parsing and
   // validating these parameters. COs will treat these as opaque.
   map<string, string> parameters = 5;
+
+  // The requested minimal access mode. The Plugin MAY return a volume
+  // that allows a broader access mode, e.g. MULTI_NODE_MULTI_WRITER
+  // when CO asked just for SINGLE_NODE_WRITER. The Plugin MUST NOT
+  // return a volume with more restrictive access mode than the CO
+  // requested in this field, e.g. return SINGLE_NODE_READER_ONLY
+  // volume when CO requested SINGLE_NODE_WRITER.
+  // This field is OPTIONAL.
+  AccessMode access_mode = 6;
 }
 
 message CreateVolumeResponse {
@@ -1170,6 +1179,14 @@ message Error {
       // Recovery behavior: Caller MUST fix the invalid value before
       // retrying.
       INVALID_PARAMETER_VALUE = 7;
+
+      // Indicates that the access mode is not allowed by the Plugin.
+      // More human-readable information MAY be provided in the
+      // `error_description` field.
+      //
+      // Recovery behavior: Caller MUST fix the access mode before
+      // retrying.
+      UNSUPPORTED_ACCESS_MODE = 8;
     }
     
     // Machine parsable error code.


cpuguy83:
+1, this LGTM

jdef:
This came up a few times while we wrote the spec, but I forget the reason why we pushed access mode to the `*Publish` calls. /cc @saad-ali @thockin @jieyu do you remember the reasoning here?

jieyu:
Thanks for the PR! This PR made me think: should we move `AccessMode` to `VolumeCapability`?
 I was asking myself why `AccessMode` is any different than thing like `Block/Mount`? It looks like they are the same: all representing some capability of a volume. In `CreateVolumeRequest`, a CO can request a volume with certain capabilities, and the volume create MUST have those capabilities. `AccessMode` is just one of those capabilities.

For the sake of consistency, we should either add `VolumeCapability` to `VolumeInfo`, or remove `AccessMode` from VolumeInfo.

Something like the following:
```protobuf
message VolumeCapability {
  ...
  message AccessMode {
    ...
  }

  oneof value {
    BlockVolume block = 1; 
    MountVolume mount = 2; 
    AccessMode access_mode = 3;
  }
}

message VolumeInfo {
  uint64 capacity_bytes = 1;
  VolumeID id = 4;
  VolumeMetadata metadata = 5;
  repeated VolumeCapability volume_capabilities = 6;
}
```

Another nice thing about this proposal is that we now can get the `AccessMode` that a Plugin supports by calling `ControllerGetCapabilities`.

jsafrane:
I did not want to put it into `VolumeCapabilities` because it's not useful in `NodePublishVolumeRequest` - there already is `readonly` flag and eventual MULTI_NODE / SINGLE_NODE is IMO irrelevant here, it does not change the way how the plugin mounts the volume.

Also, what would be in `NodeGetCapabilitiesResponse` and `ControllerGetCapabilitiesResponse`? If a plugin supports SINGLE_NODE_WRITER, SINGLE_NODE_READER_ONLY and MULTI_NODE_READER_ONLY and ext2, ext3, ext4, xfs, btrfs and zfs, would it return 3 x 6 VolumeCapabilities? That does not seem to be useful.

jieyu:
@jsafrane good point on `AccessMode` is not necessary for `NodePublishVolumeRequest`. Currently, given that `VolumeCapability` is part of `NodePublishVolumeRequest`, according to your comment above, that means we should not have any volume capability that's not *node* related? That sounds a bit unfortunate.

> RE: would it return 3 x 6 VolumeCapabilities? That does not seem to be useful.

Why 3*6, not 3+6?

spec.md
```diff
@@ -391,11 +391,20 @@ message CreateVolumeRequest {
...
+
   // Plugin specific parameters passed in as opaque key-value pairs.
   // This field is OPTIONAL. The Plugin is responsible for parsing and
   // validating these parameters. COs will treat these as opaque.
   map<string, string> parameters = 5;
+
+  // The requested minimal access mode. The Plugin MAY return a volume
+  // that allows a broader access mode, e.g. MULTI_NODE_MULTI_WRITER
+  // when CO asked just for SINGLE_NODE_WRITER. The Plugin MUST NOT
+  // return a volume with more restrictive access mode than the CO
```
jieyu: We need to clearly define what "restrictive" means if we want to go with this route. If we define the order in AccessMode based on "restrictiveness", I don't think that forms a total order (it's a partial order meaning that there might not be an order between two enums).

jsafrane:
>> RE: would it return 3 x 6 VolumeCapabilities? That does not seem to be useful.

> Why 3*6, not 3+6?

My interpretation of `repeated NodeServiceCapability capabilities` is that all possible combinations of `fs_type` and `access_mode` should be returned. Most filesystems like ext4 support single mount read/write mount or multiple read only mounts with no writer (i.e. SINGLE_NODE_WRITER, SINGLE_NODE_READER_ONLY and MULTI_NODE_READER_ONLY), but there are some filesystems that allow multiple writers, e.g. gfs2. So it's not enough to return one VolumeCapability for each fs_type + one VolumeCapability for each:

```
[
  {fs_type: "ext4"},
  {fs_type: "xfs"},
  {fs_type: "gfs2"},
  {access_mode: "SINGLE_NODE_WRITER"},
  {access_mode: "SINGLE_NODE_READER_ONLY"},
  {access_mode: "MULTI_NODE_READER_ONLY"},
  {access_mode: "MULTI_NODE_SINGLE_WRITER"},
  {access_mode: "MULTI_NODE_MULTI_WRITER"}
]
```
How CO knows what fs supports which access mode? You must return the allowed combinations:

```
[
  {fs_type: "ext4", access_mode: "SINGLE_NODE_WRITER"},
  {fs_type: "ext4", access_mode: "SINGLE_NODE_READER_ONLY"},
  {fs_type: "ext4", access_mode: "MULTI_NODE_READER_ONLY"},
  {fs_type: "xfs", access_mode: "SINGLE_NODE_WRITER"},
  {fs_type: "xfs", access_mode: "SINGLE_NODE_READER_ONLY"},
  {fs_type: "xfs", access_mode: "MULTI_NODE_READER_ONLY"},
  {fs_type: "gfs2", access_mode: "SINGLE_NODE_WRITER"},
  {fs_type: "gfs2", access_mode: "SINGLE_NODE_READER_ONLY"},
  {fs_type: "gfs2", access_mode: "MULTI_NODE_READER_ONLY"},
  {fs_type: "gfs2", access_mode: "MULTI_NODE_SINGLE_WRITER"},
  {fs_type: "gfs2", access_mode: "MULTI_NODE_MULTI_WRITER"}
]
```

And that's too ugly I think.

jieyu:
@jsafrane I thought about it a bit more. I agree with you that asking the plugin to return a complete combination of `VolumeCapability` that it supports in `ControllerGetCapabilities` probably does not make sense and not scalable. There probably will be a lot of filesystems that a plugin supports. Not to mention the fact that one might consider building a meta plugin that supports multiple backends.

I think a more practical way is to let the CO drive the volume capability check. So here is my proposal:
1. Remove `volume_capability` in `ControllerServiceCapability`. For a plugin that manages multiple pools of storage space that have different volume capabilities, probably make more sense to integrate with `GetCapacity` (related to #59)
2. Make `AccessMode` be one of the `VolumeCapability`. Remove `oneof` for the fields in VolumeCapability. That means a `VolumeCapability` can represent a combination (e.g., `(MountVolume(xfs, noquota), "SINGLE_NODE_WRITER")`). Certain combination will be disallowed. For instance, either 'block' or 'mount' can be set, but not both.
```
message VolumeCapability {
  ...
  message AccessMode {
    ...
  }

  BlockVolume block = 1; 
  MountVolume mount = 2; 
  AccessMode access_mode = 3;
}
```
3. Remove `access_mode` in `VolumeInfo`. For pre-existing volumes, the CO can use `ValidateVolumeCapabilities` to check if a pre-existing volume has certain volume capabilities it intends to use.
```
message VolumeInfo {
  uint64 capacity_bytes = 1;
  VolumeID id = 4;
  VolumeMetadata metadata = 5;
}
```
4. Add `VolumeCapabilities` to `ControllerPublishVolumeRequest`. Similar to `NodePublishVolumeRequest`. This tells the Plugin about how the CO is about to use the volume and what volume capability that the CO expect. This should be one of the volume capabilities that the CO specifies in `CreateVolumeRequest` (or in `ValidateVolumeCapabilities` for pre-existing volumes).
```
message ControllerPublishVolumeRequest {
  ...
  VolumeCapability volume_capability = 6;
}
```
@jsafrane let me know if that makes sense to you or not. If you agree, let me know if you're interested in updating the spec. If not, I'm happy to prepare a PR for this.

cc @saad-ali @jdef @cpuguy83 @julian-hj

jsafrane:
Just thinking out loud...

> For pre-existing volumes, the CO can use ValidateVolumeCapabilities to check if a pre-existing volume has certain volume capabilities it intends to use.

I could imagine a GUI where CO asks user what kind of FS should be created on a volume. There should be a way how CO asks the plugin "give me list of everything you support" so the GUI provides a list of filesystems. This is not possible with `ValidateVolumeCapabilities`, where CO needs to know all filesystems in the world in advance and then check if the plugin supports each of them.

Is it a valid use case for CSI?

The rest of the proposal looks solid to me. Feel free to prepare a PR on my behalf, thanks in advance :-).

jdef:
Is the intent to **also** modify `NodeServiceCapability` by removing its `volume_capability` field?

Perhaps we amend Jie's point (1) such that supported filesystems (not filesystem/mount-option combos) can still be queried?
```
message ControllerServiceCapability {
  message RPC {
    enum Type {
      UNKNOWN = 0;
      ... // RPC types here
    }

    Type type = 1;
  }
  
  oneof type {
    // RPC that the controller supports.
    RPC rpc = 1;
    
    // File system type that the Plugin supports. Empty strings not allowed.
    string fs_type = 2;
  }
}
```

cpuguy83:
Does it make sense to move the mount/block fields into access mode at this point?
It still seems like it would need to be a `OneOf` for block vs mount.

jieyu:
@cpuguy83 this is pretty much what I did in the PR (except the naming). Take a look at #82

jieyu:
Close this one as #82 is merged.

PR #69, opened 2017-08-04
Add build support for CXX consumers

jieyu:
This patch chain adds the support for CXX consumers. It uses cmake as the configuration tooling. Currently, it only generates the headers and a static library.


CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.3)
+
+set(PACKAGE_NAME      "csi")
+set(PACKAGE_VERSION   "0.0.1")
```
chhsia0: Shouldn't the version be "0.1.0"?

CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+)
+
+add_custom_command(
+  OUTPUT    ${CSI_PROTO}
+  COMMAND   cat ${CSI_SPEC} | sed -n -e '/```protobuf$$/,/```$$/ p' | sed -e 's@^```.*$$@////////@g' > ${CSI_PROTO}
+  DEPENDS   ${CSI_SPEC}
+  COMMENT   "Generating protobuf file from the spec"
+)
+
+add_library(${PACKAGE_NAME} ${CSI_CXX_PROTO_SOURCES})
```
chhsia0: When the protobuf and grpc headers are in non-standard paths, the compilation of this library will fail. Let's add the following lines after Line 52:
```cmake
find_path(PROTOBUF_INCLUDE_DIR google/protobuf/message.h)
find_path(GRPC_INCLUDE_DIR grpc++/grpc++.h)
```
and the following after this line:
```cmake
target_include_directories(
  ${PACKAGE_NAME}
  PUBLIC ${PROTOBUF_INCLUDE_DIR}
  PUBLIC ${GRPC_INCLUDE_DIR}
)
```

CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+  ${CSI_CXX_PROTO_SOURCES}
+)
+
+#############################
+# Build rules.
+#############################
+
+set(CMAKE_CXX_STANDARD 11)
+
+find_program(PROTOBUF_COMPILER NAMES protoc)
```
chhsia0: `NAMES` not required here. Ditto below.

CMakeLists.txt
chhsia0: To use the headers with a non-standard path, we can then add `-DCMAKE_INCLUDE_PATH="<protobuf_path>;<grpc_path>;..."` when running `cmake`.

spec.md
```diff
@@ -228,6 +228,9 @@ There are three sets of RPCs:
 * **Node Service**: The Node Plugin MUST implement this sets of RPCs.
 
 ```protobuf
+syntax = "proto3";
```
chhsia0: These lines will make the original `make` command fail due to Line 138 in `csi.mk`. We probably want to either: 1. remove this and re-generate them in CMake; or 2. fix `csi.mk`; or 3. deprecate `csi.mk` and change the CI test.

CMakeLists.txt
chhsia0: On a second thought, let's not do this but ask the user to set up `CMAKE_CXX_FLAGS` instead.

CMakeLists.txt
chhsia0: How about doing `find_program` only if `PROTOBUF_COMPILER` is not set? Then the user can set up it directly instead of setting up `CMAKE_PROGRAM_PATH`. Ditto below.

README.md
```diff
@@ -2,50 +2,47 @@
...
-possible to generate the protobuf from a local spec file:
+In the above example, the protobuf file is generated by parsing `spec.md`. `csi.proto` will be generated in the build directory.
+
+### Generate Go Source
+
+The following command generates the Go source file from the protobuf:
 
 ```bash
-$ CSI_SPEC_FILE=spec.md make csi.proto
+$ mkdir build
```
jdef: if the point of including golang build support here is to encourage people to reuse these bits then ... we've failed. this is too much to ask of golang devs. a typical golang devs workflow is `go get github.com/org/project/package`.

this all feels too complex for what the original CI scripts were attempting to do: validate that the protos in the spec actually compiled to something. I somewhat regret accepting a PR that complicated the process. This PR adds additional prerequisites and process .. things that I don't think are actually appropriate for this repo.

I guess I could be convinced otherwise, but as of this moment I don't see the justification for making this repo more complex.

CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+set(CMAKE_CXX_STANDARD 11)
+
+# Find the protobuf compiler and plugins if not specified.
+if (NOT PROTOBUF_COMPILER_PATH)
+  find_program(PROTOBUF_COMPILER protoc)
+else()
+  set(PROTOBUF_COMPILER ${PROTOBUF_COMPILER_PATH})
+endif ()
+
+if (NOT GRPC_CXX_PLUGIN)
```
chhsia0: `if (NOT GRPC_CXX_PLUGIN_PATH)`

Also, do we want to document these variables in the README file?

CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+  find_program(GRPC_CXX_PLUGIN grpc_cpp_plugin)
+else()
+  set(GRPC_CXX_PLUGIN ${GRPC_CXX_PLUGIN_PATH})
+endif ()
+
+if (NOT GRPC_GO_PLUGIN)
```
chhsia0: `if (NOT GRPC_GO_PLUGIN_PATH)`

CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+# Find the protobuf compiler and plugins if not specified.
+if (NOT PROTOBUF_COMPILER_PATH)
+  find_program(PROTOBUF_COMPILER protoc)
+else()
+  set(PROTOBUF_COMPILER ${PROTOBUF_COMPILER_PATH})
+endif ()
+
+if (NOT GRPC_CXX_PLUGIN)
+  find_program(GRPC_CXX_PLUGIN grpc_cpp_plugin)
+else()
```
chhsia0: Let's follow the same style and put a space between `else` and `()`. Ditto below.

akutz:
This will need to be rebased onto the PR #74 regarding bindings. I concur re: documenting the vars in the readme. See what I did for the old csi.mk file. However, please use the lib/cxx/README.md file for this (again, please see PR #74).

jieyu:
Close this for now. I'll resend a PR later that rebase onto the current master.

PR #73, opened 2017-08-09
Simplified the build of the spec

jieyu:
The original build is unnecessarily complex. This patch simplifies the
build process by just generating the protobuf file.


.travis.yml
```diff
@@ -1,26 +1,18 @@
...
 install:
   - go get -u github.com/golang/protobuf/proto
   - go get -u github.com/golang/protobuf/protoc-gen-go
   - go get -u google.golang.org/grpc
-  - CSI_SPEC_FILE=spec.md make csi.proto
 
 script:
-  - make csi.pb.go
-
-after_success:
```
jdef: I really liked having the generated artifacts, can we keep these? (requires the earlier `gem install ...`)

akutz:
Hi @jieyu,

I don't see the value in this patch. The original build works fine and the external `csi.mk` file makes it easy for others to use that locally.

jieyu:
Did an offline chat with @akutz and @saad-ali. We will proceed with the following plan:
1. Make protobuf the source of truth
2. Reference the protobuf in the spec.md
3. Remove `libraries` repo in the csi org
4. Introduce a `lib/` folder in the spec repo
5. Putting language specific libraries (used for spec validation) in `lib/` folder. First step will be a minimal client based on generated grpc code and a mock server. It'll try to validate the workflow specified in the spec doc.

cc @saad-ali

Issue #68, opened 2017-08-03
golang code generation


julian-hj:
There's a very nice & useful makefile in the spec repo that will extract the protobuf snippets from `spec.md` and generate golang code from that protobuf, but we lack a good way to make that makefile work with standard go tooling.

In the normal golang world, I can write a plugin that imports `github.com/container-storage-interface/spec` and then I can `go get` that plugin, and go will fetch all the various code packages and build my plugin executable.  But that mechanism only works if the go code is actually available.  As it stands today, that `go get` would clone the spec repo and then fail to build because the code is not yet generated.  For that reason, by convention `go generate`d code is normally checked in.  (Checking in generated code seems kind of gross if you are coming to golang from just about any other programming language, but nonetheless, that seems to be the convention.)

In order to make CSI more go-friendly, I propose to have a CI task that watches for changes to the spec repo, and generates the protobuf and go code whenver there is a spec change.  We're making a version of that in our cloudfoundry pipelines, but we're happy to push the generated code back to a CSI repo if others agree that this is desirable.

jieyu:
@julian-hj I am actually in the process of refactor the make process for C++ as well. Let's collaborate on that.

akutz:
Hi @julian-hj,

A way to make that work with Golang tooling is adding this to the top of a Go source file:

```
//go:generate make csi/csi.go
```

You can read more on `go:generate` [here](https://blog.golang.org/generate).

I personally believe the spec should be a protobuf first and Markdown second, in part to make automated tooling easier. 

I also think hooks that cause additional commits are a bad idea. Please check the spec repo and note I already process out the protobuf and a generated Go source with every build and [push those](https://travis-ci.org/container-storage-interface/spec/builds/257853236#L309) to a Gist.

I've offered to provide additional assistance in the realm of CI to the CSI team. I think they're happy with the way things are ATM. I agree that it could be better.

akutz:
Hi @julian-hj,

By the way, I've migrated away from `make` a bit myself and am using `go generate` in collab with in-tree Go programs to do things like generate my version info. See [`semver.go`](https://github.com/akutz/rexray/blob/feature/simplified-build/core/semver/semver.go) and how I use it at the root of REX-Ray to generate `./core/core_generated.go` in concert with a Go template:

```
//go:generate go run semver/semver.go -f semver.tpl -o core_generated.go
//go:generate go run semver/semver.go -x -f env -o semver.env
```

[...]

PR #74, opened 2017-08-10
Version the CSI protobuf

akutz:
This patch versions the generated CSI protobuf and provides a Travis-CI build step that verifies the protobuf is up-to-date before allowing any changes to be merged.
    
Additionally this patch also adds minimal Go language bindings in order to verify the specification. Finally, scaffolding was put into place in order to support Cxx language bindings in the future as well.


jdef:
This patch also removes golang stub generation support. Other issues/PRs are trying to standardize proto stub generation: #68 #69 
At some point in the near future I want the CI to test that the .proto file actually compiles -- losing that capability as part of this patch is unfortunate.

akutz:
Hi @jdef,

You were invited but not able to make the call with @saad-ali and @jieyu and myself yesterday. There is a plan to handle your concerns, and in fact I even mentioned your exact concern in our group chat in Slack a few hours back:

image

@jieyu summarized yesterday's meeting [in this comment](https://github.com/container-storage-interface/spec/pull/73#issuecomment-321414292), but it looks like you were not tagged.

I hope this helps assuage any concerns you may have.

jdef:
@akutz thanks for the summary. Looking forward to the PR that adds back .proto validation. If it doesn't happen over the next couple of days then I'd like to track the need via an issue.

akutz:
Hi @jdef,

To be clear, it won't be validation but Go language binding generation. The validation @jieyu, @saad-ali, and I discussed includes using a bare-minimum Cxx and Go client to replicate the understood workflows against a mock plug-in, similar to CNI's methodology.

jdef:
That's fine. The lost capability that I'm looking to get back is validation
of the syntax of the .proto file, which we'll gain as soon as *any*
bindings are generated in CI from the .proto file

README.md
```diff
@@ -2,50 +2,5 @@
...
-| `CSI_GIT_OWNER` | The GitHub user or organization that owns the git repository that contains the CSI spec file | `container-storage-interface` |
-| `CSI_GIT_REPO` | The GitHub repository that contains the CSI spec file | `spec` |
-| `CSI_GIT_REF` | The git ref to use when getting the CSI spec file. This value can be a branch name, a tag, or a git commit ID | `master` |
-| `CSI_SPEC_NAME` | The name of the CSI spec markdown file | `spec.md` |
-| `CSI_SPEC_PATH` | The remote path of the CSI markdown file | |
-| `CSI_PROTO_NAME` | The name of the protobuf file to generate. This value should not include the file extension | `csi` |
-| `CSI_PROTO_DIR` | The path of the directory in which the protobuf and Go source files will be generated. If this directory does not exist it will be created. | `.` |
-| `CSI_PROTO_ADD` | A list of additional protobuf files used when building the Go source file | |
-| `CSI_IMPORT_PATH` | The package of the generated Go source | `csi` |
+This project contains the CSI [specification](spec.md) and
```
jieyu: Please follow the markdown style guid here:
https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#markdown-style

One sentence per line.

Makefile
```diff
@@ -1,8 +1,31 @@
...
+	diff "$@" "$?"
+else
+	cp -f "$?" "$@"
+endif
+
+build: $(CSI_PROTO)
 
 clean:
-	rm -f $(CSI_PROTO) $(CSI_GOSRC)
+	rm -f $(CSI_PROTO)
```
jieyu: Remove the tmp file as well?

.travis.yml
```diff
@@ -1,26 +1,11 @@
...
-  - CSI_SPEC_FILE=spec.md make csi.proto
-
-script:
-  - make csi.pb.go
-
-after_success:
-  - printf "%s" "$GITHUB_API_TOKEN" > "${HOME}/.gist"
-  - gist -d "https://travis-ci.org/container-storage-interface/spec/jobs/${TRAVIS_BUILD_ID}" *.{proto,go}
+# run make
+script: make
```
jieyu: For travis CI for now (until we have validation tooling checked in), can we try to call protoc to generate the go file to capture any protobuf syntax errors?

akutz: Hi @jieyu,

That's not necessary. The other PR is almost on deck.

Makefile
akutz: Hi @jieyu,

Good catch.

.travis.yml
jieyu: ok, but I do want to merge this first. The other might take a while to review. Sounds like not too much work here. Just call protoc in travis yaml?

README.md
akutz: Hi @jieyu,

It is a single sentence. Do you mean I should hard-wrap it? Sure, if that's this project's style. I prefer hard-wrapping for the sake of readability, but I will gladly adhere to the project's style guidelines.

.travis.yml
akutz: Hi @jieyu,

You can then review them in tandem if you wish or I'll put them into the same PR. Your call. I don't want to reintroduce a lot of what I already pulled out.

Makefile
```diff
@@ -1,8 +1,31 @@
 all: build
 
-include csi.mk
+CSI_SPEC := spec.md
+CSI_PROTO := csi.proto
 
-build: $(CSI_PROTO) $(CSI_GOSRC)
+# the target for building the temporary csi protobuf file.
```
akutz: Hi @jieyu,

As I was the first to create comments in the Makefiles for this project, is this your personal preference or part of the project's style guidelines? What you see is my preference, but I will change it in accordance with a posted style manual. Thanks!

akutz:
Hi @jdef,

FWIW it's already [in progress](https://github.com/akutz/csi-spec/tree/feature/lang-bindings-skeleton) with successful, [forked builds](https://travis-ci.org/akutz/csi-spec/builds/263262452). But the second PR will be dependent upon this PR. I will rebase the latter once this one is accepted.

README.md
jieyu: I mean this is a single sentence, you should not wrap it. see the rationale here:
https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#markdown-style

akutz: Fixed.

Makefile
akutz: Fixed.

.travis.yml
jieyu: Sounds good. Let's pull the other one into the same PR.

lib/go/Makefile
```diff
@@ -0,0 +1,68 @@
+all: build
+
+include ../../protoc.mk
+
+# If GOPATH isn't defined then set its default location.
+ifeq (,$(strip $(GOPATH)))
+GOPATH := $(HOME)/go
+else
+# If GOPATH is arleady set then update GOPATH to be its own
```
jdef: s/arleady/already/

akutz: Fixed

akutz:
Hi @jieyu,

Would you please mark this as approved? I've address all your suggested changes/fixes.

protoc.mk
```diff
@@ -0,0 +1,28 @@
...
+ifeq (Darwin,$(PROTOC_OS))
+PROTOC_OS := osx
+endif
+
+PROTOC_ARCH := $(shell uname -m)
+ifeq (i386,$(PROTOC_ARCH))
+PROTOC_ARCH := x86_32
+endif
+
+PROTOC := ./protoc
```
jieyu: I don't like the way we build protoc all the time. For C++ projects like Mesos/grpc, it might already bundle protoc and probably do not want to build protoc again. I'd suggest we allow downstream project to pass in a path to protoc (pre-existed at some location).

If you do need a `*.mk` file for modularization, please put that under `build/make` subdirectory. I might be interested in adding CMake support under `build/cmake`

protoc.mk
```diff
@@ -0,0 +1,28 @@
+# This Makefile makes it easy for the language bindings
+# to download and install the protoc binary.
+
+PROTOC_OS := $(shell uname -s)
+ifeq (Darwin,$(PROTOC_OS))
+PROTOC_OS := osx
+endif
+
+PROTOC_ARCH := $(shell uname -m)
```
jieyu: Instead of that, let's allow users to pass in the value directory. `uname` does not work on windows.

Makefile
```diff
@@ -1,8 +1,47 @@
...
+# is compared with the real file, and if they differ the build
+# will fail.
+#
+# Locally the temp file is simply copied over the real file.
+$(CSI_PROTO): $(CSI_PROTO).tmp
+ifeq (true,$(TRAVIS))
+	diff "$@" "$?"
+else
+	diff "$@" "$?" > /dev/null 2>&1 || cp -f "$?" "$@"
+.PHONY: $(CSI_PROTO).tmp
```
jieyu: Can you explain why this is a PHONY target?

Makefile
```diff
@@ -1,8 +1,47 @@
...
+endif
+
+build: $(CSI_PROTO)
+
+# If this is not running on Travis-CI then for sake of convenience
+# go ahead and update the language bindings as well.
+ifneq (true,$(TRAVIS))
```
jieyu: Why not build library as well in CI? We need to capture library build error on CI as well, right?

lib/go/Makefile
```diff
@@ -0,0 +1,68 @@
...
+PROTOC_GO_PKGS += github.com/golang/protobuf/proto
+
+# The following recipe downloads the go packages necessary
+# to generate the go language bindings.
+PROTOC_GO_PKGS := $(addprefix $(GOPATH)/src/,$(PROTOC_GO_PKGS))
+$(PROTOC_GO_PKGS):
+	go get $(@:$(GOPATH)/src/%=%)
+
+PROTOC_GO := $(PROTOC_GO_PKGS) $(PROTOC_GEN_GO)
+protoc-go: $(PROTOC_GO)
```
jieyu: this is a PHONY target?

lib/go/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	$(PROTOC) -I "$(<D)" --go_out=plugins=grpc:"$(@D)" "$<"
+
+# The temp language bindings are compared to the ones that are
+# versioned. If they are different then it means the language
+# bindings were not updated prior to being committed.
+$(CSI_GO): $(CSI_GO_TMP) | $(PROTOC_GO)
+ifeq (true,$(TRAVIS))
+	diff "$@" "$?"
+else
+	diff "$@" "$?" > /dev/null 2>&1 || cp -f "$?" "$@"
```
jieyu: So this output will depends on the version of protoc as well as the plugins i think? I am not sure if there will be any compatibility issue or not. For instance, what if a downstream wants to use a specific version of protoc because of some other dependencies?

lib/go/Makefile
```diff
@@ -0,0 +1,68 @@
...
+.PHONY: $(CSI_GO_TMP)
+endif
+
+build: $(CSI_GO)
+
+clean:
+	go clean -i -v
+
+clobber:
+	rm -f $(CSI_GO) $(CSI_GO_TMP)
```
jieyu: rm `.build` directory as well?

protoc.mk
akutz: Hi @jieyu,

CSI does not support Windows today, so it is irrelevant.

lib/go/Makefile
akutz: Hi @jieyu,

Yes and no. It's an alias.

protoc.mk
akutz: Hi @jieyu,

I do like it. It ensures a specific version.

Makefile
akutz: Hi @jieyu,

Because it ensures that a local, stale temp file won't cause the protobuf from not being rebuilt. We discussed this yesterday on Slack.

Makefile
akutz: Hi @jieyu,

Because I don't want the default target in the root Makefile to do recursive Make.

akutz:
Hi @jieyu,

This patch *was* originally separate commits, but I collapsed them at your request. In fact both you and @jdef objected to losing the Go code generation. I do not think it is prudent to divide them up again.

lib/go/Makefile
akutz: Hi @jieyu,

Then they can use that version. Please note that the version variable allows itself to be specified externally.

> `PROTOC_VER ?= 3.3.0`

lib/go/Makefile
akutz: Hi @jieyu,

That's a good catch, thanks.

jieyu:
@akutz The merge was meant for the makefile refactor. Merging top level makefile change with golang binding changes. I am fine with that. But the commit style change and whitespace fix should be in separate commits.

jdef:
> This patch was originally separate commits...

@akutz I think Jie would like all the changes to remain as part of this PR, but that this PR itself should be broken up into logically distinct commits (that won't be squashed at merge time). @jieyu correct me if I'm wrong

akutz:
Hi @jdef and @jieyu,

Guys, I am not going to spend time breaking this back up. Please feel free to do so. I wouldn't have even touched the spec.md file if not for adding two lines to the spec in order to specify the package. I did this because @jieyu did the same in a now-closed PR. I thought it was a good idea. I cannot edit that file without it updating the WS. I cleared this with @jdef and @saad-ali. @jdef, please feel free to add that text via your editor and I'll reset my copy of spec.md. 

Just... *sigh* Yeah. :)

akutz:
And @jieyu, the merge was meant for a use case, not a single file refactor. I get that you're being careful with the repo, but please don't forget there is a purpose to provide a feature/use case here, not simply the refactor of a single file.

jieyu:
@akutz at the very least, the style guide change should be a separate commit. Make sense?

akutz:
@jieyu, why? What purpose does that serve? The update to the contribution file regarding proper procedure for updating the spec is irrelevant without the rest of the changes. Commits should be atomic features, and the remark in the contribution doc is pointless on its own.

akutz:
Hi @jieyu,

Also, you keep referring to `CONTRIBUTING.md` as a style guide when it's far more than just that. You should break it up if you want to keep process separate from style. I recommend you look at http://rexray.readthedocs.io/en/stable/ and how we provide separate documentation for style, process, and build reference.

akutz:
Hi @jdef,

Could you please file a PR where you add the following to `spec.md`?

> ## Container Storage Interface
> 
> This section describes the interface between COs and Plugins.
> 
> ```protobuf
> syntax = "proto3";
> package csi;
> ```

It's around line 212. That way I can reset `spec.md` in this PR once I rebase off of your merged commit. It will remove all the whitespace changes. Thanks!

protoc.mk
akutz: I only made `proto.mk` external from the Go process in the case that you might want to re-use it. If that's not the case I'll move it into `lib/go/Makefile` directly. Thank you for the feedback.

jdef:
https://github.com/container-storage-interface/spec/pull/78

saad-ali:
We need to ensure that this community is fosters an environment of collaboration while maintaining a high quality bar. I know we all have good intentions, and we'll sort this out. To that end, I sent out an invite for a meeting @ 1:30 PM PT today to see if we can resolve the open questions around this PR. Message me on slack (saad-ali) if you want an invite.

akutz:
Hi @jdef,

It is going to, and is, fail(ing) because the existing build process places that text at the top of the file. You'd have to modify the existing code generation process.

jdef:
thanks @akutz, fixed

akutz:
Hi @jieyu,

I removed `spec.md` from the PR after rebasing off of `master` with @jdef's merged changes.

lib/go/Makefile
akutz: As for the plug-ins, I'm thinking I will vendor the Go pkg dependencies in `lib/go` and build the plug-in out of that to ensure a specific version. I'll use Glide to record to actual dependency manifest. Good call.

saad-ali:
We had a productive meeting. Follow up items we agreed on:
* Add comment to about what PHONY target is
* Add a README to root of lib/
* We are ok using `go get` for now, and not vendoring go dependencies. 
* Will leave make file for C as a hook for travis, and Jie can modify it in follow up PR.
* `uname` --> switch to `go version`
* proto buf should be major version (can be a follow up, @jdef)

akutz:
Hi @saad-ali,

One slight change - I'm sticking with `uname` as CSI does not support Windows, so it doesn't matter if `uname` isn't a Windows command. Additionally, @jieyu may have honed in on that one command, but the fact is the Makefile uses dozens more GNU commands unavailable on Windows as well.

In other words -- supporting Windows is more than just replacing the use of the `uname` command.

jieyu:
`uname` for now sounds good. Consider adding a TODO there.

akutz:
Hi @jieyu,

Thank you for your suggestion, but I'm going to leave it as is. FWIW, I had actually made the change to `go env` and decided to revert it. I understand you zeroed in on `uname`, but the fact is it's a larger issue. Go look at how `protoc` is downloaded or any other number of commands. Supporting Windows is a much broader task than just one command.

Makefile
jieyu: Can you remove the PHONY target here for now as we discussed offline?

PR #76, opened 2017-08-10
Add style guide for comments

jieyu:
Require that all comments to be full sentences.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,13 @@ This also applies to the code snippets in the markdown files.
 
 * Please wrap the code at 72 characters.
 
+## Comments
+
+This also applies to the code snippets in the markdown files.
+
+* End each sentence within a comment with a punctuation mark (please note that we generally prefer periods); this applies to incomplete sentences as well.
+* For trailing comments, leave one space between the end of the code and the beginning of the comment.
+
 ## Git commit
 
 ### Commit Style


CONTRIBUTING.md
```diff
@@ -15,6 +15,13 @@ This also applies to the code snippets in the markdown files.
 
 * Please wrap the code at 72 characters.
 
+## Comments
```
akutz: Hi @jieyu,

I think the style of comments in Makefiles vs Source files vs etc. can be different. As long as the intent is understood and a basic style is adhered (line length, wrapping, etc), I would relax a little.

akutz:
Hi @jieyu,

It's good practice to always include a useful description when filing a pull request. If none is available I recommend using the body of the PR's commit since [it is good practice](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message) for all Git commits to include a subject and body.

CONTRIBUTING.md
jieyu: @akutz Comments are for human to read, thus making it a real sentence makes sense to me. I prefer we have consistency among all types of files.

akutz: > for human 

_for humans_

CONTRIBUTING.md
```diff
@@ -15,6 +15,13 @@ This also applies to the code snippets in the markdown files.
...
+
+This also applies to the code snippets in the markdown files.
+
+* End each sentence within a comment with a punctuation mark (please note that we generally prefer periods); this applies to incomplete sentences as well.
+* For trailing comments, leave one space.
```
jdef: > leave one space

... where? between the end of the code and the beginning of the comment? if so .. we should say that

jieyu: fixed.

PR #79, opened 2017-08-11
Remove Lang Binding Builds from Root Makefile

akutz:
This patch removes part of the root Makefile that would call the language-binding builds when running `make` locally from the root of the project. If the `lib/go` build was executed outside of the `GOPATH` errors would occur. Since the repository is not Go-specific, running a Go-specific build by default should not be assumed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,15 +30,16 @@ endif
 
 build: $(CSI_PROTO)
 
-# If this is not running on Travis-CI then for sake of convenience
-# go ahead and update the language bindings as well.
-ifneq (true,$(TRAVIS))
-build:
-	$(MAKE) -C lib/go
-	$(MAKE) -C lib/cxx
-endif
+# the LANGS variable defines the languages for which bindings
+# are generated and used to validate the specification
+LANGS := $(notdir $(wildcard lib/*))
+$(addprefix build-,$(LANGS)):
+	$(MAKE) -C lib/$(@:build-%=%)
+$(addprefix clean-,$(LANGS)):
+	$(MAKE) -C lib/$(@:clean-%=%) clean
+.PHONY: $(foreach l,$(LANGS),build-$l clean-$l)
 
-clean:
+clean: $(addprefix clean-,$(LANGS))
 
 clobber: clean
 	rm -f $(CSI_PROTO) $(CSI_PROTO).tmp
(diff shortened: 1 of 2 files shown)


jieyu:
Can we have targets in the top level makefile for language bindings, but not put those targets in the default `all` target?
```
all: build

cxx:
    $(MAKE) -C lib/cxx

go:
    $(MAKE) -C lib/go
```

akutz:
Hi @jieyu,

We could, and I in fact [proposed](https://kubernetes.slack.com/archives/G6LFKSX44/p1502491700856190) this to you [multiple](https://kubernetes.slack.com/archives/D6LKLMFB7/p1502476093954430) times in Slack last week. You never responded.

```make
# the LANGS variable defines the languages for which bindings
# are generated and used to validate the specification
LANGS := $(notdir $(wildcard lib/*))
$(addprefix build-,$(LANGS)):
    $(MAKE) -C lib/$(@:build-%=%)
$(addprefix clean-,$(LANGS)):
    $(MAKE) -C lib/$(@:clean-%=%) clean
.PHONY: $(foreach l,$(LANGS),build-$l clean-$l)
```

jieyu:
@akutz Sorry about that. I think that sounds good! Can you update this PR?

akutz:
Hi @jieyu,

I could, but I'm not sure I should. Please see the current Slack chat between you, myself, @jdef , and @saad-ali. I think we *must* update all bindings for now. Until we come up with a better solution that is.

akutz:
Hi @jieyu,

Here's the conversation to which I'm referring:

image

Makefile
```diff
@@ -30,15 +30,16 @@ endif
...
-# If this is not running on Travis-CI then for sake of convenience
-# go ahead and update the language bindings as well.
-ifneq (true,$(TRAVIS))
-build:
-	$(MAKE) -C lib/go
-	$(MAKE) -C lib/cxx
-endif
-
-clean:
+# the LANGS variable defines the languages for which bindings
```
jieyu: Please make it a full sentence.

jieyu:
Second thought on this, I think it probably makes sense to always call make for bindings as well. If a PR modifies the spec protobuf but not the binding, then the PR should probably not be submitted. It does not make sense to ask the someone else to update the bindings for you.

I am actually in favor of always including building of the binding to `all` target in the root Makefile. Maybe still keep this PR (because it build for each language which is nice), but adding `build-xxx` to `build` target in the root Makefile.

jdef:
Looks like @jieyu requested changes here. Are we still interested in this PR?

saad-ali:

saad-ali:

PR #85, opened 2017-08-17
spec: Add volume capabilities and parameters to GetCapacityRequest

jieyu:
This change allows the CO to get the capacity of a storage pool with the
given volume capabilities and parameters. This is particularly useful
when the CO defines a storage class or profile, and wants to know the
capacity of the storage under that class or profile so that it can
perform allocation or track utilization.

For example, assume that a Plugin can provision both raid0 and raid1
volumes. Depending on the type, the capacity is different. This patch
allows the CO to query the capacity of the storage pool for each type:

```
request:
  version:
    major: 0
    minor: 1
    patch: 0
  volume_capabilities:
    - mount:
      fs_type: xfs
  parameters:
    type: raid1
response:
  result:
    total_capacity: 100000
```

```
request:
  version:
    major: 0
    minor: 1
    patch: 0
  volume_capabilities:
    - mount:
      fs_type: xfs
  parameters:
    type: raid0
response:
  result:
    total_capacity: 200000
```

This patch addresses #59.


codenrhoden:
I'm confused by both this and #59. The spec calls out the value of `total_capacity` as "Available + Used". That makes sense with the title of "total" capacity -- but in my view it's a number that never changes (excepting expansion or contraction of storage).  If you have a 1TB pool, the total capacity is 1TB whether the used portion is 1GB or 100GB.

In other words, I don't think this change does anything helpful unless you do the other part of @jdef proposal to add an `adjusted_capacity` param to the response.  Even then, I think that's ambiguous. What you really want is `available_capacity`, and having that response be adjusted based on volume capabilities makes total sense.

If you are looking at enhancing this response, it may be most useful just to report all three values: (1) total raw capacity, (2) total used, (3) available, and have available be optionally adjusted based on volume capabilities. Capacity planning in these scenarios is tricky, especially once thin-provisioning or over subscription is introduced. Just because a certain amount of storage (X) is available doesn't mean you could create a volume of size (X) and consume it all.

Regardless, I do hope to see some improvement here. `total_capacity` as defined today doesn't seem to be helpful to anyone. Knowing that you have a pool of a certain size doesn't help you make any decisions unless you know how much of that pool is available.  Am I interpreting `total_capacity` wrong?

csi.proto
```diff
@@ -456,6 +456,17 @@ message ListVolumesResponse {
 message GetCapacityRequest {
   // The API version assumed by the CO. This is a REQUIRED field.
   Version version = 1;
+
+  // If specified, the Plugin SHOULD only report the capacity of the
```
gpaul: Why is this SHOULD and not MUST?

gpaul: Answering my own question: some plugins may have no way of determining capacity. Anything else?

jdef:
@codenrhoden the underlying use case is this: a plugin may interface with a storage backend that may provision different kinds of logical volumes. I think the answer to your questions can be broken into a couple of pieces.

First, there's an assumption that a CO either knows about the volumes it has **already** created, or that it can recover that information via ListVolumes. In this case `total_capacity` is sufficient because the CO already "knows" *used-capacity* and so it may calculate *available-capacity*. Is this assumption flawed?

Second, when a new volume is created it may actually consume more physical capacity than the logical capacity requested. Logical volumes backed by storage implementations that MAY leverage RAID or erasure coding can consume more physical capacity than the logical capacity they expose to an application/container. The goal of this API change is to support storage backends that support a variety of logical volume types, and to allow the CO to reason about how much *logical* capacity is remaining given some set of volume capabilities and/or parameters, without forcing the CO to understand things like RAID or erasure coding algebra.

For example, let's say that you have 1TB of total physical storage backing a plugin instance. The plugin lets you create logical volumes in JBOD or RAID1 mode. When `GetCapacity` is invoked with `parameters` that indicate JBOD mode, the `total_capacity` is 1TB. When it's invoked with `parameters` that indicate RAID1 mode, the `total_capacity` is 500GB.

I don't think I should have included `adjusted_capacity` in my original proposal. If anything, including an additional `available_capacity` field makes more sense because it mitigates the (possibly broken?) assumption that the CO can accurately track *used-capacity* on the basis of logical volume type.

csi.proto
jdef: If a plugin can't determine capacity, then it should say so via controller capabilities, right?

gpaul: Good point - I'm back to "why SHOULD?" now.

codenrhoden:
@jdef Thank you so much for the detailed reply. It helps me understand the intent here. It took me a while to think about why this still seems not quite right to me, and I think I finally have it.

> First, there's an assumption that a CO either knows about the volumes it has already created, or that it can recover that information via ListVolumes. In this case total_capacity is sufficient because the CO already "knows" used-capacity and so it may calculate available-capacity. Is this assumption flawed?

There are two things I see here. The first one is just being pedantic. Both `GET_CAPACITY` and `LIST_VOLUMES` are optional RPCs, gated by Capabilities. What if `GET_CAPACITY` is supported, but not `LIST_VOLUMES`? It's an unlikely permutation, but valid per the spec. In that scenario, a CO cannot repopulate it's list of volumes to subtract from `total_capacity` to arrive at an `available_capacity`.

The second thing is a little more clear if I build off of your example:

> For example, let's say that you have 1TB of total physical storage backing a plugin instance. The plugin lets you create logical volumes in JBOD or RAID1 mode. When GetCapacity is invoked with parameters that indicate JBOD mode, the total_capacity is 1TB. When it's invoked with parameters that indicate RAID1 mode, the total_capacity is 500GB.

Mutating `total_capacity` based on parameters seems wrong to me, as it assumes that all **used** space is coming from volumes with the same parameters. Using JBOD and RAID1 as an example, let's pretend our 1TB of storage has the following volumes:


| Vol ID | Vol Type | Logical Space | Physical Space |
| - | - | - | - |
| A | JBOD | 100G | 100G |
| B | RAID1 | 100G | 200G |

In your scenario, a `GetCapacity` without parameters should return **1TB**, and the CO would know (by internal tracking or `ListVolumes`) that it has **200GB** of volumes, leaving **800GB** leftover. We know this is incorrect, so we call `GetCapacity` with parameters instead. But which Volume Type to use? If you specify JBOD, you'll get 1TB back (same as w/o parameters), and if you specify RAID1, you'll get 500GB back (just like in your example). But, because we have a mix of volumes, once you subtract out the logical volumes that exist (200GB worth), you'll think there is 300GB of available space for RAID1 volumes, when in fact there is 350GB.

The CO can't properly account for a mixture of volume types that consume different ratios of physical space, and I don't think anyone wants to push that logic upon the CO either.

I'd much rather see an SP return something akin to the following:
```
raw_capacity: 1TB
raw_available: 700GB
adjusted_available: 350GB
```

In that example, the `adjusted_available` would only be different than `raw_available` if parameters were provided that allowed the SP to determine a physical storage ratio that is different than the logical one. I'm sure that does not apply to all storage platforms.

Anyways, I hope my example makes it clear why applying an adjustment to `total_capacity` likely does not make sense in a mixed volume environment.

codenrhoden:
@jdef One more reason I think having an SP report available capacity is important. Some storage systems can be configured in such a way that there is a global aggregation of storage, but an SP may only have insight into volumes created a in a partition of that storage, even though all partitions consume from the aggregate...

For a concrete example, a Ceph cluster can have multiple pools. Perhaps one pool is used for hosting OpenStack VM disks, and another for hosting k8s volumes. If you query the cluster for capacity, you actually get the global amount of storage. Consumers in other pools (say, OpenStack) take away from available storage in your pool -- so just taking the total capacity and subtracting the logical capacity of volumes in your pool may reflect an inaccurate amount of available storage.

An SP may be able to take this into account, by querying other pools' usage (even though it can't see the volumes or act on them), but a CO cannot do that, as it can only list volumes from the configured pool.

jieyu:
@codenrhoden Thanks for the detailed example and the feedback! In fact, we had the same thought exercise initially. 

> In your scenario, a GetCapacity without parameters should return 1TB, and the CO would know (by internal tracking or ListVolumes) that it has 200GB of volumes, leaving 800GB leftover. We know this is incorrect, so we call GetCapacity with parameters instead. But which Volume Type to use? If you specify JBOD, you'll get 1TB back (same as w/o parameters), and if you specify RAID1, you'll get 500GB back (just like in your example). But, because we have a mix of volumes, once you subtract out the logical volumes that exist (200GB worth), you'll think there is 300GB of available space for RAID1 volumes, when in fact there is 350GB.

The thinking is: the CO will only call `GetCapacity` for given storage profile (volume capability + parameters combo). Raw capacity is most likely not gonna be useful for the CO in terms of allocation because the knowledge about how much raw space is going to be used by a volume is only known by the SP. Surface that information to the CO and the end user probably does not make sense.

In this example, assuming the underlying physical disk capacity is 1TB and there are 2 existing volumes like you listed above. By using `GetCapacity` RPC, the CO knows that the total capacity for JBOD profile volumes is 800GB (and the available is 700G), and the total capacity for RAID1 profile volumes is 450G (and the available is 350G).

Now I think this more as you mentioned it, I think maybe returning `available` will make the semantics more clear. It means the `available` storage space for creating volumes with the capabilities and parameters specified in the request.

csi.proto
jieyu: Fixed. Used SHALL instead.

codenrhoden:
@jieyu Thanks for the history. I hope I am not just dragging up old problems that the WG has long solved. I'm not purposely trying to be dense, but I still don't quite get it. At a high-level, it seems like the CO is having to do an awful lot of work if it needs to track space for every permutation of storage profile.  But, to address some specifics from your comment...

> Raw capacity is most likely not gonna be useful for the CO in terms of allocation because the knowledge about how much raw space is going to be used by a volume is only known by the SP. Surface that information to the CO and the end user probably does not make sense.

I completely agree with you here. That's why I find the current definition of `total_capacity` to be not useful: "The total capacity (available + used) of the storage pool". As long as it contains used space, it's a constant number IMO.

> In this example, assuming the underlying physical disk capacity is 1TB and there are 2 existing volumes like you listed above. By using GetCapacity RPC, the CO knows that the total capacity for JBOD profile volumes is 800GB (and the available is 700G), ...

Okay, to get here you are calling `GetCapacity` with the JBOD profile. Since `total_capacity` is defined as **available + used** this will always return 1TB, not 800GB. From your latter comments, I do believe that this is going to be changed to just be `available_capacity` instead, so from here on out I will assume that.  So under that scenario, it returns 800GB as available (1000 - 2 100 GB volumes if you treat them both as JBOD). Then the CO subtracts 100GB for the one JBOD volume it knows about, arriving at 700GB available for JBOD, which is the correct answer. I have to say this seems a bit silly, because we're already asking the SP for the *available* capacity, and the CO is calculating available a second time...  I think the SP should just return the right answer the first time. But things get weirder when we look at the second storage profile.

> ...and the total capacity for RAID1 profile volumes is 450G (and the available is 350G).

This is where I get lost. A `GetCapacity` for the RAID1 profile is going to return 600GB, not 450GB. Can you explain where you get 450GB from? I get 600GB by taking 1TB, and subtracting two 200GB volumes (because you've asked the SP to treat the volumes as RAID1, not JBOD). This is consistent with the JBOD example above.  From there, the CO would subtract the 1 100GB logical volume that is of the RAID1 storage profile, and think the available capacity is 500GB, which is wrong.

That's how I would walk through your logic, but clearly we differ here.

> Now I think this more as you mentioned it, I think maybe returning available will make the semantics more clear. It means the available storage space for creating volumes with the capabilities and parameters specified in the request.

Touched on this already, but I agree. `total_capacity` has no use. `available_capacity` does, and the description should be modified to remove the mention of *used*.

codenrhoden:
Ha - and `total_capacity` got turned into `available_capacity` as I was writing my previous comment. So that's a bit 👍  from me. :)

codenrhoden:
@jieyu Why have the CO subtract any available space from the result of `GetCapacity` at all? I think that's pushing too much login into the CO, when the SP can do it much better.

I'd rather see the newly renamed `available_capacity` act just like the `adjusted_available` parameter I outlined in https://github.com/container-storage-interface/spec/pull/85#issuecomment-323196574.  If that's the intent here, I think this is good. If COs are still going to try and modify that available space based on volume info they *think* they have an accurate view of, I think you will alway run into the consistency issues I outlined at https://github.com/container-storage-interface/spec/pull/85#issuecomment-323401306.

jieyu:
@codenrhoden yes, `available_capacity` is intended to be what you recommended. In other words, the "used" capacity won't be included in this number. If you think the wording in the current PR is not clear enough, let me know how can I make it better! Thanks again for the great feedback!

codenrhoden:
@jieyu Cool, when I read the current/new comments (as of this PR), I like what it says and think it is clear. It was just the scenario you outlined where the CO is going to modify the results that concerned me. But I don't think that's the case any longer, and perhaps you were just bringing it up as the thought exercise the group went through.

jieyu:
@saad-ali can you take a look at this PR if you have time? Looks like we have consensus here.

jdef:
fixes #59

saad-ali:
I like the direction this is going. This call is beginning to transform from a "GetCapacity", which it's not clear exactly how the CO should use that info, into a way for the CO to query the SP and ask if it can satisfy a particular request to provision based on current usage/capacity.

I've been thinking about topology (regions, zones, racks, etc.) a lot recently and how to incorporate the concept into the CSI spec. And one of the questions I'd like to be able to answer is given a topological constraint, let's say 3 zones, tell me which of them can successfully provision my request. We can discuss in future revisions. 

This LGTM for now.

Issue #91, opened 2017-08-24
tools: Go Dep tool


akutz:
I discovered that the official Go dependency management tool ([Dep](https://github.com/golang/dep)) has issues with this repository:

```bash
$ dep ensure 
ensure Solve(): No versions of github.com/container-storage-interface/spec met constraints:
    70e7ca5be45bd1c87c2a46836b9e930e9fcdc128: Could not introduce github.com/container-storage-interface/spec@70e7ca5be45bd1c87c2a46836b9e930e9fcdc128, as its subpackage github.com/container-storage-interface/spec does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
    master: Could not introduce github.com/container-storage-interface/spec@master, as its subpackage github.com/container-storage-interface/spec does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
    jdef-protobuf-headers: Could not introduce github.com/container-storage-interface/spec@jdef-protobuf-headers, as it is not allowed by constraint 70e7ca5be45bd1c87c2a46836b9e930e9fcdc128 from project github.com/codedellemc/gocsi.
    jdef_github_api_token: Could not introduce github.com/container-storage-interface/spec@jdef_github_api_token, as it is not allowed by constraint 70e7ca5be45bd1c87c2a46836b9e930e9fcdc128 from project github.com/codedellemc/gocsi.
```

The error occurs because there are no Go sources at the root of the repository. When I nil import `github.com/container-storage-interface/spec/lib/go/csi` in order to force the Dep tool to see that my project *does* depend on this repo via one of its sub-dirs with Go sources the tool will allow it.

However, this causes its own issues. See, this repo's `lib/go/csi` directory contains a Go source file generated by the `protoc` compiler using the gRPC Go plug-in. Importing this package into a Go program causes gRPC [to reregister all of the types](https://travis-ci.org/codedellemc/gocsi/jobs/268106805) found in the aforementioned package. However, that means it's not possible to *import* this repo without also being *forced* to use the generated Go gRPC types from the repo.

The only reason I want to depend on this repo, and thus vendor it in my own project, is to make it easy to obtain this repo's `spec.md` file so I'm able to generate the protobuf sources myself. I shouldn't be forced to use the sources provided by this repository. But as I said, I cannot import this repo without also using those types. The side-effect is that no other Go-based project that depends upon CSI types must also use the types from this repo.

I recommend one of two solutions:

## Solution 1
[...]

PR #93, opened 2017-08-24
Dep Support

akutz:
This patch fixes issue #91 by adding support for the official Go dependency tool, Dep. The solution is a nearly-empty Go source file at `lib/go/csi.go` that defines a package with no functional code. However, this package will enable the ability to nil-import `github.com/container-storage-interface/spec/lib/go` so that the Dep tool discovers a valid Go package (a directory with one or more Go sources in it).

For information on why the resolution is not to simply import `github.com/container-storage-interface/spec/lib/go/csi` please see issue #91.


saad-ali:
LGTM!

Issue #58, opened 2017-07-08
Remove ListVolumes RPC


cpuguy83:
I would like to propose that we remove the `ListVolumes` RPC.

The case for this RPC, IIRC, is that the CO would like to know what volumes are available. I don't think this is required though and the RPC itself is pretty complex (with [filtering](https://github.com/container-storage-interface/spec/issues/54), pagination). The list itself could become "corrupt" in between page results. 
Generally these can't (or shouldn't) be used as volumes until a `CreateVolume` is called since we'll usually want to store some metadata about the volume in the CO which won't be available for "listed" volumes.

Ideally the CO would keep a list of the volumes it knows about (which happened presumably got there by some request that eventually called `CreateVolume`). In the event of some data lost on the cluster, we can just call `CreateVolume` when needed since this is expected to be idempotent.

Additionally I have some experience with this as I implemented the `List` endpoint for docker volume plugins (for all the same reasons we implemented here), where we basically just end up with a bunch of a volumes and none of our expected metadata, and it ends up in a weird experience. I would actually like to phase out this call in Docker at some point.

jieyu:
> In the event of some data lost on the cluster, we can just call CreateVolume when needed since this is expected to be idempotent.

How do we find out the `name` and `parameters` for the `CreateVolumeRequest` in that case?

`ListVolumes` is *optional*. Neither CO or SP is forced to implement this. I think we'll end up with some experimental features that we're going to experiment with. If we can find a good way to allow CO/SP to experiment new APIs without forking the spec, I am all for making it experimental.

cpuguy83:
> How do we find out the name and parameters for the CreateVolumeRequest in that case?

Wouldn't the backend reject the create request if the params don't match?

jieyu:
@cpuguy83 In the case CO loses metadata (or crashes, and wants to do a reconciliation), this interface allows the CO to get all the known volumes from the SP. The CO don't have a list of volume names to call `CreateVolumeRequest`. Obviously, we cannot enumerate all volume names and parameters, and rely on rejection to tell the existence of a volume.

gpaul:
This likely supports the use case where external volumes are detached from one cluster and attached to another, too.

cpuguy83:
I guess I'm fine as long as it's optional. However, it seems like it is a bit difficult to use volumes listed from the plugin directly (no user interaction) since there will be none of the CO metadata associated with it available.

julian-hj:
[...]

PR #95, opened 2017-08-29
spec: Add volume capabilities and parameters to ListVolumesRequest

chhsia0:
This change allows the CO to get a list of volumes that satisfy
certain volume capabilities and are provisioned with certain parameters.
This is useful when the CO defines a storage class or profile and wants
to identify which class a pre-existing volume belongs to.


akutz:
While this is a great addition on the face of it the work itself, I'm inclined to suggest a moratorium on all changes to `ListVolumes` until #58 is resolved post v1. I'm personally in favor of #58, and if it is successful, this PR will be moot.

jdef:
Not crazy about the justification for this PR. It's possible that some set of capabilities/parameters are a subset of some other capabilities/parameters. Justifying this change as a method of determining a storage class/profile for a pre-existing volume is flawed because it's entirely possible that a volume might satisfy multiple combinations of capabilities/profiles.

Closing this out. If there's justification for this change along other lines, or it's critical for another workflow then please re-open with an explanation.

Issue #29, opened 2017-05-30
Deprecate VolumeMetaData in controller and node plugin API?


govint:
Given the CO never modifies any of the volume meta-data or plugin supplied volume-info is it necessary to keep passing the volume-meta-data on plugin API? The volume ID should be sufficient for the plugin to identify the meta-data for a volume on its own.

 Most likely, the plugin may return a subset of meta-data its keeping per volume via API responses. Passing that to the plugin isn't needed frankly.

jdef:
There were at least a few goal posts here:

1. Non-identifying volume metadata (`VolumeMetadata`), useful for plugins when interacting w/ volumes, may contain sensitive bits (keys, certs, etc), and is separate from...
1. Identifying volume metadata (`VolumeID`), useful to both CO and plugins, shouldn't contain sensitive bits (and may, for example, appear in log files).
1. The CSI working group (and wider storage community) felt that CSI should clearly separate these two kinds of metadata: `VolumeMetadata` and `VolumeID`.
1. It should be possible to write a relatively stateless CSI plugin; this is made easier if the CO can track both types of metadata and send it back to the plugin for subsequent RPCs.

As a result, the spec requires that CO track this additional per-volume metadata. This both increases the "data burden" of the CO and also adds a small amount of overhead to some RPCs.

Are there additional consequences that you're concerned with?

julian-hj:
The primary purpose of this metadata block was to give the plugin a place to put volume information so that it does not need to maintain state (#4 above).  Plugins can return a data blob that will get passed back to them in subsequent lifecycle calls.

For plugins that maintain their own backing store, I agree--the metadata block is totally unnecessary.  But we did not want to force plugin authors to have a backing store as that makes plugin implementation and configuration quite a bit more complex.

jdef:
Thanks for clarifying @julian-hj, next time I'll arrange my bullets in priority order :)

oritnm:
VolumeMetadata should not be the place to store sensitive data such as secrets, certs etc. Those needs to be passed thru a different mechanism since the same volume may be used by different containers each with a different secret and different access permission such as in the case the  volume is shared filesystem.

govint:
Agree on the CO keeping identifying meta-data (volume-ID), thats needed. But,
1. Storage plugins can mostly be expected to be stateless, as they front-end a storage system that generates/owns the meta-data. The plugin most likely is ferrying meta-data from the storage system and doesn't need the CO to keep any on its behalf.

[...]

Issue #87, opened 2017-08-18
Add size limits to calls where arbitrary amount of data can be passed in


saad-ali:
https://github.com/container-storage-interface/spec/pull/83#discussion_r133594577 adds a limit of 1 MB for credential data. We should consider something similar for other areas in the interface where an arbitrary amount of data can be passed in. It's always easier to start with a restriction and loosen it later then start with no limit and try to impose it later.

jdef:
field size limits were added here: https://github.com/container-storage-interface/spec/blob/master/spec.md#field-size-limits

please re-open this if needed

Issue #71, opened 2017-08-09
Specify which data a CO can safely cache


akutz:
Currently the spec does not specify whether or not a CO can safely cache certain data, such as a `NodeID`.  It is important to know whether or not this information should be cached or should be obtained every time it is required. For example, because the `NodeID` is not a single-value, its data map could be used to transport Node host-level information to a Controller host, such as a list of local devices on the node host. However, if the `NodeID` is cached this value should not be used for general data transport.

Cache characteristics and expectations should be explicitly detailed for each piece of data returned by the CSI RPCs.

saad-ali:
GetNodeId was replaced with NodeGetInfo. But this is still a valid issue.
We should define that the current fields in NodeGetInfo are immutable.
Wording for it may change across reinitialization needs to be decided.

jdef:
I'm having a bit of a hard time with this one. I thought I'd approach it by considering cases where the NodeID *might* reasonably change (even though COs don't want it to). For example, the plugin is uninstalled-then-reinstalled to the system. The CO isn't guaranteed to be the plugin supervisor, so it may not be aware of such a change - if the change is performed quickly, then the CO may think that the plugin has simply been restarted. Similarly, the backend storage provider may be reconfigured such that its no longer possible for the plugin to maintain the old NodeID (given that we want to support "stateless" plugins) since NodeID would be derived from the backend metadata.
Am I over-thinking this?

PR #97, opened 2017-09-01
spec: VolumeHandle replaces VolumeID, VolumeMetadata

jdef:
Refactor volume ID and metadata; provide additional clarification.
xref #29 #88


jdef:
xref #71

spec.md
```diff
@@ -537,35 +537,33 @@ message VolumeInfo {
...
+// that each Plugin keeps the information herein as small as possible.
+message VolumeHandle {
+
+  // ID is the identity of the provisioned volume specified by the
+  // Plugin. This field is REQUIRED. 
+  // This information SHALL NOT be considered sensitive such that, for
+  // example, the CO MAY generate log messages that include this data.
+  string id = 1;
+
+  // Metadata captures additional, possibly sensitive information about
```
saad-ali: nit: `additional, possibly sensitive, information`

spec.md
```diff
@@ -537,35 +537,33 @@ message VolumeInfo {
...
-  // each Plugin keeps this information as small as possible.
-  map<string, string> values = 1;
+// VolumeHandle objects are generated by Plugin implementations to
+// serve two purposes: first, to efficiently identify a volume known
+// to the Plugin implementation; second, to capture additional, opaque,
+// possibly sensitive information associated with a volume. It SHALL
+// NOT change over the lifetime of the volume and MAY be safely cached
+// by the CO.
+// Since this object will be passed around by the CO, it is RECOMMENDED
+// that each Plugin keeps the information herein as small as possible.
```
saad-ali: Maybe add in a size requirement, like we did with credentials. Always better to start with a hard limit and loosen it in the future, then not starting with a limit and trying to add it later. e.g. `The total bytes of the ID and metadata must be less than 1 Mebibyte.`

spec.md
julian-hj: I am not sure i agree with this nit.  To me it reads better without the extra comma.

saad-ali: Having a single comma suggests that "Metadata captures additional" is an introductory clause. But, "possibly sensitive information about..." can't stand on it's on (it is incomplete without the first part of the sentence).

Instead, "possibly sensitive" is nonessential information. And the sentence reads find without it.

From https://owl.english.purdue.edu/owl/owlprint/607/:

```
Commas with Nonessential Elements

Some modifying elements of a sentence are essential, restricting the meaning of a modified term, while others are nonessential and don't restrict the modified term's meaning. These nonessential elements, which can be words, phrases, or clauses, are set off with commas.

Rule: Use commas before and after nonessential words, phrases, and clauses, that is, elements embedded in the sentence that interrupt it without changing the essential meaning.
```

So I would offset "possibly sensitive" with two commas.

I'm being very pedantic here, so feel free to disregard =)

spec.md
```diff
@@ -537,35 +537,33 @@ message VolumeInfo {
...
+message VolumeHandle {
+
```
jieyu: nit: kill this line to be consistent with other places.

spec.md
jdef: agreed

spec.md
jdef: ack

spec.md
jdef: ack

spec.md
jdef: xref #87

saad-ali:
LGTM. Thanks @jdef. Merging.

PR #98, opened 2017-09-01
spec/build: address protobuf max line length violations

jdef:


Makefile
```diff
@@ -43,4 +43,8 @@ clean:
 clobber: clean
 	rm -f $(CSI_PROTO) $(CSI_PROTO).tmp
 
-.PHONY: clean clobber
+# check generated files for violation of standards
+check: $(CSI_PROTO)
+	awk '{ if (length > 72) print NR, $$0 }' $? | diff - /dev/null
```
saad-ali: Why 72 and not 80?

akutz:
No clue. I prefer 80, but y'all are the ones that decided on 72. That's certainly the limit for git commit bodies, but I believe in 80 for code.

-- 
-a

Makefile
jdef: The CONTRIBUTING guide says:
> Please wrap the code at 72 characters.

.. so that's what this PR enforces.

saad-ali: Ack, we should revisit that decision.

saad-ali:
LGTM

Opened low-pri issue, https://github.com/container-storage-interface/spec/issues/101, to change limit to 80.

PR #109, opened 2017-09-20
spec: Add UNEXPECTED_FIELD general error code.

gpaul:
This PR adds a general error code that can be used when optional fields are
provided when they are not expected.

These cases are:

- `NodePublishVolumeRequest.publish_volume_info` is provided
  even though the corresponding Controller Plugin has no
  `PUBLISH_UNPUBLISH_VOLUME` capability.
- `ControllerPublishVolumeRequest.node_id` is specified
  even though none was returned from `GetNodeID`.
- `ControllerUnpublishVolumeRequest.node_id` is specified
  even though none was returned from `GetNodeID`.


gpaul:
The alternative is to use `GeneralError: UNDEFINED` with a suitable `ErrorDescription`. Then again, that's true for virtually every error that reports incorrect plugin<->CO semantics, hence this PR.

cpuguy83:
In general I feel like we should be moving to fewer codes.
I generally see these codes as a class of error, e.g. bad user input and then a descriptive message.
For a "class of error",  I see this as "how will the caller react to this", and in the case of bad user input it will always be to tell the user they did something wrong.

There is a PR open from @saad-ali to move error handling over to use GRPC error codes rather than custom error codes: #115 which I think is relevant here.

jdef:
this should probably be re-cast now that #115 has landed. maybe as a case where `InvalidArgument` may be returned?

jdef:
The spec documents that InvalidArgument should be returned for this case, for all RPCs. Closing this out. See https://github.com/container-storage-interface/spec/blob/master/spec.md#error-scheme

PR #111, opened 2017-09-20
Remove Trailing Whitespace from the Spec

akutz:
This patch removes trailing whitespace from the CSI specification.


gpaul:
This is great. Do consider adding a 'strip whitespace' command to the Makefile `build` target and enforce the check in CI - otherwise it is just a matter of time until some `nano` user adds spaces again.

akutz:
Hi @gpaul,

Good call. I know most of it was added by @jdef when he authored the documents, but I **also** know he uses Vim on, presumably, Linux. My attempts to edit them with Vim on macOS also strip the whitespace. It's when I surmised it was some default behavior based on OS/editor combination.

Issue #23, opened 2017-05-27
Missing error code (for volume lifecycle, plugin-specific behavior, etc.).


chhsia0:
Should we add new error codes to indicate a plugin-specific error for each RPC call?

**DeleteVolume**: When a volume is in the NODE_READY or PUBLISHED state, what error code should we return?

**ControllerUnpublishVolume**: When a volume is in the PUBLISHED state, what error code should we return?

**ListVolumes**: Should we add a `ListVolumesError`?

Misc:
What is `UNSUPPORTED_VOLUME_TYPE` for?
`GetNodeIDError` and `ProbeNodeError` seems redundant.

jdef:
Don't we already have a general/generic error code that implementations can use for vendor-specific codes?

**ListVolumes**: are there specific errors that you'd like to see? how would the recovery semantics differ across such errors?

chhsia0:
For vendor-specific error handling: the closest error code we have is `GeneralError::UNDEFINED`. I guess a plugin can use this to return any vendor-specific error.

For *ListVolumes*, maybe we can have an `INVALID_TOKEN`.

jdef:
Specific error codes are only useful if there are meaningful recovery semantics. Do you have any recommendations for recovery semantics for additional error codes that you'd find useful?

cpuguy83:
How come we aren't using grpc error codes?

jdef:
gRPC error codes are intended for gRPC-protocol-level errors. for CSI ops
they're too low level: they don't allow us to encapsulate additional
information, and don't allow us to accurately model/advertise the recovery
semantics that we want clients to use.

cpuguy83:
grpc error codes are for standard ways to communicate classes of errors across the RPC boundary, including errors that are temporary, requests that can be retried vs should not be retried, etc.

chhsia0:
I had a similar discussion with @jdef and @jieyu. The gRPC error codes are used to indicate connection/protocol errors, so we might not want CSI-level errors mix with or rely on gRPC's error handling. And also we can only return strings through gRPC errors, but we might want to have richer formats for CSI errors. I'm personally not familiar with gRPC's error handling though. Is there any benefits not defining CSI errors independently?

chhsia0:
FYI we can use `grpc::Status::error_details()` to deliver a binary string that represents a protobuf or anything we want for error handling.

stevvooe:
[...]

PR #115, opened 2017-10-02
Use gRPC standard status error model

saad-ali:
As discussed in https://github.com/container-storage-interface/spec/issues/23


akutz:
Hi @saad-ali,

A few things from a brief glance:

1. I personally like PR descriptions to have more than just a link to the original issue, but I understand why people do it. Still, a short problem summary is always helpful.

2. I would call this out in the description wayyy up front: `Default value for backwards compatibility`. I didn't see those in the diff; not until I viewed the file sans the diff mode. It's probably me, but it was very confusing seeing those old error enums at first.

3. It would also be useful to have the PR's description show the old behavior vs. the new behavior from a client's, and possibly server's, perspective.

Thanks for this hard work. This is going to make CSI even better and set a strong foundation on which to build future features!

cpuguy83:
We discussed some issues with a couple of cases that you felt didn't quite line up with grpc error codes, can you explain them?

We also discussed that while the protocol supports it, the various libraries will not let you return an error and some object (like more detailed/structured error data, or a volume handle for the `AlreadyExists` case).
It looks like the go library handles metadata via the context object (which you can set values on for requests and responses).
I'm not sure if metadata is the right place for this or should be considered at all (@stevvooe ?)

csi.proto
```diff
@@ -305,11 +293,7 @@ message DeleteVolumeRequest {
 message DeleteVolumeResponse {
   message Result {}
 
-  // One of the following fields MUST be specified.
-  oneof reply {
-    Result result = 1;
-    Error error = 2;
-  }
+  Result result = 1;
```
stevvooe: When you have an empty message, go ahead and use `google.protobuf.Empty`.

Was there a reason for the extra encapsulation here?

spec.md
```diff
@@ -1116,16 +1056,24 @@ message NodeServiceCapability {
 }
 ```
 
-### Error Type/Codes
+### Error Handling
+
+All CSI API calls defined in this spec MUST return a [standard gRPC status](https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto).
```
stevvooe: Double check, but I think this is implied by the protocol on error.

stevvooe:
@cpuguy83 You use the [`status package`](https://godoc.org/google.golang.org/grpc/status). It places a detail object on the header (effectively, metadata from grpc perspective) but it is transparent. [Here](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Status.java) is the same piece in java. Looks like python has support with [Call.details()](https://grpc.io/grpc/python/grpc.html#grpc.Call.details).

We can figure out how to upstream changes for other client packages, if it is missing.

Let me know if that addresses the concern.

csi.proto
saad-ali: Will do.

saad-ali:
@akutz:

> I personally like PR descriptions to have more than just a link to the original issue, but I understand why people do it. Still, a short problem summary is always helpful.

I agree! Wanted to get this out in front of everyone as a work-in-progress ASAP.

> It would also be useful to have the PR's description show the old behavior vs. the new behavior from a client's, and possibly server's, perspective.

Yep, will do as soon as this is no longer WIP.

> I would call this out in the description wayyy up front: Default value for backwards compatibility. I didn't see those in the diff; not until I viewed the file sans the diff mode. It's probably just me, but it was very confusing seeing those old error enums at first.

Not sure I follow.

@cpuguy83:

+1 to @stevvooe. As for the cases that we need to think through, the ones that involve idempotency: specifically `CreateVolume`, if it returns an error (AlreadyExists), then we may need to introduce a `GetVolumeInfo(...)` call (I'm leaning towards this)

Everyone:
I'm looking in to resusing [google/rpc/error_details.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) for error details, instead of defining our own custom error details. Any objections to this?

spec.md
saad-ali: I don't see anywhere in the gRPC docs where it is prescribed, so I want it to be very explicit for CSI.

stevvooe: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

> All RPCs started at a client return a status object composed of an integer code and a string message.

There might be something more normative, elsewhere.

stevvooe:
> +1 to @stevvooe. As for the cases that we need to think through, the ones that involve idempotency: specifically CreateVolume, if it returns an error (AlreadyExists), then we may need to introduce a GetVolumeInfo(...) call (I'm leaning towards this)

I'd recommend adding that anyways. ;)

> I'm looking in to resusing google/rpc/error_details.proto for error details, instead of defining our own custom error details. Any objections to this?

Check with the grpc team and see what the scope of this is. This looks like it is generally usable, but they may have other ideas.

spec.md
```diff
@@ -1116,16 +986,24 @@ message NodeServiceCapability {
...
+The status `code` MUST contain a [canonical error code](https://github.com/grpc/grpc-go/blob/master/codes/codes.go).
+The status `message` MUST contain a human readable description of error, if the status `code` is not `OK`.
+This string MAY be surfaced by CO to end users.
+The status `details` MUST return the machine-parsable `Error` protobuf message defined below, if the status `code` is not `OK`.
+This allows CO to implement smarter error handling and fault resolution.
 
 ```protobuf
 message Error {
   // General Error that MAY be returned by any RPC.
   message GeneralError {
```
jieyu: Sounds like we don't need to define the enum for general errors as we'll use the general grpc error code. Here is my suggestion:
1) For each general error (that applies to all RPCs) code from grpc that we'll use, have a text to explain what that means in this section (like what we have for now), and define the corresponding error details struct. We can define a GeneralErrorDetails message that'll be reused by all general errors.
2) For RPC specific errors, we should place the corresponding error explanation close to where the RPC is. We might want to define RPC specific error details type.

stevvooe:
@saad-ali While we're not using error_details.proto yet, https://github.com/containerd/containerd/pull/1597 is an example of where we brought those protobufs in for use. Your project is a little different because you have generated go and protos in different places, but hopefully that helps with some hints to get started.

akutz:
Hi All,

This, to me, is the most impactful, outstanding PR scheduled for inclusion in v0.1.0. If the latter is going to have any hope of being tagged in time for COs like Mesos and K8s to be compatible with plug-ins being actively developed, this PR **must** be resolved ASAP.

jieyu:
+1 to @akutz said. @saad-ali do you need any help here? I'd like to see this get merged asap.

stevvooe:
@jieyu It looks like the only outstanding thing is the removal of the extra error codes that overlap with the standard error codes. Did you still need any help on my end?

lpabon:
Some extra info for those googling this: http://avi.im/grpc-errors/

lpabon:
Looks like the csi.pb.go needs to be updated from the csi.proto. Example:

[ csi.pb.go ](https://github.com/container-storage-interface/spec/pull/115/files#diff-308dbd5d943fd4b49fc73f1092c253abL426
) : [ csi.pb.go ](https://github.com/container-storage-interface/spec/blob/d68da7a7c330a88253e26c06b7c157ec8119d5d6/lib/go/csi/csi.pb.go#L2469)

vladimirvivien:
So, if this PR adopts gRPC's standard error-signaling and error-handling types/mechanism, @saad-ali (et al) will the multi-level nested response (with Response, Reply) also go away in favor of a flatter model ?  

I am guessing the `oneof reply` approach was adopted so that either a valid response or an error could be returned in one response as shown below.

```protobuf
message GetSupportedVersionsResponse {
  message Result {
    repeated Version supported_versions = 1;
  }

  oneof reply {
    Result result = 1;
    Error error = 2;
  }
}
```
So, if the `Error` message will no longer be part of the `xxxResponse` message, will responses be flatter as in something like:
```protobuf
message GetSupportedVersionsResponse {
    repeated Version supported_versions = 1;
}
```

saad-ali:
Rebased.

saad-ali:
> So, if this PR adopts gRPC's standard error-signaling and error-handling types/mechanism, @saad-ali (et al) will the multi-level nested response (with Response, Reply) also go away in favor of a flatter model ?

Yes!

saad-ali:
Ok, I finally got around to this PR! Apologies for the delay.

For now I decided not to have any custom error details proto. Instead I specified standard gRPC error codes for all existing error conditions. In the future, if we decide that error details are critical we can introduce them. I also changed the idempotent retries (e.g. volume already exists) to success (0 OK) instead of an error code.

PTAL, I'd like to get this sucker merged once and for all.

spec.md
```diff
@@ -583,6 +584,22 @@ message VolumeInfo {
...
 
+##### CreateVolume Errors
+
+If the plugin is unable to complete the CreateVolume call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Volume already exists | 0 OK | Indicates that a volume corresponding to the specified volume `name` already exists. | Caller MUST assume the `CreateVolume` call succeeded. |
```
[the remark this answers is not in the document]
saad-ali: No, this is on purpose. For idempotency if the volume already exists the request should succeed.

spec.md
```diff
@@ -782,25 +817,28 @@ message ValidateVolumeCapabilitiesRequest {
...
 
+##### ValidateVolumeCapabilities Errors
+
+If the plugin is unable to complete the ValidateVolumeCapabilities call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Unsupported `volume_capability` | 0 OK | Indicates that one or more of the specified `volume_capability` options (mount flags, volume type, FS type, etc.) are not supported. `ValidateVolumeCapabilitiesResponse.supported` MUST be `false`. `ValidateVolumeCapabilitiesResponse.message` must detail what volume capabilities are incompatible. | Caller MUST fix the `volume_capability` before retrying. |
```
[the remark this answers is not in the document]
saad-ali: Ditto.

spec.md
```diff
@@ -307,6 +310,22 @@ These limits apply for messages generated by both COs and plugins.
...
 | 4 KiB      | map<string, string> |
 
+### Error Scheme
+
+All CSI API calls defined in this spec MUST return a [standard gRPC status](https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto).
+Most gRPC libraries provide helper methods to set and read the status fields.
+
+The status `code` MUST contain a [canonical error code](https://github.com/grpc/grpc-go/blob/master/codes/codes.go). COs must handle all valid error codes. Each RPC defines a set of gRPC error codes that MUST be returned by the plugin when specified conditions are encountered. In addition to those, if the conditions defined below are encountered, the plugin MUST return the associated gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
```
jieyu: Can you also include a line for the following in the current spec.
```
// Indicates that a required field is missing from the request.
// More human-readable information MAY be provided in the
// `error_description` field. The `caller_must_not_retry` field
// MUST be set to true.
//
// Recovery behavior: Caller MUST fix the request by adding the
// missing required field before retrying.
MISSING_REQUIRED_FIELD = 3;
```

I think using `INVALID_ARGUMENT` is more appropriate here because CO should not retry in this case.

spec.md
```diff
@@ -583,6 +584,22 @@ message VolumeInfo {
...
+##### CreateVolume Errors
+
+If the plugin is unable to complete the CreateVolume call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Volume already exists | 0 OK | Indicates that a volume corresponding to the specified volume `name` already exists. | Caller MUST assume the `CreateVolume` call succeeded. |
+| Invalid `name` | 3 INVALID_ARGUMENT | Indicates that the specified volume name is not allowed by the Plugin. More human-readable information MAY be provided in the gRPC `status.message` field. | Caller MUST fix the name before retrying. |
```
jieyu: My suggestion is to have a general "Invalid or unsupported field in the request" condition (`INVALID_ARGUMENT`), and put that in the general error section above so that we don't have to mention them here for each call. What do you think?

spec.md
```diff
@@ -583,6 +584,22 @@ message VolumeInfo {
...
+| Unsupported `parameters` key or value | 3 INVALID_ARGUMENT | Indicates that either one (or more) of the keys in the opaque key/value `parameters` field is not supported by the Plugin and/or one of the values contains invalid data. More human-readable information MAY be provided in the gRPC `status.message` field. This MAY occur, for example, due to caller error, Plugin version skew, etc. | Caller MUST remove the unsupported key/value pair(s) from the list of parameters before retrying. |
```
jieyu: Ditto. Once we have the general "Invalid field in the request", we can probably just remove this row here.

spec.md
jieyu: Why "retry" is needed if the response is OK?

spec.md
```diff
@@ -877,23 +917,25 @@ message GetCapacityRequest {
...
 
+##### GetCapacity Errors
+
+If the plugin is unable to complete the GetCapacity call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Unsupported `volume_capability` | 3 INVALID_ARGUMENT | Indicates that one or more of the specified `volume_capability` options (mount flags, volume type, FS type, etc.) are not supported. | Caller MUST fix the `volume_capability` before retrying. |
```
jieyu: This fits into the general invalid/unsupported fields condition? Consider remove it?

spec.md
```diff
@@ -906,17 +948,18 @@ message ControllerProbeRequest {
...
-  }
-}
-```
+If the plugin is unable to complete the ControllerProbe call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Bad plugin config | 3 INVALID_ARGUMENT | Indicates that the plugin is misconfigured. | Caller MUST assume the plugin is not healthy. |
```
jieyu: `INVALID_ARGUMENT` here is weird because we don't really specify any argument for the call :)
I prefer `FailedPrecondition` here.

spec.md
```diff
@@ -1023,17 +1062,19 @@ message NodePublishVolumeRequest {
...
-  }
-}
-```
+If the plugin is unable to complete the NodePublishVolume call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Unsupported `volume_capability` | 3 INVALID_ARGUMENT | Indicates that one or more of the specified `volume_capability` options (mount flags, volume type, FS type, etc.) are not supported. | Caller MUST fix the `volume_capability` before retrying. |
```
jieyu: Ditto

spec.md
```diff
@@ -1076,17 +1117,19 @@ message NodeUnpublishVolumeRequest {
...
-  }
-}
-```
+If the plugin is unable to complete the NodeUnpublishVolume call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Unsupported `volume_capability` | 3 INVALID_ARGUMENT | Indicates that one or more of the specified `volume_capability` options (mount flags, volume type, FS type, etc.) are not supported. | Caller MUST fix the `volume_capability` before retrying. |
```
jieyu: Ditto.

spec.md
```diff
@@ -1132,17 +1171,18 @@ message NodeProbeRequest {
...
-  }
-}
-```
+If the plugin is unable to complete the NodeProbe call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Bad plugin config | 3 INVALID_ARGUMENT | Indicates that the plugin is misconfigured. | Caller MUST assume the plugin is not healthy. |
```
jieyu: Ditto. "FAILED_PRECONDITION" ?

spec.md
```diff
@@ -248,13 +248,13 @@ service Controller {
     returns (CreateVolumeResponse) {}
 
   rpc DeleteVolume (DeleteVolumeRequest)
-    returns (DeleteVolumeResponse) {}
+    returns (Empty) {}
```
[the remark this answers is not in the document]
akutz: I'm conflicted on this. While I agree with much of @jieyu's reasoning, this bit strikes me in a funny way:

> add an optional field to `DeleteVolumeResponse` will be backward compatible (without the need for bumping a CSI version).

While this is technically *true*, if there *is* additional information added to otherwise empty responses, is bumping the version such a bad thing? Otherwise the version wouldn't reflect the addition of new information. Let's look at the stated behavior of SemVer:

> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

It would seem then that a MINOR version change would be required in the case of keeping `DeleteVolumeResponse` for now in order to possibly add new data to it in the future. However, I'm not sure I agree with this. To me a MINOR change is valid if we, for example, added a new RPC that provides extended functionality. However, changing the response data for an RPC, even if technically backwards compatible, is to me a MAJOR change as it affects an existing RPC. Again, I understand it is technically a MINOR change, but I'm talking about in practice, not in theory.

Therefore I'm fine leaving this items `Empty` for now and then requiring a MAJOR version bump if we should ever change them.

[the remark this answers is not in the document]
cpuguy83: This would only be backwards incompatible at the library layer, right?
The proto/wire format should be fine.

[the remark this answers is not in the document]
jieyu: Yeah, i think wire format should be the same. Just i feel the benefit of doing this is thin. It's nice to keep request/response symmetric.

spec.md
```diff
@@ -307,6 +310,22 @@ These limits apply for messages generated by both COs and plugins.
...
+### Error Scheme
+
+All CSI API calls defined in this spec MUST return a [standard gRPC status](https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto).
+Most gRPC libraries provide helper methods to set and read the status fields.
+
+The status `code` MUST contain a [canonical error code](https://github.com/grpc/grpc-go/blob/master/codes/codes.go). COs must handle all valid error codes. Each RPC defines a set of gRPC error codes that MUST be returned by the plugin when specified conditions are encountered. In addition to those, if the conditions defined below are encountered, the plugin MUST return the associated gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Unsupported request version | 3 INVALID_ARGUMENT | Indicates that the version specified in the request is not supported by the Plugin. | Caller MUST NOT retry; caller SHOULD call `GetSupportedVersions` to discover which CSI versions the Plugin supports. |
```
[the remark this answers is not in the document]
akutz: Hmm. This **is** tricky. As @jieyu points out, the [gRPC error code docs](https://github.com/grpc/grpc-go/blob/master/codes/codes.go) state:

> // Note that this differs from FailedPrecondition. It indicates arguments
> // that are problematic regardless of the state of the system

However, the `FailedPrecondition` doc says:

> // FailedPrecondition indicates operation was rejected because the
> // system is not in a state required for the operation's execution.

I read this as a `FailedPrecondition` being an argument's relationship to the state of the system. The question becomes -- are the versions supported by a CSI SP part of the system's state? I don't know. I could go either way here.

spec.md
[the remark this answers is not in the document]
akutz: I disagree (or rather I agree with @cpuguy83). One part of the purpose behind this new error model was the ability to return both valid data **and** associated error codes. I want to know that the volume already exists **and** have the `VolumeInfo` for that volume.

spec.md
akutz: That's what I'm wondering. I don't think an error code of 0 is appropriate if one of the options are not supported.

spec.md
akutz: I concur 100%. This is the entire purpose of `FailedPrecondition` as it relates to *the state of the system*.

spec.md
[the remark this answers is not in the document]
akutz: > This would only be backwards incompatible at the library layer, right?

That's correct. Truth be told though, adding a new RPC, while a MINOR SemVer bump, would be backwards incompatible for any Go library.

I think we have to agree (or not) that compatibility is with regards to the spec, not with regards to a specific language binding/implementation of the spec.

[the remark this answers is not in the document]
akutz: > I think we have to agree (or not) that compatibility is with regards to the spec, not with regards to a specific language binding/implementation of the spec.

Actually, I think the above is a good topic for next week's meeting. Is Go prevalent enough that we should consider it when considering SemVer for the spec?

spec.md
saad-ali: Will do.

spec.md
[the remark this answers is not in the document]
saad-ali: I'm going to leave it as `InvalidArgument` because I don't want the caller to retry without human intervention.

spec.md
saad-ali: I'd prefer to leave them split up so that if we want we can convert these in to detailed error codes in the future.

spec.md
saad-ali: Let me know if you agree with my statement above. If you feel strongly, I'm willing to change it.

akutz:
If this PR is no longer going to return `ALREADY_EXISTS` error codes along with the data for the existing object, then I fail to see the value in this PR. I was under the impression the guiding purpose of this PR was to be able to return both informative error codes and related data concurrently as the result of a single call. Otherwise this is back to the same discussion regarding whether to return an error or the result in an idempotent situation.

spec.md
saad-ali: The `ValidateVolumeCapabilitiesResponse` response has a `supported` bool. My understanding was that even if capabilities were not compatible, for this RPC the code should be ok and the boolean would indicate that. But based on what @akutz mentioned above it sounds like we can have both a non-OK error code and a `ValidateVolumeCapabilitiesResponse` response. So I'll change this to `INVALID_ARGUMENT`

spec.md
saad-ali: One second thought, I'll go ahead and change it as you described. We can always re-add these in the future, if needed.

spec.md
saad-ali: One second thought, I'll go ahead and change it as you described. We can always re-add these in the future, if needed.

spec.md
saad-ali: Will do

spec.md
saad-ali: Agreed. Brain farted on that one.

akutz:
Hi @saad-ali,

> If this PR is no longer going to return ALREADY_EXISTS error codes along with the data for the existing object, then I fail to see the value in this PR. I was under the impression the guiding purpose of this PR was to be able to return both informative error codes and related data concurrently as the result of a single call. Otherwise this is back to the same discussion regarding whether to return an error or the result in an idempotent situation.

The more I think about what I wrote above, the more I must propose we delay this PR to post v0.1. If this PR isn't going to introduce the ability to receive both error codes **and** data as the  result of a single call then I do not think the change this PR introduces is worth the trouble to adopt it so close to when people expect to be able to use CSI.

I will have to go back and find our discussion, but we all agreed that COs may both wish to know if an object does or doesn't already exist and then still have that object's data returned. I disagree with @jieyu's statement:

> If AlreadyExists is used, the CO might not be able to figure out the VolumeInfo for the already created volume. cc @julian-hj

Of course the CO can figure that out. There are many examples of when a client/receiver must consider more than just `0` as a successful exit code. This is the same as that. For example:

```go
if err == nil || err.Code == ALREADY_EXISTS
```

The above Go-based pseudo-code demonstrates how simple this is.

If this PR will not return error codes related to idempotency along with existing object data then I propose this PR be delayed until after v0.1.

jieyu:
@akutz I think setting both gRPC status and response sounds good to me too. The CO will have slightly more logic to deal with that, but that should be small like you said.

spec.md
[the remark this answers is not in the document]
saad-ali: I don't feel strongly either way. Since both @akutz and @cpuguy83 perfer the error code. I'll change it to that.

spec.md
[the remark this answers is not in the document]
saad-ali: I don't feel strongly either way. Since both @jieyu strongly prefers symmetry, I'll go ahead and make it symmetric.

saad-ali:
Feedback addressed. PTAL

akutz:
FWIW, I'm already 95% of the way to adopting it in thecodeteam/gocsi#32. Except for the client, the new error model has been pushed into the interceptors, mock SP, tests, etc. Working very, very well. Thank you again @saad-ali!

spec.md
```diff
@@ -583,6 +583,20 @@ message VolumeInfo {
...
 
+##### CreateVolume Errors
+
+If the plugin is unable to complete the CreateVolume call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Volume already exists | 6 ALREADY_EXISTS | Indicates that a volume corresponding to the specified volume `name` already exists. Plugin MUST also return a valid `CreateVolumeResponse`. | Caller MUST assume the `CreateVolume` call succeeded. |
```
jdef: I don't think this is possible, at least for golang-based servers. Has anyone tested this approach? I'm looking at the golang grpc code and it seems like if an RPC generates an error, then the xxxResponse is dropped and just "status" is returned to the client: https://github.com/grpc/grpc-go/blob/8ff8683602df4fb6a85c22f2da944bfc8fc73c50/server.go#L877

jdef: clarification: xxxResponse being the non-status payload

spec.md
```diff
@@ -583,6 +583,20 @@ message VolumeInfo {
...
+| Unsupported `capacity_range` | 11 OUT_OF_RANGE | Indicates that the capacity range is not allowed by the Plugin. More human-readable information MAY be provided in the gRPC `status.message` field. | Caller MUST fix the capacity range before retrying. |
+| Call not implemented | 12 UNIMPLEMENTED | CreateVolume call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST not retry caller MAY call `ControllerGetCapabilities` or `NodeGetCapabilities` to discover Plugin capabilities. |
```
jdef: UNIMPLEMENTED is also returned by the grpc libs for other cases. Maybe consider another, non-overlapping error code for this? see https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

jdef: ditto for all other uses of UNIMPLEMENTED by this spec

spec.md
```diff
@@ -683,21 +703,28 @@ message ControllerPublishVolumeRequest {
...
+| Volume published to another node | 6 ALREADY_EXISTS | Indicates that a volume corresponding to the specified `volume_id` is already attached to another node and does not support multi-node attach. If this error code is returned, the Plugin SHOULD specify the `node_id` of the node the volume is already attached to as part of the gRPC `status.message`. | Caller SHOULD ensure the specified volume is not attached to any other node before retrying with exponential back off. |
+| Max volumes attached | 8 RESOURCE_EXHAUSTED | Indicates that the maximum supported number of volumes that can be attached to the specified node are already attached. Therefore, this operation will fail until at least one of the existing attached volumes is detached from the node. | Caller MUST ensure that the number of volumes already attached to the node is less then the maximum supported number of volumes before retrying with exponential backoff. |
```
jdef: RESOURCE_EXHAUSTED is also returned for other reasons by the grpc library: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

jdef: suggest FAILED_PRECONDITION instead

spec.md
jdef: suggest ABORTED instead

spec.md
```diff
@@ -721,10 +748,10 @@ message ControllerUnpublishVolumeRequest {
...
   // clear) this field to match the node ID returned by `GetNodeID`.
-  // `GetNodeID` is allowed to omit node ID from a successful `Result`;
+  // `GetNodeID` is allowed to omit node ID from a successful response;
   // in such cases the CO SHALL NOT specify this field.
   //
-  // If `GetNodeID` does not omit node ID from a successful `Result`,
+  // If `GetNodeID` does not omit node ID from a successful response,
   // the CO MAY omit this field as well, indicating that it does not
   // know which node the volume was previously used. The Plugin SHOULD
   // return an Error if this is not supported.
```
jdef: s/Error/error/

spec.md
```diff
@@ -828,30 +864,32 @@ message ListVolumesRequest {
...
 
+##### ListVolumes Errors
+
+If the plugin is unable to complete the ListVolumes call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+Condition | gRPC Code | Description | Recovery Behavior
+| --- | --- | --- | --- |
+| Invalid `starting_token` | 10 ABORTED | Indicates that `starting_token` is not valid. | Caller SHOULD start the `ListVolumes` operation again with an empty `starting_token`. |
```
jdef: there's no UNIMPLEMENTED row here, like for the other optional calls. suggest adding another row for this but as ABORTED (as suggested above).

spec.md
cpuguy83: I believe you do this with `status.New(code, msg).WithDetails(proto)`... and then `s.Err()` to get the underlying `error` implementation.

chhsia0: It seems a response with a a non-OK status is not recommended from the discussion of the open issue here: https://github.com/grpc/grpc/issues/12824#issuecomment-335624435

cpuguy83: In the example above the detailed error is encoded with the status message and can be decoded on the other side (or at least that's how I understand the API, I haven't tried to use it yet).

akutz: See the GoCSI PR I linked. I’m already using the error message there.

julian-hj: That was my understanding also, although in my case it was based on hearsay.  Since we claim that these calls are idempotent, I think it would be OK if we just returned no error for the case where the input options are all the same and the volume is already created.  That would allow us to maintain idempotency.

We could then treat ALREADY_EXISTS as an error case, for the instance when someone tries to create the same volume with different inputs.

chhsia0: Although the current gRPC implementation allows us to return a response with a non-OK status, it doesn't seem a reliable feature to me, as the convention described in the comment I referenced is against that. Serializing `VolumeInfo` in the error details sounds more reliable, but the CO would have to deserialize it manually, and this should be explained in the spec.

akutz: Hi @chhsia0,

It actually works fine as is. Please see GoCSI's [idempotent create tests](https://github.com/akutz/gocsi/blob/b356efea1c3729a5fd50455cb592e12314353083/controller_test.go#L147-L256) where I scale up to 100,000 concurrent `CreateVolume` RPCs, and they all return the `VolumeInfo` and gRPC error successfully.

chhsia0: I've read your PR and I'm sure your tests work just fine. I raise this question because it seems to me not a good idea to rely on a gRPC implementation-dependent feature that is not ensured by gRPC's semantics, and I was wondering if this should be a concern, or someone can tell me that I'm totally wrong and the support of response with a non-OK status is guaranteed by gRPC's community.

akutz: Hi @chhsia0,

> to rely on a gRPC implementation-dependent feature that is not ensured by gRPC's formal semantics

I'm not sure what you mean? Perhaps I'm missing something, but how is returning a tuple with both fields populated any less formal than the gRPC error codes? Or is the ability to return both a non-nil response **and** error not actually a part of gRPC? Is it purely a side-effect of Go's implementation?

chhsia0: According Dr. Kumar's (I think he's the engineering lead of gRPC) comment I cross-referenced above, the convention of gRPC's protobuf-based type-safe API layer for unary response is that "Client side SHOULD drop any response messages received with a non-OK response."

However, he also mentioned that gRPC's wire protocol does allow any combination. My interpretation is that the wire protocol does accept a non-OK status with a response, but proto-based API designs such as CSI should not rely on this. @saad-ali can you clarify on this? Thanks.

akutz: Hi @chhsia0,

>  "Client side SHOULD drop any response messages received with a non-OK response."

If the above is in fact true then I agree that the error details is the correct location for any idempotent data.

chhsia0: Not sure if this has been discussed, but another possible option (if we want to avoid using the error details for any reason) is to return an OK and introduce a boolean field in the response to indicate if a volume of the same capabilities/parameters has been created before, and use `ALREADY_EXISTS` as an error suggested by @julian-hj.

akutz: Hi @chhsia0,

I don't like that approach because it subverts (or compliments in a way that replaces) the existing error model that is designed to handle the case of something that already exists. I recommend using the gRPC error details as you've already suggested.

julian-hj: As far as I understand it, if a function is idempotent, then calling it twice with the same inputs should not generate an error code.  So if "ALREADY_EXISTS" is an error, we should not use it for the case where the client just got impatient and repeated an RPC call, or repeated a call after a dropped connection.

What I am not clear on is whether "ALREADY_EXISTS" is deemed as an error condition or not?

cpuguy83: I'm fine either way.
For me idempotent means you can run the same thing multiple times without side-effects. The action can be idempotent (nothing changed) while still providing more information to the caller (this thing already exists) instead of "yep we did that".

However there is a case where a volume could exist but the state of it conflicts with the passed in options. In which case ALREADY_EXISTS could work... but also maybe a different error code applies to this.

PR #118, opened 2017-10-04
spec: Add tags to VolumeInfo

akutz:
This patch adds an optional field `tags` to `VolumeInfo`. The `tags` field is a `map[string]string` and can contain additional attributes related to a volume.

Related to #116, #117. cc/ @saad-ali @jieyu


saad-ali:
This looks mostly fine to me. It does complicate VolumeInfo in that it adds a field that should be populated on responses but should not be present on requests. Wondering if there is anything we could do to make that cleaner, but I do like having essentially a single object that captures all volume attributes. So I'm not sure.

akutz:
Hi Saad,

I’d argue that the validate volume caps doesn’t need the volume info obj anyway, only the handle it contains. Why not just update the validate rpc to use the handle directly?

saad-ali:
Good point. I would support replacing VolumeInfo with id in `ValidateVolumeCapabilities` along with this change.

akutz:
Cool. I’ll update the PR with this change when I return to my desk.

jdef:
Are `tags` stable over time, or do they have a TTL? How are these different than `metadata`, other than the fact that they're not passed back into the plugin for publish/validation calls?

jdef:
There are no concrete use cases for this, so I'm closing it out. We can revisit later if needed. If you disagree, please re-open for comment.

Issue #18, opened 2017-05-25
error codes: clarify gRPC error code usage vs. CSI error code usage


jdef:
gRPC has an error code mechanism that's independent of the CSI error code mechanism. Provide clarifying docs about error reporting expectations. Namely that CSI errors are returned with a successful/normal gRPC status: we explicitly do NOT mix gRPC errors and CSI errors.

jdef:
https://github.com/container-storage-interface/spec/issues/23#issuecomment-316528952

jdef:
#63

julian-hj:
Note that as the spec is written today, if we use the CSI error and not the gRPC error, then CreateVolume cannot be idempotent as claimed.

The reason is that in the response, either a result or an error can be returned, but not both.  So if CreateVolume gets called twice, there is no way for the plugin to return `VOLUME_ALREADY_EXISTS` and also return `volume_info`.  As a result, the CO will know that the call succeeded, but lack the required information to do anything with that volume (including deleting it.)

I think we should either make use of gRPC errors, or allow for returning status codes and response information in the same reply.

```
message CreateVolumeResponse {
  message Result {
    // Contains all attributes of the newly created volume that are
    // relevant to the CO along with information required by the Plugin
    // to uniquely identifying the volume. This field is REQUIRED.
    VolumeInfo volume_info = 1;
  }

  // One of the following fields MUST be specified.
  oneof reply {
    Result result = 1;
    Error error = 2;
  }
}
```

jdef:
Although we've decided to use gRPC for CSI I think we should carefully consider the impact to "protocol adapters" each time we consider more tightly coupling the CSI spec to gRPC. For example, there's a standard, default status/error translator for the gRPC gateway service (that adapts REST-like RPCs to/from gRPC). That translator doesn't seem to include gRPC status `Details` in the generated response entity. https://github.com/grpc-ecosystem/grpc-gateway/blob/7195ea445241f9773e71fc7b94c306d3ed6b35b6/runtime/errors.go#L92

[...]

PR #123, opened 2017-10-20
Refactor VolumeHandle into id and vol attributes

### LINKED ISSUE #116: Remove Metadata from VolumeHandle

jdef:
* Incorporates the `VolumeHandle` changes from #117 (thanks @akutz!), which eliminates `metadata` and defines the unique volume identifier as a string`id`.
* Defines `attributes` for a `VolumeInfo` and passes them through to publish and validation requests.

Fixes #116


jdef:
If/when this lands, let's "rebase and merge" to retain the commit stream

jdef:
xref #118

akutz:
Hi @jdef,

I'm not a fan of separating the ID and attributes. I personally like the `VolumeHandle` you created. If attributes are necessary to perform some type of operation on a Volume, then I think they should be included in the same structure as the ID. The `VolumeHandle` was perfect for this IMO. The question all along was the necessity of the metadata/attributes. Since they've been deemed necessary I believe they should be part of the `VolumeHandle`, just not part of the identity. Also, what's wrong with the name `Metadata` or even `Tags`? The former is consistent with gRPC naming conventions and the latter is widely used.

jdef:
There was a follow up community meeting where the fate of metadata was
decided. Basically, it shouldn't contain sensitive data and wasn't
necessary for all calls. I renamed it because I thought it would bring
clarity of purpose.

Let the bikeshedding on naming begin! :)

akutz:
When was the follow up meeting? I wasn’t aware it had occurred yet.

csi.proto
```diff
@@ -245,31 +245,25 @@ message VolumeInfo {
...
-  VolumeHandle handle = 2;
+  // subsequent calls to refer to the provisioned volume. This field
+  // should not exceed 1MiB.
+  string id = 2;
+
+  // Attributes reflect static properties of a volume and MUST be passed
+  // to volume validation and publishing calls. Attributes MAY NOT
+  // uniquely identify a volume. A volume uniquely identified by `id`
+  // SHALL always report the same attributes. This field is REQUIRED.
+  Attributes attributes = 3;
```
akutz: Hi @jdef,

It doesn't make sense to me that `Attributes` should be a discreet message type when it could simply be defined like so:

```proto
map<string,string> attributes = 3;
```

I dislike the trend in this spec to wrap single values.

jieyu: +1

I don't have a strong opinion, but given that `parameters` is in `map<string, string>`, i'd prefer use do `map<string,string> attributes = 3;` here.

jieyu:
@saad-ali @julian-hj can you please take a look at this PR? It looks good to me, except for the naming part. But I am fine with any naming.

csi.proto
akutz: FWIW, my opinion **is** strong. I prefer consistency :)

julian-hj: +1

csi.proto
```diff
@@ -248,6 +248,22 @@ message VolumeInfo {
   // subsequent calls to refer to the provisioned volume. This field
   // should not exceed 1MiB.
   string id = 2;
+
+  // Attributes reflect static properties of a volume and MUST be passed
+  // to volume validation and publishing calls. Attributes MAY NOT
+  // uniquely identify a volume. A volume uniquely identified by `id`
+  // SHALL always report the same attributes. This field is REQUIRED.
```
julian-hj: During the meeting, I think we said that `attributes` was required to be passed by the CO iff it is returned by the SP.  We seem to be losing that nuance in this spec and declaring it as required always.  Does that mean that we will always require the SP to return a map, and in the case where the SP doesn't care about parameters, it should just return an empty map?  I guess I am OK with that requirement, but just want to make sure we are all on the same page.

akutz: Hi @julian-hj,

If it's on the SP to indicate if the data is required it can return an error if an empty map is transmitted, right? If you're just wanting to clarify the spec then I agree it needs to have tighter language.

julian-hj:
LGTM, with the caveats that I agree with @akutz about the `Attributes` type being unnecessary, and I think we need to make it clearer that although `attributes` is required in the CreateVolume response, it is OK for it to be empty.

csi.proto
jieyu: I think in @jdef's PR, it's a struct `Attributes`. So it's required to set this field, but the map inside the struct can be empty. If we change this to `map<string, string>`, we need to clarify in the wording. This field is indeed optional, but if set, the CO should pass that to the subsequent calls.

julian-hj: @jieyu Ah, that makes sense.  So we didn't lose the nuance, we just codified it as an optional map inside a required struct.  Makes sense.  I still think we'd be better off with a plain map & some explicit language about the behavior though.

jdef:
Addressed feedback but left it as a separate commit for easier review. If it looks good, I'll squash it with the prior commit (but NOT squash the whole PR). And we can "rebase and merge" when ready.

csi.proto
```diff
@@ -245,31 +245,19 @@ message VolumeInfo {
...
+
+  // Attributes reflect static properties of a volume and MUST be passed
+  // to volume validation and publishing calls.
+  // Attributes SHALL be opaque to a CO. Attributes SHALL NOT be mutable
+  // and SHALL be safe for the CO to cache. Attributes SHOULD NOT
+  // contain sensitive information. Attributes MAY NOT uniquely identify
+  // a volume. A volume uniquely identified by `id` SHALL always report
+  // the same attributes. This field is OPTIONAL and when present MUST
+  // be passed to volume validation and publishing calls.
+  map<string,string> attributes = 3;
```
jdef: I guess we should probably have a limit for the encoded size of this field. 1MiB, or something else?

akutz:
Lord I regret ever filing #112. I sure liked `VolumeHandle` for what it was. Oh well. In lieu of that, this LGTM :) Great work @jdef! Thank you for taking this on!

jdef:
Also, should we have an INVALID_VOLUME_ATTRIBUTES error code for the related RPCs?

akutz:
Hi @jdef,

I'd be hesitant to add any new errors until @saad-ali completes #18.

jdef:
ready to "rebase and merge" once CI completes

PR #129, opened 2017-10-24
spec: Simplify PublishVolumeInfo

akutz:
This patch simplifies `PublishVolumeInfo` by removing it as a distinct message and refactoring its references to `map<string, string>`. This change is consistent with other locations in the spec where a map of strings is not the sole field in an outer message.

cc @jieyu @jdef @saad-ali


akutz:
Hey, three approvals!

image

Issue #102, opened 2017-09-12
ProbeController rpc on Controller service


jeffpak:
Hello,

We are wondering why there is no dedicated RPC for probing the controller's readiness, when there is one for the node. We noticed it was difficult to determine whether or not we had a functioning controller until we started creating volumes, which seems a bit late in the lifecycle. It would be great to have a dedicated probe request for the controller in addition to the node.

Looking forward to your thoughts.
-Diego Persistence Team for Cloud Foundry

akutz:
Hi @jeffpak,

Hmmm. This is curious. While some storage platforms enable centralized creates with no host dependencies, there are still yet those that *do* require host dependencies, especially with regards to potential configuration and permissions issues. Imagine a virtual filesystem (VFS) plug-in that lacks permissions to write to the local filesystem? Or an Amazon Elastic Block Service (EBS) plug-in that lacks the configured credentials or role to create new volumes?

Today any of the above errors could be handled by the Controller plug-in failing to start and providing some error to the CO.

However, if that's the intended behavior it begs the question why not follow the same pattern for the Node?

I agree that the Controller and Node should either:

* Both implement a `Probe` RPC to indicate if the service is configured correctly and has the required dependencies.
* Both return an error on start to the CO with the same result that would otherwise be returned from their respected `Probe` RPCs.

I personally like the `Probe` approach as it follows the pattern of a type registry where a constructor should never return an error and some associated initialization call is what is allowed to return the failed state based on config/dependency issues. This way it's at least possible to validate a plug-in can load and be recognized by the CO even if the plug-in is not properly configured. That has value for dev and testing.

jdef:
I tend to agree: consistency is important. Thanks for this proposal. Would like to hear from some others @jieyu @saad-ali @cpuguy83 @julian-hj

julian-hj:
+1 to add a ProbeController method

saad-ali:
SGTM

j-griffith:
Seems like it would be easy enough to implement and certainly useful IMO

cpuguy83:
SGTM

Kaixiang:
Hi,

This is Kai from Diego Persistence Team for Cloud Foundry,
I've submitted a PR implementing this, Could you help review it?

Regards

jdef:
[...]

PR #130, opened 2017-10-26
Add ProbeController rpc on controller service

Kaixiang:
Hi:

This is a PR trying to add ProbeController RPC on controller service.
which has been discussed [here](https://github.com/container-storage-interface/spec/issues/102).

Could you help review it?

Regards
Kai


spec.md
```diff
@@ -1711,6 +1761,7 @@ message Error {
     NodeUnpublishVolumeError node_unpublish_volume_error = 8;
     ProbeNodeError probe_node_error = 9;
     GetNodeIDError get_node_id_error = 10;
+    ProbeControllerError probe_controller_error = 11;
```
akutz: Hi @Kaixiang,

I wonder if this should be reordered so that it's grouped with the other Controller RPCs. Yes, it will be a breaking change due to the field order, but there's no tagged version yet, so I think that is fine.

spec.md
```diff
@@ -864,6 +867,32 @@ message GetCapacityResponse {
 }
 ```
 
+#### `ProbeController`
+
+A Controller Plugin MUST implement this RPC call.
+The Plugin SHALL assume that this RPC will be executed on the controller where the volume will be created.
+The CO SHOULD call this RPC for the controller at which it wants to create volume.
+This RPC allows the CO to probe the readiness of the Plugin on the controller where the volumes will be created.
+The Plugin SHOULD verify if it has everything it needs (binaries, kernel module, drivers, etc.) to run on that controller, and return a success if the validation succeeds.
```
akutz: Hi @Kaixiang,

Because the Controller service is likely to be where any centralized configuration lives, I wonder if there should be some text about verifying expected configuration files/settings?

spec.md
```diff
@@ -864,6 +867,32 @@ message GetCapacityResponse {
 }
 ```
 
+#### `ProbeController`
+
+A Controller Plugin MUST implement this RPC call.
+The Plugin SHALL assume that this RPC will be executed on the controller where the volume will be created.
```
jdef: There's a bunch of wording here that doesn't make sense for a controller service since it could be either a central service, or a distributed one. Want to take a stab at hacking the wording a bit? If not, I'll happily make a suggestion.

akutz:
Hi @jdef,

> Also, we now have RPCs that read like Verb{Controller,Node} and {Controller,Node}Verb{something}. It's not consistent. Would this be a good time to rename the ProbeXyz RPCs to ControllerProbe and NodeProbe for consistency w/ the other RPCs? If we're going to tackle a rename, we might as well do it before this spec becomes versioned. What do others think?

I agree, but I don't think this is the right place to make that decision. I'd argue in favor of letting this PR stand consistent with the existing `ProbeNode` RPC and then do a follow-up PR where both are renamed at the same time.

Thoughts?

jdef:
I'd be OK w/ a follow-up PR to address renaming

akutz:
Hi @jdef,

Just so we could target both `ProbeController` and `ProbeNode` at once. It feels like there should be a single commit that adds `ProbeController`. Heck, I'm fine if it's all in this PR but separated into two commits:

1. Add `ProbeController`
2. Rename `ProbeController` and `ProbeNode`

jdef:
Yes, at least separate commits (if not a new PR)

Kaixiang:
That was fast, Thanks for all the reviews!
I've reworded the description a bit about `ProbeController` and also renamed the `ProbeController` and `ProbeNode` to be more consistent in another commit.

See if it works.

spec.md
```diff
@@ -864,6 +867,31 @@ message GetCapacityResponse {
 }
 ```
 
+#### `ProbeController`
+
+A Controller Plugin MUST implement this RPC call.
+The Plugin SHALL assume that this RPC will be executed on the controller which could be either a central service or a distributed one.
```
jdef: I'd rather just replace this line with something a bit different:
The CO SHALL invoke this RPC prior to any other controller service RPC in order to allow the CO to determine the readiness of the controller service.
A CO MAY invoke this call multiple times with the understanding that a plugin's implementation MAY NOT be trivial and there MAY be overhead incurred by such repeated calls.

spec.md
```diff
@@ -864,6 +867,31 @@ message GetCapacityResponse {
...
+The CO SHOULD call this RPC for the controller at which it wants to create volume.
```
jdef: > The CO SHOULD call...

This line is awkward. Should we just remove it?

spec.md
```diff
@@ -864,6 +867,31 @@ message GetCapacityResponse {
...
+This RPC allows the CO to probe the readiness of the Plugin on the controller.
```
jdef: This line becomes redundant with the above suggested change

spec.md
```diff
@@ -864,6 +867,31 @@ message GetCapacityResponse {
...
+The Plugin SHOULD verify if it has the right configurations to run the controller, and return a success if the validation succeeds.
```
jdef: I'd bump this up to be the 2nd line of the paragraph

jdef: s/configurations/configurations, devices, dependencies and drivers in order/
s/controller/controller service/

cpuguy83:
Since this is essentially a heathcheck, I'm wondering if following a known grpc pattern for this is a better approach?
For reference: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

It would be good to include this on both controller and node components.

jdef:
This has come up before. It is NOT the intent of this RPC to form the basis
of a periodic health check: a CO is only guaranteed to invoke the plugin
xxxProbe functions ONCE, at startup time. A CO *may* invoke it multiple
times, but that's completely optional. The xxxProbe functions *may* be very
expensive to execute, unlike what's typically implemented for a health
check.

If we want a health check API it should be its own RPC, not part of
xxxProbe.

akutz:
Hi @jdef,

I concur with regards to the intent of the Probe functions. I've implemented such functions for 10+ storage platforms, and while the cost ranges from cheap to expensive, there *is* always a cost. Additionally, to me `Probe` doesn't mean the service is "ready":

> The Plugin SHOULD verify if it has everything it needs (binaries, kernel module, drivers, etc.) to run on that node, and return a success if the validation succeeds. 

So again, to reiterate what James said: the Probe functions indicate if the SP is supported on a host -- not necessarily if the SP's services are "ready" to receive requests.

jdef:
ping @julian-hj @saad-ali

cpuguy83:
Ok I seem to recall having this conversation now. LGTM

PR #131, opened 2017-10-30
spec: use official proto wrappers for primitive fields

### LINKED ISSUE #22: Wrapping OPTIONAL fields in proto3.
In proto3 we no longer have a `has_field()` function if `field` is of a primitive type, so we may want to wrap them. Here is a list of OPTIONAL fields that have this problem:

* `MountVolume::fs_type`: If an empty string is allowed, then we should wrap it. Otherwise we can explicitly specify that it must not be empty when set.
* `ListVolumeRequest::starting_token` and `ListVolumeResponse::Result::next_token`: Similarly, is an empty string a valid token?
* `ListVolumeRequest::max_entries`: Can we issue a request to return 0 entries?

Here is a list of OPTIONAL fields that may not need to be wrapped:
* `GetPluginInfoResponse::manifest`: `map<string, string>`
* `CreateVolumeRequest::parameters`: `map<string, string>`
* `VolumeMetadata::values`: `map<string, string>`
* `ValidateVolumeCapabilitiesResponse::Result::message`: `string`
* All OPTIONAL repeated fields

NOTE, the OPTIONAL field `VolumeInfo::capacity_bytes` requires no wrapping because we specify in the spec that it must be non-zero when set.

jdef:
* updated spec uses official protobuf wrapper types
* updated golang build file mangles path for wrappers.proto  
* fixes #22


jdef:
I'm not very happy w/ the Makefile updates. I couldn't think of a better way to manage the path to wrappers.proto. It doesn't make sense to include the path to the golang version of wrappers in the official spec because the official spec should be language neutral. Looking at the standard version of wrappers.proto that ships w/ protoc, it's clear that the golang package for file is expected to live in the `golang/protobuf` tree. Furthermore, when NOT doing the mangle step the `go get ...` step fails because the official `google/protobuf` repo doesn't have any golang files in `src/google/protobuf` (where the upstream wrappers.proto lives).

Suggestions welcome

jdef:
I should add, there's also the possibility of vendoring the wrappers.proto file that we're using, and then referring to the vendored copy. We're not vendoring anything yet, so I'm hesitant to open that can of worms simply for this change set.

jdef:
The OOB feedback that I've received so far is that these changes are too aggressive: only the OPTIONAL fields should use the wrapper types.

jdef:
Discussed in the CSI WG meeting, determined that default values are "good enough" signals and that we simply need stronger language in the spec regarding the use of defaults to serve as signals for unspecified fields. The wrappers can be useful but are a bit unwieldy to deal with, depending on the programming language of the generated code.

PR #132, opened 2017-11-01
spec: default field sizes; clarify OPTIONAL fields

### LINKED ISSUE #22: Wrapping OPTIONAL fields in proto3.

### LINKED ISSUE #124: spec: clarify expectations and reduce recommended size of volume ID
Volume IDs uniquely identify a volume. They should generally be human readable and useful for logging (debugging and tracing) purposes. The spec defines the type of `id` as string, which is fine. There's some debate around what the size of an ID should be.

This issue proposes that the CSI spec recommends a size limit of the protobuf-encoded volume `id` field to 256 bytes. Such a size provides ample space for uniqueness and also represents the upper length limit of a token that's useful w/ respect to logging and readability.

jdef:
As discussed in the CSI WG meeting on 2017-10-31, a "general field size" clause has been added to the spec. Also, language clarifying the intent and purpose of zero-values for OPTIONAL fields has been added to fields as appropriate.

* supersedes #131
* fixes #22
* fixes #124


jdef:
No changes for OPTIONAL error code fields since revisions for the error code section are WIP.

spec.md
```diff
@@ -295,6 +295,18 @@ However, in some circumstances, the CO may lose state (for example when the CO c
...
+
+CSI defines general size limits for fields of various types (see table below).
+The general size limit for a particular field may be overridden by specifying a different size limit in said field's description.
+The general size limits apply for messages generated by both COs and plugins.
+
+| Size       | Field Type          |
+|------------|---------------------|
+| 128 bytes  | string              |
+| 4 KiB      | map<string, string> |
+| 4 KiB      | repeated string     |
```
jieyu: As part of the change you did in #133, `node_ids` will be `repeated string` as well. If we limit this, we'll introduce a limit for the number of nodes this spec can handle. I'd suggest we remove this limitation here. Instead, explicitly mention that `mount_flags` has a limit of 4K bytes.

spec.md
```diff
@@ -527,24 +541,25 @@ message VolumeCapability {
...
-  // set, it indicates that the capacity of the volume is unknown (e.g.,
-  // NFS share). If set, it MUST be non-zero.
+  // set (value of 0), it indicates that the capacity of the volume is
+  // unknown (e.g., NFS share).
   uint64 capacity_bytes = 1;
 
   // Contains identity information for the created volume. This field is
   // REQUIRED. The identity information will be used by the CO in
-  // subsequent calls to refer to the provisioned volume. This field
-  // should not exceed 1MiB.
```
akutz: Hi @jdef,

I just want to be clear that the ID string previously would be allowed a field size of 1MiB, but according to the spec today the volume ID is now limited to 128 bytes. Is that correct?

jdef:
Yes. I forget where it was discussed (CSI WG sync, CSI Community sync,
both?) but the consensus was that 1MiB was probably overkill for an ID
field - that we should start small and if we find that it's too limiting we
can change the spec later to allow for longer IDs.

On Fri, Feb 2, 2018 at 5:42 PM, Schley Andrew Kutz <notifications@github.com

PR #139, opened 2017-11-06
spec: Consistency updates, Error Enum Fix, Spelling Fix

akutz:
This PR contains the following patches:

* Updates instances of `user_crendentials` to no longer use the message `Credentials`, but instead be direct instances of `map<string, string>`. This change is consistent with other, similar changes made to like-fields.
* Corrects the `ValidateVolumeCapabilities` error enum value `INVALID_VOLUME_INFO` to be `INVALID_VOLUME_ID`
* Corrects the misspelled word "accessable"  |

I feel strongly about the credentials change as it:
* Updates the field to be consistent with other, similar fields
* Makes working with the field in a language binding more straight-forward

cc @jdef @jieyu @saad-ali


akutz:
Given the timeframe left to make changes, could people please take a look when they're able? Thanks!

cc @cpuguy83 @jdef @jieyu @saad-ali @julian-hj

akutz:
Hi @julian-hj,

Oh geez, did I really not remove that? What an idiot I can be sometimes. Thanks for catching that!

akutz:
Hi @jieyu and @julian-hj,

I've rebased this PR and removed the vestigial `Credentials` message from the spec. Thank you both for catching that.

csi.proto
```diff
@@ -152,8 +152,17 @@ message CreateVolumeRequest {
...
+  // password. Each key must consist of alphanumeric characters, '-',
+  // '_' or '.'. Each value MUST contain a valid string. An SP MAY
+  // choose to accept binary (non-string) data by using a binary-to-text
+  // encoding scheme, like base64. An SP SHALL advertise the
+  // requirements for credentials in documentation. COs SHALL permit
+  // users to pass through the required credentials. This information is
+  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
   // This field is OPTIONAL.
-  Credentials user_credentials = 6;
+  // The total bytes this field must be less than 1 Mebibyte.
```
jdef: do we really need to limit this to 1MiB? there's wording in the spec now that declares a default limit for fields declared this way. in this case, the field would be limited by default to 4KiB

jdef: xref #144

PR #143, opened 2017-11-08
spec: Clarify next_token in ListVolumes

### LINKED ISSUE #138: ListVolumesResponse should return an ID
The current model which provides a StartToken on ListVolumesRequest does not provide enough information if there are multiple clients to the CSI driver. For this reason ListVolumesResponse should return an ID that is passed in with StartToken to identify the caller.

Unless grpc provides a method to get some type of id from the connection?

lpabon:
Closes #138

Signed-off-by: Luis Pabón <luis@portworx.com>

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -475,8 +475,10 @@ message ListVolumesResponse {
     // This token allows you to get the next page of entries for
     // `ListVolumes` request. If the number of entries is larger than
     // `max_entries`, use the `next_token` as a value for the
-    // `starting_token` field in the next `ListVolumes` request. This
-    // field is OPTIONAL.
+    // `starting_token` field in the next `ListVolumes` request. The
+    // Plugin my choose to use the token to contain all the necessary
+    // information to satisfy subsequent requests. This field
+    // is OPTIONAL.
     string next_token = 2;
   }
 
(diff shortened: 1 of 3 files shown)


csi.proto
```diff
@@ -475,8 +475,10 @@ message ListVolumesResponse {
     // This token allows you to get the next page of entries for
     // `ListVolumes` request. If the number of entries is larger than
     // `max_entries`, use the `next_token` as a value for the
-    // `starting_token` field in the next `ListVolumes` request. This
-    // field is OPTIONAL.
+    // `starting_token` field in the next `ListVolumes` request. The
+    // Plugin my choose to use the token to contain all the necessary
```
jdef: s/my/may/

I'm not sure I understand the need for this change. Maybe I'm too close to the spec? My understanding was that the statement added here is implicit since the plugin generates the tokens to begin with.

jdef: In other words, in what case would a plugin NOT generate a token that's sufficient to serve subsequent requests?

akutz: I agree with you James. I’m not sure there needs to be clarification. There’s a reason it’s “string next_token” and not “unit64 next_page”. The token is whatever the SP wants it to be and opaque to the client.

lpabon: Yep, I agree that both of you are too close to the spec :smile: 

As a driver developer, this text clears up how to use it.

lpabon:
i thought about it and is really fine how it is. So I will close it.

jdef:
If you think it would help clarify things, we've been thinking about
creating a workflows/ directory in the community project:
https://github.com/container-storage-interface/community

Maybe it would be helpful for people if there were a
workflows/{storage-type}/list-volumes.md, where {storage-type} is something
like local, nfs, cifs, iscsi or whatever you're working on?

PR #144, opened 2017-11-09
spec: credentials have a default 4KiB size limit

jdef:
follow-up to #132 after #139 already landed


akutz:
I approve this message. :)

jdef:
rebased and force-pushed

PR #146, opened 2017-11-10
Fix ControllerPublish/Unpublish error codes

jsafrane:
- ControllerUnpublishVolume must be idempotent and thus return no error instead of `VOLUME_NOT_ATTACHED_TO_SPECIFIED_NODE`
- `VOLUME_ALREADY_ATTACHED` does not exists, use `VOLUME_ALREADY_PUBLISHED` instead.


jsafrane:
BTW, a standard gRPC error codes as suggested in #115 would be *much* easier to use. Close this if #115 is merged.

jsafrane:
closing in favor of #115

PR #148, opened 2017-11-16
Make node_id required in GetNodeIDResponse

saad-ali:
* Make `node_id` a required field in the `GetNodeIDResponse` if the plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability.
* Make `node_id` a required field in the `ControllerPublishVolume` and `ControllerUnpublishVolume` calls.
* Require CO to pass it's `node_id` in `GetNodeIdReqeust` so that SP's that don't have their own custom internal naming scheme do not have to implement complicated logic to figure out the node name. 

Fixes https://github.com/container-storage-interface/spec/issues/121

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -284,10 +284,8 @@ message ControllerPublishVolumeRequest {
   // This field is REQUIRED.
   string volume_id = 2;
 
-  // The ID of the node. This field is OPTIONAL. The CO SHALL set (or
-  // clear) this field to match the node ID returned by `GetNodeID`.
-  // `GetNodeID` is allowed to omit node ID from a successful response;
-  // in such cases the CO SHALL NOT specify this field.
+  // The ID of the node. This field is REQUIRED. The CO SHALL set this
+  // field to match the node ID returned by `GetNodeID`.
   string node_id = 3;
 
   // The capability of the volume the CO expects the volume to have.
@@ -332,15 +330,11 @@ message ControllerUnpublishVolumeRequest {
   // The ID of the volume. This field is REQUIRED.
   string volume_id = 2;
 
-  // The ID of the node. This field is OPTIONAL. The CO SHALL set (or
-  // clear) this field to match the node ID returned by `GetNodeID`.
-  // `GetNodeID` is allowed to omit node ID from a successful response;
-  // in such cases the CO SHALL NOT specify this field.
-  //
-  // If `GetNodeID` does not omit node ID from a successful response,
-  // the CO MAY omit this field as well, indicating that it does not
-  // know which node the volume was previously used. The Plugin SHOULD
-  // return an Error if this is not supported.
+  // The ID of the node. This field is OPTIONAL. The CO SHOULD set this
+  // field to match the node ID returned by `GetNodeID` or leave it
+  // unset. If the value is set, the SP MUST unpublish the volume from
+  // the specified node. If the value is unset, the SP MUST unpublish
+  // the volume from all nodes it is published to.
   string node_id = 3;
 
   // End user credentials used to authenticate/authorize controller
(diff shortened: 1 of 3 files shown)


csi.proto
```diff
@@ -332,15 +330,8 @@ message ControllerUnpublishVolumeRequest {
...
-  // clear) this field to match the node ID returned by `GetNodeID`.
-  // `GetNodeID` is allowed to omit node ID from a successful response;
-  // in such cases the CO SHALL NOT specify this field.
-  //
-  // If `GetNodeID` does not omit node ID from a successful response,
-  // the CO MAY omit this field as well, indicating that it does not
-  // know which node the volume was previously used. The Plugin SHOULD
-  // return an Error if this is not supported.
+  // The ID of the node. This field is REQUIRED. The CO SHALL set this
+  // field to match the node ID returned by `GetNodeID`.
```
jdef: We killed a use case here by making this REQUIRED: a CO may intentionally leave this field blank to indicate to the plugin that the volume should be unpublished from every node in the cluster. Is killing that use case really our intent here?

csi.proto
```diff
@@ -576,13 +567,18 @@ message NodeUnpublishVolumeResponse {}
 message GetNodeIDRequest {
   // The API version assumed by the CO. This is a REQUIRED field.
   Version version = 1;
+
+  // The ID of the node as understood by the CO which MAY be returned by
+  // the SP in the GetNodeIDResponse if the SP does not implement a
+  // custom node naming scheme and understands (or internally maps) the
+  // CO's node id.
```
jdef: REQUIRED? OPTIONAL?

csi.proto
jieyu: Yeah. I think we deliberately made this field optional to capture the case where the CO just want to detach the volume from any node.  I still believe this is useful in some reconciliation scenarios.

saad-ali: Ack. I'll restore this then.

csi.proto
saad-ali: Will add.

saad-ali:
@jdef @jieyu feedback addressed, PTAL

spec.md
```diff
@@ -1132,13 +1126,18 @@ The result of this call will be used by CO in `ControllerPublishVolume`.
 message GetNodeIDRequest {
   // The API version assumed by the CO. This is a REQUIRED field.
   Version version = 1;
+
+  // The ID of the node as understood by the CO which MAY be returned by
```
jieyu: cc @akutz 

This is still controversial. Can we make this change a separate PR?

saad-ali: Done. Will pull this in to a separate PR.

saad-ali:
PTAL

spec.md
```diff
@@ -1135,10 +1129,9 @@ message GetNodeIDRequest {
 }
 
 message GetNodeIDResponse {
-  // The ID of the node which SHALL be used by CO in
-  // `ControllerPublishVolume`. This is an OPTIONAL field. If unset,
-  // the CO SHALL leave the `node_id` field unset in
-  // `ControllerPublishVolume`.
+  // The ID of the node as understood by the SP which SHALL be used by
+  // CO in subsequent `ControllerPublishVolume`.
+  // This is a REQUIRED field.
```
jdef: suggestion: This is an OPTIONAL field that is REQUIRED only if the controller advertises the `PUBLISH_UNPUBLISH_VOLUME` capability.

because not implementing the call means that GetNodeIDResponse is empty, and that violates the REQUIRED semantics as currently documented.

saad-ali: https://github.com/container-storage-interface/spec/pull/152

PR #152, opened 2017-11-17
Add a GetNodeID error

saad-ali:
Follow up to https://github.com/container-storage-interface/spec/pull/148

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1145,6 +1145,12 @@ message GetNodeIDResponse {
 ##### GetNodeID Errors
 
 If the plugin is unable to complete the GetNodeID call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1145,6 +1145,10 @@ message GetNodeIDResponse {
 ##### GetNodeID Errors
 
 If the plugin is unable to complete the GetNodeID call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+| Call not implemented | 12 UNIMPLEMENTED | GetNodeID call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` or `NodeGetCapabilities` to discover Plugin capabilities. |
```
jdef: I think this needs a table header? It doesn't render into a table when I view this "rendered" in GH.

saad-ali: Ah, good catch. Brain fart

saad-ali:
PTAL

PR #153, opened 2017-11-18
spec: clarify REQUIRED for map/repeated fields

### LINKED ISSUE #90: spec: This field is REQUIRED
I have a question regarding REQUIRED fields, for example:

```proto
  // The capabilities that the provisioned volume MUST have: the Plugin
  // MUST provision a volume that could satisfy ALL of the
  // capabilities specified in this list. The Plugin MUST assume that
  // the CO MAY use the  provisioned volume later with ANY of the
  // capabilities specified in this list. This also enables the CO to do
  // early validation: if ANY of the specified volume capabilities are
  // not supported by the Plugin, the call SHALL fail. This field is
  // REQUIRED.
  repeated VolumeCapability volume_capabilities = 4;
```

The field must be non-nil, but what if the the array has no elements? Does a zero-length array still meet the intent of REQUIRED? It's important as it affects the way that message validation is handled.

jdef:
Fixes #90


saad-ali:
LGTM

gpaul:
Cool, thanks.

PR #155, opened 2017-11-20
Clarify NodePublishVolume target_path semantics

### LINKED ISSUE #60: NodePublishVolume target_path needs additional clarifications
The spec says:

```
  // request. This is a REQUIRED field.
  string target_path = 5;
```
However, the spec doesn't describe:
1) Whether the plugin should expect the directory at the target_path to already be created by the CO or the plugin is responsible for creating the directory at target_path itself.
2) If the CO pre-creates the directory, what permissions the plugin can expect it to have.
3) If the plugin must create the directory, what permissions the parent path should have.
4) If more than one directory in the path must be created by the plugin, what permissions the parent directories should be given in order to not break other plugins that will perhaps use the same path.

Points 2 and 3 are mainly relevant for plugins that use fuse mounts and will not run with CAP_SYS_ADMIN.  One advantage to allowing the plugin to determine its own mountpoint is that the plugin can be responsible for making sure that it has create rights on the path where the mount points will go, even if it is running as some arbitrary non-root user.  

As the spec is written, the CO is responsible for dictating the mount path so I suppose that the mount points must be open for world write, or the CO needs some way to know what user the plugin will run as and `chown` the directory to that user.

julian-hj:
Fixes #60

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -511,6 +511,8 @@ message NodePublishVolumeRequest {
   // The path to which the volume will be published. It MUST be an
   // absolute path in the root filesystem of the process serving this
   // request. The CO SHALL ensure uniqueness of target_path per volume.
+  // The CO SHALL ensure that the path exists, and that the process
+  // serving the request has `read` and `write` permissions to the path.
   // This is a REQUIRED field.
   string target_path = 4;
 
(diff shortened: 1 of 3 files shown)


jdef:
@julian-hj please run `make` to regenerate the proto file based on your spec changes

julian-hj:
whups, I managed to convince myself that wasn't necessary, since the changes are all to verbiage and not code, but of course since they are in the comments on the protobuf snippet, that's totally not true.  Should be all set now.

saad-ali:
LGTM

PR #156, opened 2017-11-23
Go Bindings Canonical Import Path

### LINKED ISSUE #142: lang: Official CSI bindings should live in this repo
This repo currently has the following directory structure:

```
<root>
|
|--lib
   |
   |--cxx
   |
   |--go
```

The `lib/*` directories exist in the official repo as the result of a Slack conversation between myself, @jdef, @jieyu, and @saad-ali. We all agreed that the official repo should have some automated means via CI for validating changes to `spec.md`. The most obvious solution was generating a `csi.proto` protobuf file from `spec.md` and then generating language bindings from `csi.proto`. However, @saad-ali reminding us that there was consensus that this repo should not be host to source code, only `spec.md` -- the single truth when it comes to the CSI specification.

We agreed upon a compromise -- there *would* be language bindings generated in the spec repo, but only for the purpose of validating changes to `spec.md`. The reason the bindings are committed instead of occurring only as part of CI is so there can also be a committed version of `csi.proto` that is known to be valid and from which bindings can be generated.

Everything works.

Except...it kind of doesn't :(

I work on a project called [GoCSI](https://github.com/thecodeteam/gocsi) -- a CSI library, client, and other helpful utilities created with Go. That project's `Makefile` is able to pull `spec.md` from this repo, generate `csi.proto` and then generate Go language bindings. The Go language bindings for GoCSI live in the project's [root package `csi`](https://github.com/thecodeteam/gocsi/tree/master/csi).

People that use Go and wish to use GoCSI for working with CSI in Go are free to do so. People that use Go and wish to **not** use GoCSI are free to do that as well.

However, here is what people cannot do:

**People that use Go and wish to use GoCSI and some other package or project that imports different CSI bindings.**

Protobuf generated types are all registered in a global, static map via [`RegisterType(msg, name string)`](https://github.com/golang/protobuf/blob/master/proto/properties.go#L833-L844). If `RegisterType` is invoked more than once for the same type, the Go program will panic and exit with a stack trace.

Thus it is not possible for GoCSI to generate its own CSI bindings and for a program to both use GoCSI and some other library that also generates its own CSI bindings.

You may be thinking, "Okay, so I'll only use GoCSI, or I'll not use it and generate my own bindings."

That's fair, but the way protobuf works makes that a cumbersome proposition. Protobuf generates Go sources in such a way that it is not possible to use Go interfaces to abstract away the language bindings being used. That means that if someone wants to use the type `csi.VolumeInfo` (generated from the CSI spec) in their code there has to be a **single** package that defines that type. That is true for a Go program and any of the packages imported directly and both transitively.

There are two ways around this:
1. A library **can** use its own, custom CSI language bindings if the library also vendors `github.com/golang/protobuf/proto` and `google.golang.org/grpc`.
2. A single package containing CSI language bindings that all Go programs and other libraries reference as the source of truth for CSI types.

The first option is not precluded by the second, but the second **is** necessary if Go developers are to easily consume and develop for CSI.

I propose that either the `lib/go/csi` package in this repo become the official source for CSI language bindings and types or we deem some other package as such.

Thoughts?

cc @saad-ali @jdef @jieyu

akutz:
This patch fixes #142 and adds a Go canonical import path of "github.com/container-storage-interface/spec/lib/go/csi" to the CSI Go bindings.

@jdef @saad-ali

### DIFF
--- a/lib/go/Makefile
+++ b/lib/go/Makefile
@@ -80,11 +80,13 @@ CSI_GO := csi/csi.pb.go
 CSI_A := csi.a
 CSI_PROTO := ../../csi.proto
 CSI_GO_TMP := csi/.build/csi.pb.go
+CSI_CANON_IMPORT := github.com/container-storage-interface/spec/lib/go/csi
 
 # This recipe generates the go language bindings to a temp area.
 $(CSI_GO_TMP): $(CSI_PROTO) | $(PROTOC) $(PROTOC_GEN_GO)
 	@mkdir -p "$(@D)"
 	$(PROTOC) -I "$(<D)" --go_out=plugins=grpc:"$(@D)" "$<"
+	sed -i -e 's|package csi|package csi // import "$(CSI_CANON_IMPORT)"|g' "$@"
 
 # The temp language bindings are compared to the ones that are
 # versioned. If they are different then it means the language
--- a/lib/go/csi/csi.pb.go
+++ b/lib/go/csi/csi.pb.go
@@ -47,7 +47,7 @@ It has these top-level messages:
 	NodeGetCapabilitiesResponse
 	NodeServiceCapability
 */
-package csi
+package csi // import "github.com/container-storage-interface/spec/lib/go/csi"
 
 import proto "github.com/golang/protobuf/proto"
 import fmt "fmt"


akutz:
I'm closing this and we should abandon the idea of canonical imports in #142 as they are not compatible with vendored dependencies. Instead Golang [dep](https://github.com/golang/dep) should be used to enforce the official CSI bindings.

cpuguy83:
There's a couple of ways we can deal with this.
If we specify `go_package` in the protobufs this should give us a canonical import.
Alternatively I believe we only need the canonical import path in one file in the package, in which case we can just place that in something like `lib/go/csi/imports.go` or something with just a package delcaration with the canonical import path.

However I would prefer if we don't modify the generated proto files after the fact.

saad-ali:
We don't have resources to implement a redirect service at this time, and that would be required with this change. Considering 1.0 is coming up, this is not a priority.

Issue #166, opened 2018-01-11
CSI support for hypervisor container runtimes 


resouer:
We would like to propose a meaningful feature to CSI spec, which aims at leveraging hypervisor based container runtimes e.g. (KataContainers, virtlet, KubeVirt etc) to use CSI in the future.

- The aim is to make it possible for runtimes like [KataContainers](https://katacontainers.io/) to bypass the attach phase and go to mount phase directly, and then, Kata will mount a block device to the VM-based-pod directly, instead of doing bind mount which is much slower in hypervisor case.
- Currently, we (Miratis, Hyper etc) are using flexvolume as workaround, e.g. https://github.com/kubernetes/frakti/blob/master/pkg/flexvolume/flexvolume.go While this solution is not portable and can not serve any general purpose since it should be bound with specific plugin (e.g. Cinder etc).
- This feature is also in the scope of Secure Runtime feature in sig-node's Q1 plan (p0). We already integrated Kata with CRI and CNI. And CSI will help us a lot to integrate Kata with containerd cri-o etc.
- We've discussed this in sig-storage meeting as a start and got positive feedback.

To serve the minimal purpose, only a minor change is expected from CSI side, please refer this slides for details:

https://docs.google.com/presentation/d/1kPeia7wLqoKQI0oX4pvVdH1UpcPx3lpmFK4P_E6oiIc/edit#slide=id.p

The pseudo code of CSI change is here: https://github.com/bergwolf/spec/tree/detached_volume

We can of course schedule meeting or talk in next sync for future discussion, while this issue can be used as feature tracker.

cc Kata maintainers @bergwolf @sameo @gnawux
cc sig-storage @saad-ali @jingxu97
cc CSI @jieyu 
cc RH: @rootfs Miratis: @ivan4th

rootfs:
Instead of changing API fingerprint, making new APIs make easier for compatibility purpose.

In our use case, we also need Node Reserve/Release Volume to ensure volumes are only used by one node, if these volumes don't support multi-attach. I believe this also helps Kata container.

cc @fabiand

fabiand:
Please also take into account that delegating the whole volume consumption to the hypervisor runtime is also benfitial - i.e. if we want to let qemu directly connect to the iSCSI target.

Not sure if this issue should serve just the original request, or also the one from @rootfs and mine (which are all three different ones).

bergwolf:
[...]

PR #169, opened 2018-01-18
NodeStageVolume/NodeUnstageVolume

davidz627:
Fixes Issue: #119 

Adds two new RPCs, NodeStageVolume/NodeUnstageVolume. These RPCs MUST be called by the CO if the Node Plugin advertises the STAGE_UNSTAGE_VOLUME capability. Plugins that advertise this capability SHOULD defer volume reference counting to the CO.

This PR supersedes #122, @akutz gets credit for the majority of the code, I'm just picking up where he left off and pushing this through.

cc @saad-ali @jieyu @clintkitson @jdef


davidz627:
Still taking suggestions on naming. I have nothing against NodePublishDevice/NodeUnpublishDevice, I just think it could be better.

I personally like: *NodeStageVolume/NodeUnstageVolume*. The naming is consistent with current calls as the “Node” part refers to the scope of the call and the “Volume” part follows the existing convention. “Stage” is the action word here that refers to the fact that this call is not finally publishing the volume for usage, but is performing an intermediary staging step to make it available for publishing. I am open for other suggestions on the naming.

Other suggestions include:
*NodePrepareVolume/NodeReleaseVolume*
[others?]

spec.md
```diff
@@ -1021,6 +1056,104 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
 
 ### Node Service RPC
 
+#### `NodePublishDevice`
+
+A Node Plugin MUST implement this RPC call if it has `PUBLISH_UNPUBLISH_DEVICE` node capability.
+This RPC is called by the CO when a workload that wants to use the specified volume is placed (scheduled) on a node.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+This RPC MUST be called by the CO a maximum of once per node, per volume.
```
jdef: suggest: s/MUST/SHOULD/ -- the RPCs are idempotent, so it's possible that this RPC would be executed more than once.

csi.proto
```diff
@@ -325,7 +331,8 @@ message ControllerPublishVolumeRequest {
 
 message ControllerPublishVolumeResponse {
   // The SP specific information that will be passed to the Plugin in
-  // the subsequent `NodePublishVolume` call for the given volume.
+    // the subsequent `NodePublishDevice` and `NodePublishVolume` calls
```
jdef: nit: spacing

csi.proto
```diff
@@ -496,6 +503,51 @@ message ControllerServiceCapability {
...
+  // left unset if the corresponding Controller Plugin does not have
+  // this capability. This is an OPTIONAL field.
+  map<string, string> publish_volume_info = 3;
+
+  // The path to which the volume will be published. It MUST be an
+  // absolute path in the root filesystem of the process serving this
+  // request. The CO SHALL ensure uniqueness of global_target_path per
+  // volume.
+  // This is a REQUIRED field.
+  string global_target_path = 4;
```
jdef: naming: if we renamed the RPCs to the "staging" variants (e.g. NodeStageVolume) then maybe we could call this the `staging_target_path`? "global" seems a little far-reaching

csi.proto
```diff
@@ -41,6 +41,12 @@ service Controller {
 }
 
 service Node {
+  rpc NodePublishDevice (NodePublishDeviceRequest)
```
jdef: naming: I prefer NodeStageVolume and NodeUnstageVolume

csi.proto
```diff
@@ -510,21 +562,29 @@ message NodePublishVolumeRequest {
...
+  string global_target_path = 4;
+
   // The path to which the volume will be published. It MUST be an
   // absolute path in the root filesystem of the process serving this
   // request. The CO SHALL ensure uniqueness of target_path per volume.
   // The CO SHALL ensure that the path exists, and that the process
   // serving the request has `read` and `write` permissions to the path.
   // This is a REQUIRED field.
-  string target_path = 4;
+  string target_path = 5;
```
jdef: should we avoid renumbering proto fields now that we have a tagged/published version of the spec in the world?

csi.proto
```diff
@@ -555,10 +615,18 @@ message NodeUnpublishVolumeRequest {
...
+  // It MUST be set if the Node Plugin implements the 
+  // `PUBLISH_UNPUBLISH_DEVICE` node capability.
+  // This is an OPTIONAL field.
+  string global_target_path = 3;
+
   // The path at which the volume was published. It MUST be an absolute
   // path in the root filesystem of the process serving this request.
   // This is a REQUIRED field.
-  string target_path = 3;
+  string target_path = 4;
```
jdef: ditto re: renumbering

csi.proto
```diff
@@ -496,6 +503,51 @@ message ControllerServiceCapability {
 }
 ////////
 ////////
+message NodePublishDeviceRequest {
```
jdef: 1. should this message include the volume "attributes", like the other publish calls?
2. should this message include staging credentials? (plugin can authn/z the RPC call)

spec.md
```diff
@@ -1021,6 +1056,104 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+| Operation pending for volume | 10 ABORTED | Indicates that there is a already an operation pending for the specified volume. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified volume, and then retry with exponential back off. |
+| Exceeds capabilities | 10 FAILED_PRECONDITION | Indicates that the CO has exceeded the volume's capabilities because the volume does not have MULTI_NODE capability. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |
```
jdef: what's the relationship between supporting these staging calls and SINGLE vs MULTI capability?

spec.md
```diff
@@ -1021,6 +1056,104 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+A Node Plugin MUST implement this RPC call if it has `PUBLISH_UNPUBLISH_DEVICE` node capability.
+This RPC is a reverse operation of `NodePublishDevice`.
+This RPC MUST undo the work by the corresponding `NodePublishDevice`.
+This RPC SHALL be called by the CO once for each `global_target_path` that was successfully setup via `NodePublishDevice`.
+If the corresponding Controller Plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability and the Node Plugin has `PUBLISH_UNPUBLISH_DEVICE` capability, the CO MUST guarantee that this RPC is called and returns success before calling `ControllerUnpublishVolume` for the given node and the given volume.
+The CO MUST guarantee that this RPC is called after all `NodeUnpublishVolume` have been called and returned success for the given volume on the given node.
+
+The Plugin SHALL assume that this RPC will be executed on the node where the volume is being used.
+
+This RPC is typically called by the CO when the workload using the volume is being moved to a different node, or all the workload using the volume on a node has finished.
```
jdef: s/is typically/MAY be/
s/all the workload/all the workloads/
s/has finished/have finished/

csi.proto
```diff
@@ -555,10 +615,18 @@ message NodeUnpublishVolumeRequest {
   // The ID of the volume. This field is REQUIRED.
   string volume_id = 2;
 
+  // The path to which the device was mounted by `NodePublishDevice`.
+  // It MUST be an absolute path in the root filesystem of the process
+  // serving this request.
+  // It MUST be set if the Node Plugin implements the 
+  // `PUBLISH_UNPUBLISH_DEVICE` node capability.
+  // This is an OPTIONAL field.
+  string global_target_path = 3;
```
jdef: why is the staging target path useful here - do we really need it for the use cases we have in mind?

csi.proto
```diff
@@ -496,6 +503,51 @@ message ControllerServiceCapability {
...
+  // volume.
+  // This is a REQUIRED field.
+  string global_target_path = 4;
+
+  // The capability of the volume the CO expects the volume to have.
+  // This is a REQUIRED field.
+  VolumeCapability volume_capability = 5;
+}
+
+message NodePublishDeviceResponse {}
```
jdef: might there be a need for some opaque `staging_volume_info` that would be passed downstream to subsequent `NodePublishVolume` RPCs?

if we had this, would we need the (global/staging)_target_path as a first class field in NodePublishVolume?

spec.md
```diff
@@ -1021,6 +1056,104 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
 
 ### Node Service RPC
```
jdef: Might it be worth adding an "RPC Interations" subsection after all of the node service RPCs (like we did for the controller RPCs: https://github.com/container-storage-interface/spec/blob/master/spec.md#rpc-interactions) in order to concisely describe the relationship between some node RPCs?

I'm suggesting this because it would be a convenient place to compare/contrast the staging volume calls from the publish volume calls: staging is once per node/volume vs publish.

spec.md
davidz627: these calls will only really be useful for MULTI capability plugins (ref counting moved to CO). There is no restriction on using them for SINGLE but the benefit they provide would just be organizational.

csi.proto
davidz627: My understanding that this should be an OK change, it will introduce incompatablity between different CSI spec versions but I think thats OK. Thats why we have a supported_versions field right?

The main issue with changing proto tags is when the protos are actually persisted somewhere and new code with different tags comes along and reads the old protos and gets confused. I don't believe that is an issue here.

@saad-ali may be able to provide more context.

csi.proto
davidz627: the path that the volume is staged to will likely be a mountpoint, the unstage operation will need to be aware of the location of the mountpoint for this specific volume to "unmount" the staging point.

csi.proto
davidz627: https://github.com/container-storage-interface/spec/pull/169#discussion_r162503762

csi.proto
davidz627: from what I understand staging_target_path would be the only field inside of `staging_volume_info`. I may not be aware of other possible use cases and I wouldn't be against making this more open-ended. Do you know of any other fields in particular that would need to be added?

spec.md
davidz627: agreed

spec.md
davidz627: I realize I may have misunderstood your question. This call should be supported for both SINGLE and MULTI capability.

csi.proto
saad-ali: Client/server mismatch of proto field number can cause issues. However, for v0.2, we should be ok making breaking changes, if needed.

spec.md
davidz627: #### RPC Interactions
`NodeStageVolume`, `NodeUnstageVolume`, `NodePublishVolume`, `NodeUnpublishVolume`

The following IS REQUIRED if the plugin advertises the `STAGE_UNSTAGE_VOLUME` capability. 
`NodeStageVolume` MUST be called and return success once per volume per node before any `NodePublishVolume` MAY  be called for the volume.
All `NodeUnpublishVolume` MUST be called and return success for a volume before `NodeUnstageVolume` MAY be called for the volume.

Suggestions?

akutz:
Hi @davidz627,

Regarding naming, I personally like `NodePublishVolume`. I know it's already in use, but I wonder if it doesn't make sense to keep the two existing RPCs, `NodePublishVolume` and `NodeUnpublishVolume` and simply add a field to the former that indicates the SP is responsible for reference counting?

csi.proto
davidz627: 1. Yes, I will add
2. Yes, I will add, also for Unstage

davidz627:
@akutz

I believe the rationale for adding this as a separate call is so that the Storage Plugin will not have to reference count, this allows `NodePublishVolume` and `NodeUnpublishVolume` to have the same functionality each time they are called. Without the new NodeStageVolume/NodeUnstageVolume calls, many storage plugins would have to implement functionality that does an extra "staging mount" and "staging unmount" for the first `NodePublishVolume` and the last `NodeUnpublishVolume` called for each volume for each node.

Adding this new functionality will push the "decision" to make these staging/unstaging mounts to the CO so that it may call these functionalities in specific and separate RPC's. This frees up `NodePublishVolume` and `NodeUnpublishVolume` to only be responsible for the "last mile" of making the volume available to the consumer - it will no longer have to keep track of whether this is the first or fifth `NodePublishVolume` called for the volume.

akutz:
Hi @davidz627,

You know what, I’ve been feeling unwell today and clearly I had a massive brain fart when I suggested that. It’s obviously anathema to the intended behavior :)

Mea culpa.

akutz:
Regarding naming, I keep coming back to NodePublishDevice, but that ties it to device-backed storage, and honestly, it’s a volume-centric API operation anyway.

How about augmenting NodePublishVolume with a “source_path” field that, when specified, indicates that the operation should bind mount “source_path” to “target_path”. Otherwise the volume should be directly mounted to “target_path” per the norm. The NodeUnpublishVolume call would not have to change since it would be called once for each corresponding publication call.

davidz627:
Thanks for the suggestion Andrew, we've added in staging_target_path in the publish call for the bind mount indication. It is an optional field, so if unspecified, the volume will be directly mounted to target_path. I have also not added it to NodeUnpublishVolume because of the reasoning you mentioned.

jieyu:
xref #166

akutz:
Hi @davidz627,

I guess I'm not sure why we need to create a new RPC when the existing ones would work as-is, and the SP could make the decision based on whether or not the `staging_target_path` field is empty or not?

davidz627:
NodeStageVolume/NodeUnstageVolume are required to setup/tear-down the staging_target_path mount.

akutz:
Hi @davidz627,

Yes, but the point I'm trying to make is that they do not have to be. If the staging path does not exist the first time a call is made to `NodePublishVolume` then that RPC could do all of the same work as `NodeStageVolume`.

akutz:
Hi All,

FWIW, @davidz627 and I had a conversation offline where I came around to the PR's solution. I was approaching the PR from the point-of-view that an SP must support COs that may or may not implement ref counting. David pointed out that **all** COs must implement ref counting -- **IFF** an SP advertises the `STAGE_UNSTAGE_VOLUME` capability. I was arguing that if an SP must be able to do ref counting anyway if a CO doesn't then why not wrap it all in the existing `NodePublishVolume`. Once I realized it was up to the SP to determine if the CO handles ref counting (since all COs **MUST** be able to do if the SP asks them to), then I understood that distinct RPCs make sense.

Sorry for the confusion. I've had a touch of dumbassitis for the last several years and have been slow on the pickup it seems. :)

jieyu:
According to the current spec, only volumes with MULTI_NODE_XXX capabilities potentially need this extra RPC (because for SINGLE_NODE_XXX volumes, `NodePublishVolume` won't be called for different `target_path`, thus the logic can be in `NodePublishVolume`).

So for plugins that're capable of dealing with MULTI_NODE volumes, how many of them will benefit with this extra RPC? Can you provide some concrete examples? I imagine that most of the fuse based plugins (MULTI_NODE capable) won't benefit from this extra RPC.

I think my main worry about this change is that it introduces two ways for a plugin to implement the same functionality. A plugin can either choose to implement this new RPC, or do ref counting itself. I understand the benefit of this: allowing plugin to be stateless and no need to do ref counting. But isn't that hard? In most of the case I imagine it's just parsing the mount table.

davidz627:
@jieyu I know of a few Storage Plugin types that will/may benefit:
AWS EBS, Azure DD, Fiber Channel, GCE PD, ISCSI, Photon, Rados Block Device, VSphere

Yes, the plugin will have a "choice" but if this type of functionality is required it makes sense for them to offload the work to the CO so that the SP can just respond "dumbly" to requests instead of doing its own ref counting and failure recovery.

jieyu:
@davidz627 As far as I know, `AWS EBS` should be categorized as `SINGLE_NODE_WRITER` because it cannot be published to multiple nodes. As a result, `NodePublishVolume` shouldn't be called multiple times according to the current spec. I expect many of the plugins that you mentioned above fall into the same category.

davidz627:
Many of the listed volumes support multi reader only.

davidz627:
@jieyu many of the volumes listed support `MULTI_NODE_READER_ONLY`. This PR will improve SP development for any volume that supports `MULTI_NODE_READER_ONLY` which is a commonly supported storage mode.

jieyu:
@davidz627 I am wondering for `MULTI_NODE_READER_ONLY` volumes, can the CO always call `NodeStageVolume` for each `NodePublishVolume`. In other words, can the logic in `NodeStageVolume` can be merged into `NodePublishVolume`? I'll be a bit surprised if a volume can be published at multiple nodes, but cannot perform the "NodeStage" step multiple times on a node.

davidz627:
@jieyu I've done some more research into this.
I have confirmed the following drivers *require* a local attach step that NodeStageVolume/NodeUnstageVolume that must be done per-volume/per-node that cannot be folded into PublishVolume/UnpublishVolume (specifically the unpublish volume part would require ref counting):

FC, ISCSI, NVMEOE, Lustre, RBD. Potentially Ceph and Rook (@sbezverk may be able to confirm)

Additionally, Any SINGLE_NODE_ plugin will benefit from if we clarify access modes to allow multiple workloads on a single node access the volume. Currently, that access mode paradigm is the one adopted by Kubernetes and I think it is the correct way to go. Multiple workloads accessing a SINGLE_NODE_ volume on the same node can have several benefits (live backup, migration comes to mind) and not too many drawbacks.

akutz:
Hi @davidz627,

Oh boy, now we're rehashing #119 and #150. Look, myself and @codenrhoden are both in favor of supporting non `MULTI` volumes as published to multiple target paths on a single host. But we've had that discussion. @jieyu and others said "no." Travis and I see no reason it cannot work. It works today on K8s. However, that was the decision that was made. 

If we're reopening this discussion, I'll simply say that I again think the existence of a separate `NodeStageVolume` RPC is purely a design choice. The advertised capability doesn't require a separate RPC. I know I said I agreed with the separate RPCs, and I have no new issues with them, I'm just pointing out they're not required and only an authorial choice.

Today there are several CSI plug-ins that implement private mount directories and stage `SINGLE` volumes:

* [`csi-vsphere`](https://github.com/thecodeteam/csi-vsphere)
* [`csi-nfs`](https://github.com/thecodeteam/csi-nfs)
* [`csi-blockdevices`](https://github.com/thecodeteam/csi-blockdevices)
* [REX-Ray](https://github.com/thecodeteam/rexray)

They work great! Don't need separate RPCs either. If we just added the capability to indicate to a CO that it needs to handle ref counting then the above SPs could simply stop handling it inside `NodePublishVolume`.

Anyway, here we go again :)

davidz627:
Right, lets ignore the SINGLE_NODE_ volumes for now. The new RPC will still benefit in design:
FC, ISCSI, NVMEOE, Lustre, RBD. Potentially Ceph and Rook.

The argument is that this new RPC is *useful* and doesn't just add an *unnecessary* RPCs. The main benefit (shifting ref-counting responsibility to the CO) is valid for all the above storage plugins. So my argument is that it is useful for a significant number of SPs to have this new RPC. It does not add any additional complexity to existing SP workflows if they choose not to implement this *optional* RPC.

akutz:
Hi @davidz627,

To play devil's advocate:

> The main benefit (shifting ref-counting responsibility to the CO) is valid for all the above storage plugins.

Technically the benefit is derived from the ability of the SP to advertise the capability`STAGE_UNSTAGE_VOLUME` which indicates to COs they should handle reference counting. The RPC is simply the method by which they do that. The same method could be folded into `NodePublishVolume`.

davidz627:
The same functionality could be technically folded into `NodeUnpublishVolume` but then you would need to have a indicator on the call to notify the SP that it is the "final" unpublsh call and it is time to clean up the per-volume/per-volume artifacts. Therefore the same RPC will have significantly different behaviors/codepaths depending on this indicator.

To play devil's devil's advocate:
Why have seperate RPCs at all? Why not just one big RPC `Publish` with several indicator switches in the RPC that specify which step of the publish you are calling for. Then the storage plugin can just be one big switch case.

The design with the new Stage RPC does add more complexity, I agree with you there. But my argument is that the per-volume/per-node step is a separate and distinct step that is useful for many SP's so it *should* be pulled out into a new RPC.

jieyu:
@davidz627 

> Additionally, Any SINGLE_NODE_ plugin will benefit from if we clarify access modes to allow multiple workloads on a single node access the volume. Currently, that access mode paradigm is the one adopted by Kubernetes and I think it is the correct way to go. Multiple workloads accessing a SINGLE_NODE_ volume on the same node can have several benefits (live backup, migration comes to mind) and not too many drawbacks.

The draw back is that it creates an implicit scheduling constraint that two workloads must be scheduled on the same box, and if not, it won't work because the access mode is SINGLE_NODE_XXX.

sbezverk:
@jieyu I opened issue #178 to revisit SINGLE_NODE support matter.  We have a strong ask for this support from some communities like ceph and rook which are looking forward to use CSI and this feature is important to them as it was already implemented in their plugins. We are asking them to share with CSI community their uses cases in order to re-evaluate previous decision.

As per this PR, CEPH plugin can definitely benefit from this proposal.

chakri-nelluri:
+1. Today applications use affinity to co-schedule backup jobs and mount the same volume read-only for backups. With out separate stage & publish, what will be the recommended mechanism to support it?

jieyu:
@chakri-nelluri 

>With out separate stage & publish, what will be the recommended mechanism to support it?

Mesos supports that by allowing a container to be injected into a running Pod dynamically. Introducing implicit scheduling constraint like this makes the scheduler's job much more harder.

As I mentioned in #150, instead of loosening this constraint, we should revise the `AccessMode` to explicitly distinguish single node single publish vs. single node multi publish.

davidz627:
@jieyu ISCSI has a significant local-attach step that cannot be repeated uniquely for every `NodePublishVolume` step. It needs to be done once before all the `NodePublishVolume` and cleaned up after *all* `NodeUnpublishVolumes`. 

As we can see from the documentation:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/online_storage_reconfiguration_guide/iscsilogin

You cannot specify a /dev/{device} when attaching, it will pick a device on its own (it is idempotent as well)

After the disk is locally attached you should be able to find the devices here:
/dev/disk/by-path/ip-{tp}-iscsi-{lqn}-lun-{lun}
or /dev/disk/by-path/pci-*-ip-{tp}-iscsi
depending on what interface they are attached with

The locations are deterministic based on target and LUN.

jieyu:
@davidz627 Thanks! That sounds quite convincing to me that we should add this step to simplify some SPs. I am +1 to this PR.

jdef:
@jieyu can you mark this as "approved" if you're on board with these changes?

spec.md
```diff
@@ -1021,6 +1056,122 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+  // This field is OPTIONAL.
+  map<string, string> user_credentials = 6;
+
+  // Attributes of the volume to publish. This field is OPTIONAL and
+  // MUST match the attributes of the VolumeInfo identified by
+  // `volume_id`.
+  map<string,string> volume_attributes = 7;
+}
+
+message NodeStageVolumeResponse {}
```
saad-ali: Would it be useful to have a `volume_stage_attributes` to pass between `NodeStageVolume` and `NodePublishVolume`?

spec.md
```diff
@@ -1153,6 +1312,14 @@ message NodeUnpublishVolumeRequest {
 
 message NodeUnpublishVolumeResponse {}
 ```
+#### RPC Interactions and Reference Counting
```
saad-ali: Might make sense to move this up to the `NodeStageVolume` section?

spec.md
```diff
@@ -1021,6 +1056,122 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+  // This field contains credential data, for example username and
+  // password. Each key must consist of alphanumeric characters, '-',
+  // '_' or '.'. Each value MUST contain a valid string. An SP MAY
+  // choose to accept binary (non-string) data by using a binary-to-text
+  // encoding scheme, like base64. An SP SHALL advertise the
+  // requirements for credentials in documentation. COs SHALL permit
+  // users to pass through the required credentials. This information is
+  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> user_credentials = 6;
```
saad-ali: Please change the name to `node_stage_credentials` to align with https://github.com/container-storage-interface/spec/pull/168

spec.md
```diff
@@ -1021,6 +1056,122 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+  // The API version assumed by the CO. This is a REQUIRED field.
+  Version version = 1;
+
+  // The ID of the volume. This field is REQUIRED.
+  string volume_id = 2;
+
+  // The path at which the volume was published. It MUST be an absolute
+  // path in the root filesystem of the process serving this request.
+  // This is a REQUIRED field.
+  string staging_target_path = 3;
```
saad-ali: No credentials for `NodeUnstage`?

spec.md
davidz627: ack

davidz627:
@saad-ali @jieyu PTAL. I have addressed all of Saad's and @jdef's comments.

spec.md
```diff
@@ -1021,6 +1056,150 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+#### RPC Interactions and Reference Counting
+`NodeStageVolume`, `NodeUnstageVolume`, `NodePublishVolume`, `NodeUnpublishVolume`
+
+The following interaction semantics ARE REQUIRED if the plugin advertises the `STAGE_UNSTAGE_VOLUME` capability. 
+`NodeStageVolume` MUST be called and return success once per volume per node before any `NodePublishVolume` MAY be called for the volume.
+All `NodeUnpublishVolume` MUST be called and return success for a volume before `NodeUnstageVolume` MAY be called for the volume.
+
+Note that this requires that all COs MUST support reference counting of volumes so that if `STAGE_UNSTAGE_VOLUME` is advertised by the SP, the CO MUST fufill the above interaction semantics.
+
 #### `NodePublishVolume`
```
jdef: NodePublishVolume also accepts a publish_volume_info, which may now be generated from either ControllerPublishVolume or NodeStageVolume. It's probably worth mentioning that fact for the field in the NodePublishVolumeRequest.

spec.md
```diff
@@ -1021,6 +1056,150 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+  // `volume_id`.
+  map<string,string> volume_attributes = 7;
+}
+
+message NodeStageVolumeResponse {
+  // The SP specific information that will be passed to the Plugin in
+  // the subsequent `NodePublishVolume` calls
+  // for the given volume.
+  // This information is opaque to the CO. This field is OPTIONAL.
+  map<string, string> publish_volume_info = 1;
```
jdef: Maybe some clarifying language? I think there might be some confusion now that publish_volume_info is emitted from potentially two sources (ControllerPublishVolumeResponse, and NodeStageVolumeResponse). Is the CO expected to combine the results from both calls when making a subsequent NodePublishVolume call? If this field is not present in this message, but it was in a ControllerPublishVolumeResponse, does that mean that we DO or DO NOT specify a publish_volume_info when calling NodeStageVolumeResponse?

jieyu: Can you explain why this is needed? I'd like to see a concrete use case why this extra information returned from `NodeStageVolume` is necessary.

spec.md
```diff
@@ -1021,6 +1056,150 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
 
+#### `NodeStageVolume`
+
+A Node Plugin MUST implement this RPC call if it has `STAGE_UNSTAGE_VOLUME` node capability.
+This RPC is called by the CO when a workload that wants to use the specified volume is placed (scheduled) on a node.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+This RPC SHOULD be called by the CO when a workload that wants to use the specified volume is placed (scheduled) on the specified node for the first time or for the first time since a `NodeUnstageVolume` call for the specified volume was called and returned success on that node.
+If the corresponding Controller Plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability and the Node Plugin has `STAGE_UNSTAGE_VOLUME` capability, then the CO MUST guarantee that this RPC is called after `ControllerPublishVolume` is called for the given volume on the given node and returns a success.
+The CO MUST guarantee that this RPC is called and returns a success before any `NodePublishVolume` is called for the given volume on the given node.
+This operation MUST be idempotent.
```
jieyu: Can you be more specific about the idempotency requirement like other calls.

```
If the volume corresponding to the `volume_id` has already been staged at .... the Plugin MUST reply `0 OK`.
```

spec.md
```diff
@@ -1021,6 +1056,150 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+This RPC MUST undo the work by the corresponding `NodeStageVolume`.
+This RPC SHALL be called by the CO once for each `staging_target_path` that was successfully setup via `NodeStageVolume`.
+If the corresponding Controller Plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability and the Node Plugin has `STAGE_UNSTAGE_VOLUME` capability, the CO MUST guarantee that this RPC is called and returns success before calling `ControllerUnpublishVolume` for the given node and the given volume.
+The CO MUST guarantee that this RPC is called after all `NodeUnpublishVolume` have been called and returned success for the given volume on the given node.
+
+The Plugin SHALL assume that this RPC will be executed on the node where the volume is being used.
+
+This RPC MAY be called by the CO when the workload using the volume is being moved to a different node, or all the workloads using the volume on a node have finished.
+
+This operation MUST be idempotent.
```
jieyu: Ditto. Can you be more specific about the idempotency requirement?

spec.md
davidz627: Thought about it some more and discussed with Saad, between the two of us we couldn't actually come up with a concrete use case. I think we'll omit this field, if it turns out there is a use case for it in the future it can be added later.

davidz627:
@jieyu @jdef addressed your comments. Thanks for reviewing so thoroughly!

spec.md
```diff
@@ -1021,6 +1056,153 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+
+This RPC is called by the CO when a workload that wants to use the specified volume is placed (scheduled) on a node.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+This RPC SHOULD be called by the CO when a workload that wants to use the specified volume is placed (scheduled) on the specified node for the first time or for the first time since a `NodeUnstageVolume` call for the specified volume was called and returned success on that node.
+
+If the corresponding Controller Plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability and the Node Plugin has `STAGE_UNSTAGE_VOLUME` capability, then the CO MUST guarantee that this RPC is called after `ControllerPublishVolume` is called for the given volume on the given node and returns a success.
+The CO MUST guarantee that this RPC is called and returns a success before any `NodePublishVolume` is called for the given volume on the given node.
+
+This operation MUST be idempotent.
+If the volume corresponding to the `volume_id` is already staged to the `staging_target_path`, and is compatible with the specified `volume_capability` the Plugin MUST reply `0 OK`.
```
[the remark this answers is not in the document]
davidz627: Could there be volume capabilities that are subsets of others? I know there aren't really right now, but maybe there would be in the future? Will change to "identical" for now but should keep in mind to change it if there are subset volume capabilities in the future.

davidz627:
@jdef addressed comments.

spec.md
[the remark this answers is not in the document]
jdef: Thanks. I think I speak for several approvers when I say that we'd rather the spec be more strict for now. We've been trying err on side of strictness with the intent that we can "open things up" if needed later on.

jdef:
Looks like we need another CO to sign off: @saad-ali @julian-hj @cpuguy83?

akutz:
Hmm. If the consensus in #171 is that an SP should be aware of how it is running, then I reiterate my objection to the need for separate stage RPCs. An SP should know it advertises the related capability and do the appropriate logic in the existing RPCs. This is inline with @jdef's remarks about keeping things tight.

akutz:
Also, I suggest that @saad-ali recuse himself from approving this SP as he originated the issue. I nominate either @julian-hj, @jieyu, @cpuguy83 as the second approver.

jdef:
Jie and I represent the same CO, so we need someone from another.

On Thu, Feb 8, 2018 at 1:47 PM, Schley Andrew Kutz <notifications@github.com

spec.md
```diff
@@ -1029,6 +1064,153 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
 
 ### Node Service RPC
 
+#### `NodeStageVolume`
+
+A Node Plugin MUST implement this RPC call if it has `STAGE_UNSTAGE_VOLUME` node capability.
+
+This RPC is called by the CO when a workload that wants to use the specified volume is placed (scheduled) on a node.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+This RPC SHOULD be called by the CO when a workload that wants to use the specified volume is placed (scheduled) on the specified node for the first time or for the first time since a `NodeUnstageVolume` call for the specified volume was called and returned success on that node.
```
cpuguy83: Shouldn't this be before the workload is placed?

akutz:
From earlier:

> Hmm. If the consensus in #171 is that an SP should be aware of how it is running, then I reiterate my objection to the need for separate stage RPCs. An SP should know it advertises the related capability and do the appropriate logic in the existing RPCs. This is inline with @jdef's remarks about keeping things tight.

I want to make sure this is seen since no one commented on it. The more I consider this, the more I am confident that RPCs separate from the existing `NodePublishVolume` and `NodeUnpublishVolume` is not the solution. Some reasons:

* Multiple CO members have stated that an SP should be aware of how it is running/configured. Thus a CO knows if it is running with the `STAGE_UNSTAGE_VOLUME` capability and thus can handle the appropriate logic in the `NodePublishVolume` RPC.
* The spec states that a CO will do its best effort to ensure that RPCs are called in a serial fashion, but that if a CO or SP process crashes then it is the responsibility of the SP to handle recovery behaviors. To that end, it's very possible that a CO sends a `NodeStageVolume` request to an SP that fails to complete the request and the CO also fails to receive the error or record the failed attempt. In this case the CO may send a `NodePublishVolume` request expecting the aforementioned `NodeStageVolume` was completed successfully. Based on previous conversations I expect that @saad-ali might say the SP should do its best effort to handle this behavior. I double-checked the PR's changes -- there is no error code defined for `NodePublishVolume` to return to a CO that indicates the request's `staging_target_path` field is set but it appears `NodeStageVolume` was not invoked. Without such an error, it would be easy to infer that the intent is `NodePublishVolume` should try to do both the logic for `NodeStageVolume` **and** `NodePublishVolume`. If `NodePublishVolume` has to handle both in edge cases anyway, why not just consolidate it?

I agree that the capability `STAGE_UNSTAGE_VOLUME` is a good idea. I just don't see the need for the additional RPCs. The spec should remain as opaque and simple as possible, and CO members seem to agree to that much of the time. This is one of those cases, an opportunity to either expand the spec or keep it the way it is with minimal change to achieve the desired functionality. I do not see a compelling reason why we are not electing to use the simple, straight-forward, and simple solution now and expanding it later and only if necessary.

akutz:
James,

You missed something in my remarks. I said as is with the new STAGE capability. The SP would still be relieved of ref counting, it would just handle the stage logic at the top of the existing RPCs.

davidz627:
I've reworded the RPC per @jdef and @cpuguy83's suggestions to be more clear as to what this RPC is doing.

I've also added a new error on NodePublishVolume specifically for when NodeStageVolume is not set as I believe this could use a specific call-out.

spec.md
```diff
@@ -1029,6 +1064,153 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+| Operation pending for volume | 10 ABORTED | Indicates that there is a already an operation pending for the specified volume. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified volume, and then retry with exponential back off. |
+| Exceeds capabilities | 10 FAILED_PRECONDITION | Indicates that the CO has exceeded the volume's capabilities because the volume does not have MULTI_NODE capability. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |
```
cpuguy83: In what cases would the node plugin be aware of this state?

spec.md
```diff
@@ -1029,6 +1064,153 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+  // left unset if the corresponding Controller Plugin does not have
+  // this capability. This is an OPTIONAL field.
+  map<string, string> publish_volume_info = 3;
+
+  // The path to which the volume will be published. It MUST be an
+  // absolute path in the root filesystem of the process serving this
+  // request. The CO SHALL ensure uniqueness of staging_target_path per
+  // volume.
+  // This is a REQUIRED field.
+  string staging_target_path = 4;
```
cpuguy83: I get that this falls in line with the publish RPC, but I'm wondering if this is something that really works here.
Since the CO would not consume this path, I'm not sure it's the CO's responsibility to know anything about it.

spec.md
davidz627: The Node Plugin should always be aware of its "NodeServiceCapability". The SP is the one that sets the `STAGE_UNSTAGE_VOLUME` capability in the first place.

It will receive the request parameters in the request and should be able to check whether staging_target_path is empty/nil.

spec.md
```diff
@@ -1029,6 +1064,153 @@ It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capab
...
+
+This RPC is called by the CO prior to the volume being consumed by any workloads on the node by `NodePublishVolume`.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+This RPC SHOULD be called by the CO when a workload that wants to use the specified volume is placed (scheduled) on the specified node for the first time or for the first time since a `NodeUnstageVolume` call for the specified volume was called and returned success on that node.
+
+If the corresponding Controller Plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability and the Node Plugin has `STAGE_UNSTAGE_VOLUME` capability, then the CO MUST guarantee that this RPC is called after `ControllerPublishVolume` is called for the given volume on the given node and returns a success.
+The CO MUST guarantee that this RPC is called and returns a success before any `NodePublishVolume` is called for the given volume on the given node.
+
+This operation MUST be idempotent.
+If the volume corresponding to the `volume_id` is already staged to the `staging_target_path`, and is identical to the specified `volume_capability` the Plugin MUST reply `0 OK`.
```
cpuguy83: Can we clarify what to do when volume ID exists with a different staging path?

spec.md
cpuguy83: Discussed this on slack, the description of this is really the path to publish for the node, and `NodePublish`'s `target_path` is really the path for the single workload.
This effectively treats the CO as a storage mechanism which I am ok with for this case.

Perhaps the `staging_` should be omitted in the stage RPC?

spec.md
akutz: Hi @cpuguy83,

Good catch. I firmly believe that CSI 0.2.0 needs to introduce extended error details because of this an other conditions where one of the several gRPC status codes just isn't enough information. For example, `CapacityError` for creating a volume with an invalid capacity range is an abuse of an eponymous code. It should be an `InvalidArgument` code with extended error details.

spec.md
davidz627: I think having the `staging_` makes it especially clear that this is the same path as the one that is used as `staging_target_path` in `NodePublishVolume` in later calls. Removing it would be naming the same variable differently in different contexts and may introduce some confusion.

spec.md
davidz627: As per the `NodeStageVolumeRequest` description: `The CO SHALL ensure uniqueness of staging_target_path per volume.`, so this is the CO's job to ensure. If the SP were to receive non-unique `staging_target_path` this would be undefined behavior, and I am ok leaving it as that.

Would be willing to define the behavior if you think that this is not enough for the spec and have a suggestion for what should happen.

cpuguy83: In this case, uniqueness of the path is not an issue, it's that the SP got a 2nd target path for for a given volume ID.

saad-ali:
/lgtm

Will let @cpuguy83 provide the second approval.

akutz:
Hi All,

I'm just making one last note that these new RPCs don't make things any easier on SP developers. The `STAGE_UNSTAGE_VOLUME` capability *might* help, but even that could be up for debate.

The reason is that SP developers should **still** try to be as exhaustive as they are able when it comes to publication/unpublication on nodes. For example:

* A `NodeUnstageVolume` request cannot just check to see if a directory exists or not. The request has to examine the mount table and verify that there isn't a trinary relationship in effect, perhaps even with a directory that doesn't exist but for which there still exists a bind mount entry in the mount table (an annoying, but possible scenario).
* A `NodePublishVolume` request should attempt to create missing path components and scan the bind mount relationships to determine if the operation is idempotent. 

Because developers will **still** need to implement exhaustive logic in both the Node Stage/Pub Unstage/Unpub RPCs, separating them into two sets of RPCs doesn't make anything easier. If anything it makes things more complicated for developers that have to duplicate checks across multiple RPCs.

But hey, I guess y'all have already decided that this is what's happening. I just don't understand the desire to add to the spec when opaqueness should be the guiding principal and the existing RPCs in addition to perhaps a new capability and the `stage_target_path` can already do the job.

akutz:
One last thing, I know most of the project's voting members belong to COs. However, I'm a voice of someone working on both the CO side **and** developing SPs. I would ask that my arguments be given the consideration that dual-focus might justify. Having worked on more than a half dozen CSI SPs so far, I'm saying that this change doesn't make things easier because ultimately you don't want to fail fast -- your goal for an SP is for it to succeed, even when that means doing some exhaustive discovery logic. This change just means that logic is duplicated / chopped up into two locations, making things more complex on SP devs.

akutz:
Hi all,

I'd just ask that y'all look at CSI-VFS's [`NodePublishVolume`](https://github.com/akutz/csi-vfs/blob/ee3752774c886080081655657ec3a83b495d8dde/service/node.go#L18-L124) and [`NodeUnpublishVolume`](https://github.com/akutz/csi-vfs/blob/ee3752774c886080081655657ec3a83b495d8dde/service/node.go#L126-L202) RPC implementations. Note that things don't fail fast -- we try to reconnect broken links as it were to avoid a system getting into a bad state.

cpuguy83:
@akutz I don't think the CO's are arguing for this at all. I know this was brought up by the Kube team, but only at the request of some SP's.... so maybe we need to re-sync on this topic.

I definitely agree that a good plugin would not rely simply on the CO to ensure these are correct. State will definitely get out of sync.

akutz:
Hi Brian,

Yeah, this issue was filed by Saad. If SP devs raised it with him, I was unaware.

akutz:
Hi Brian,

Anyway, my point, as you so succinctly restated it (thank you), is that SPs must, or rather **should**, be capable of detecting drift and reconciling it to sync state. These additional RPCs simply introduce new locations that must implement nearly identical reconciliation logic, and thus add little value for me, an SP dev.

Unless we also add a new RPC that allows a CO to validate its expected state against a Node or Controller, I prefer keeping he number of mutative RPCs to a bare minimum.

spec.md
davidz627: I've clarified the wording to `The CO SHALL ensure that there is only one staging_target_path per volume.` hope that helps

davidz627:
@akutz if `NodeStageVolume` does not suit the needs of your SP or it is not architected in a way that `NodeStageVolume` would be beneficial to it, then the SP does not need to implement it.

It is an *optional* call that has its use and make development of many SP's easier as outlined above.

If it is easier for your SP to implement everything in `NodePublishVolume` even if that includes reference counting, the SP is welcome to do that. It just should not advertise `STAGE_UNSTAGE_VOLUME` capability.

The arguments you made are all assuming that `NodeStageVolume` will be forced upon your SP's to implement even though it may not be the best way for your SP's to be architected. That is simply not the case.

akutz:
I’m arguing I’ve implemented multiple types of SPs and that my advice that this doesn’t actually make things easier should be given consideration. I know from experience that there is a common enough pattern between SPs that the same reasons I outlined above apply to the majority of SPs and thus this change has minimal actual value and perhaps even the opposite.

saad-ali:
> A NodeUnstageVolume request cannot just check to see if a directory exists or not. The request has to examine the mount table and verify that there isn't a trinary relationship in effect, perhaps even with a directory that doesn't exist but for which there still exists a bind mount entry in the mount table (an annoying, but possible scenario).

It doesn't have to. Per spec, all an SP has to do is make the specified device available at the specified path. Lots of volume plugins do just this in Kubernetes without extensive additional mount table verification (which is very expensive) and are deployed at scale in production workloads without issues.

> A NodePublishVolume request should attempt to create missing path components and scan the bind mount relationships to determine if the operation is idempotent.

Again, this is not true. The CO SHALL ensure that the path exists. And a plugin can simply issue a bind mount. If the bind mount already exists it will simply succeed and be idempotent.

> I'm saying that this change doesn't make things easier because ultimately you don't want to fail fast -- your goal for an SP is for it to succeed, even when that means doing some exhaustive discovery logic. 

This addition does not (and is not intended) to simplify all plugins. It is an optional addition designed to make a certain class of plugins, those that have a local attach/detach step, easier to implement. It should go without saying, you should not implement it if it doesn't make sense for your plugin or complicates the logic (as it does in your case).

Some plugins may choose to implement more complicated logic, but if we can lower the bar for writing a class of plugins, that will be beneficial to the ecosystem as a whole.

I believe @davidz627 has diligently addressed the feedback so far. And this PR is close to done. @cpuguy83 do you have any other concrete concerns?

@davidz627 could you please rebase the PR.

cpuguy83:
@saad-ali I think the argument is that this makes the happy case simpler but does not help at all in the case that there are problems.
I know that SP's are requesting this to offload reference counting onto the CO, but this can't really do that... not without opening up a lot of room for errors.

Do we have some concrete examples of a SP that cannot work without this?

akutz:
I no longer really have any dog in this fight, and while I stand by what I've said, I agree with @davidz627 that this is optional and doesn't harm SPs that don't wish to implement it. I just worry that the bigger picture with regards to state won't be as apparent to new SP devs and these new RPCs could be tricky to implement. Either way, I'm going to approve the PR and wish y'all the best of luck!

saad-ali:
@cpuguy83 the purpose of this call was always to lower the bar for writing a plugin, not to fix an blocking issue. Volume plugins can implement this logic in the existing `NodePublishVolume` call, but, for a class of plugins, doing so requires the plugin have more complicated logic.

I disagree that plugins that will use this will open themselves up for errors. If this is a concern, let's be more concrete about the error cases that would be more common. It is important to not that this is a model that many volume plugins already follow in Kubernetes without issue.

"A workload referencing a specified volume is scheduled to a node for the first time" and "the last workload referencing a specified volume is removed from a node" are important events and plugins should be able to hook in to they if they choose.

davidz627:
Rebased. Thanks everyone for the thorough reviews and questions!

PR #179, opened 2018-01-31
Updates protoc binary to 3.5.1

akutz:
This patch updates the protoc binary to 3.5.1.


jieyu:
Need one more approval. cc @jdef @saad-ali @julian-hj @cpuguy83

PR #180, opened 2018-01-31
Generating Go Bindings with Docker

akutz:
This patch introduces the ability to generate Go language bindings using
Docker with `Dockerfile-go`:

```shell
# builds the container image
$ docker build -t csi/spec -f Dockerfile-go .

# emits the go bindings to stdout
$ docker run --rm csi/spec

# emits the protobuf to stdout
$ docker run --rm csi/spec csi.proto
```

### DIFF
--- a/Dockerfile-go
+++ b/Dockerfile-go
@@ -1,0 +1,33 @@
+FROM golang:1.9.3
+
+RUN apt-get update && apt-get install -y unzip
+
+RUN mkdir -p /csi
+WORKDIR /csi
+
+ENV PROTOC_VER 3.5.1
+ENV PROTOC_ZIP protoc-$PROTOC_VER-linux-x86_64.zip
+ENV PROTOC_URL_PFX https://github.com/google/protobuf/releases/download
+ENV PROTOC_URL $PROTOC_URL_PFX/v$PROTOC_VER/$PROTOC_ZIP
+ENV PROTOC_BIN bin/protoc
+RUN curl -LO $PROTOC_URL && unzip $PROTOC_ZIP && chmod 0755 $PROTOC_BIN
+
+RUN go get -u github.com/golang/protobuf/protoc-gen-go
+
+COPY spec.md /csi
+
+RUN cat spec.md | \
+  sed -n -e '/```protobuf$/,/```$/ p' | \
+  sed -e 's@^```.*$@////////@g' > csi.proto
+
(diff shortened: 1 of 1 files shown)


saad-ali:
I like the idea of containerizing the build process. How do you envision this being used? Would Makefile be modified to build and run the container?

akutz:
Hi @saad-ali,

This is just a short-cut for now. Ultimately, even if the Travis-CI build went to sudo mode to utilize this process, we'd still want to employ `dep` to manage the `protobuf` dependency to make sure it and its dependency graph are pinned at known versions.

So, in other words, the first step is merging #113, then building on that with this to have the option to do a Docker build with known dependencies and required tools, such as `protoc`.

akutz:
Hi @saad-ali,

Additionally, I'd likely publish `csi/spec` as an executable image rather than what it is now, a way to build the current `spec.md` into the generated output. As an executable image it would simply consume `spec.md` and emit the desired output, ex. `csi.proto`, `csi.pb.go`, etc. I started down this path in fact except that the default `golang` Docker image doesn't have the `unzip` command in it, and I could not use it to get and inflate the `protoc` zip file needed to do the one-liner.

akutz:
However, even as an executable image, we'd still need `dep` to manage the Go dependency graph as no one wants to use `go get` to get random dependencies based on `HEAD` revisions. So if we need `dep` and its `Gopkg` files regardless, why not use the ones in #113 as a starting point?

Dockerfile-go
```diff
@@ -0,0 +1,33 @@
...
+RUN cat spec.md | \
+  sed -n -e '/```protobuf$/,/```$/ p' | \
+  sed -e 's@^```.*$@////////@g' > csi.proto
+
+RUN $PROTOC_BIN -I . --go_out=plugins=grpc:. csi.proto
+
+FROM alpine:3.7
+RUN mkdir /csi
+WORKDIR /csi
+COPY --from=0 /csi/csi.pb.go .
```
cpuguy83: Just a nit, you can name each stage with `FROM golang:1.9.3 AS build` and then use `COPY --from=build`

cpuguy83:
> Would Makefile be modified to build and run the container?

Ideally the Makefile would not require a container.

Dockerfile-go
akutz: Hi @cpuguy83,

Thanks! I've actually done that, but since this was a single stage being referenced I didn't care.

akutz:
Replaced by #181

PR #181, opened 2018-02-01
Generating Go Bindings with Docker

akutz:
This patch replaces #180 and introduces the ability to generate Go language bindings using Docker with `Dockerfile-go` and a spec file provided at runtime:

```shell
# builds the container image
$ docker build -t csi/golang -f Dockerfile-go .

# emits the go bindings to stdout
$ cat spec.md | docker run --rm -i csi/golang
```

## Error Handling
If the image is run without content available on STDIN then the following error is emitted:

```shell
$ docker run csi/golang
usage: cat spec.md | docker run --rm -i golang/csi

$ docker run -i csi/golang
usage: cat spec.md | docker run --rm -i golang/csi
```

## Emit the Protobuf
The image can also be used to emit the generated protobufs as well, just add the token `proto` at the end of the command used to generate the Go bindings:

```shell
# emits the protobuf to stdout
$ cat spec.md | docker run --rm -i csi/golang proto
```

## Image Size
Due to the usage of build stages, the size of the final image is only 25MB!

```shell
$ docker images
REPOSITORY                         TAG                 IMAGE ID            CREATED              SIZE
csi/golang                         latest              bfe16b5e4bb4        About a minute ago   25MB
```

## Docker Hub (`akutz/csi-golang`)
As I lack the ability to push to the Docker hub for the `csi` organization, I've pushed the image to the Docker hub as `akutz/csi-golang`. Please give it a whirl!

```shell
$ cat spec.md | docker run --rm -i akutz/csi-golang
Unable to find image 'akutz/csi-golang:latest' locally
latest: Pulling from akutz/csi-golang
ff3a5c916c92: Already exists 
f9fb51a84b3e: Pull complete 
9e169ec17526: Pull complete 
0aaec76a3e6b: Pull complete 
6182241d2b25: Pull complete 
b7d1cfe2d7c2: Pull complete 
Digest: sha256:7291fee5b3169be78ce9de02cc45334100f2958f6fc279d1e61e7172ee846a46
Status: Downloaded newer image for akutz/csi-golang:latest
// Code generated by protoc-gen-go. DO NOT EDIT.
// source: csi.proto

/*
Package csi is a generated protocol buffer package.

It is generated from these files:
	csi.proto

...
```

cc @saad-ali @cpuguy83


lpabon:
Does it take the spec.md (markdown) and spit out the proto file?

akutz:
No, it spits out the Go bindings. I can enhance it to spit out various pieces depending on flags, but by default, since it’s named “csi/golang”, it emits the Go bindings.

FWIW, the PR States it emits the Go bindings.

lpabon:
Sorry, yeah, i meant go bindings.

akutz:
Ha! I'm glad you mistyped it, because I just rebased this PR and now it can emit the protobuf as well, just add the command `proto` at the end :)

```shell
cat spec.md | docker run -i csi/golang proto
```

akutz:
Hi @saad-ali,

Would you, @jieyu, @julian-hj, and @cpuguy83 please review and approve (or not) this PR? Even if it's not yet in the build pipeline, this PR provides a very simple, standard way for people to generate the bindings locally.

lib/go/spec.sh
```diff
@@ -0,0 +1,24 @@
...
+# Check to see if there is any content in STDIN. If nothing is read
+# after 1 second, then exit with an error.
+if IFS= read -rt 1 line; then
+  echo "$line" > csi.spec
+else
+  echo "usage: cat spec.md | docker run -i golang/csi"
+  exit 1
+fi
+
+while IFS= read -r line; do
```
jdef: could we replace this `while` loop with a simple `cat >> csi.spec`?

Dockerfile-go
```diff
@@ -0,0 +1,69 @@
+FROM golang:1.9.3-alpine as build
```
jdef: why is this dockerfile at the root of the repo, instead of in the lib/go subtree?

saad-ali:

saad-ali:
We should do this at some point though.

saad-ali:

PR #184, opened 2018-02-05
Capabilities and Probe RPCs for Identity service.

### LINKED ISSUE #162: Controller service should be optional
Today there the following calls exist:
* `Controller` service
  * `ControllerGetCapabilities()`
* `Node` services
  * `NodeGetCapabilities()`

A call to `ControllerGetCapabilities()` is required to determine if a plugin implements the `CreateVolume()` or `NodePublishVolume()` calls.

If a volume plugin does not implement any of the `Controller` service capabilities, it should not have to implement the `Controller` service. To enable this we should move `GetCapabilities()` in to the `Identity` service.

### LINKED ISSUE #171: Clarify Identity Service vs Probe request
Today, we have Identity service which establishes the identity and discovers the list of supported versions.
Also, a service can crash and spin back up anytime and the error codes defined in the probe can ideally be returned for any RPC call.
Given that do we really need a Probe call? Can we just use Identity service to establish the identity and readiness of the plugin?

jdef:
* fixes #162 - consolidate plugin capabilities into an Identity-level RPC.
* fixes #171 - consolidate controller/node xxxProbe() calls into an Identity-level RPC.


csi.proto
```diff
@@ -99,6 +96,50 @@ message GetPluginInfoResponse {
...
 ////////
+message GetPluginCapabilitiesRequest {
+  // The API version assumed by the CO. This is a REQUIRED field.
+  Version version = 1;
+}
+
+message GetPluginCapabilitiesResponse {
+  // All the capabilities that the controller service supports. This
+  // field is OPTIONAL.
+  // If any of the `CONTROLLER_xxx` RPC capabilities are RPC then it is
```
jdef: s/are RPC/are specified/

spec.md
```diff
@@ -230,9 +230,9 @@ package csi;
 
 There are three sets of RPCs:
 
-* **Identity Service**: Both the Node Plugin and the Controller Plugin MUST implement this sets of RPCs.
-* **Controller Service**: The Controller Plugin MUST implement this sets of RPCs.
-* **Node Service**: The Node Plugin MUST implement this sets of RPCs.
+* **Identity Service**: All Plugin endpoints MUST provide this set of RPCs.
+* **Controller Service**: Plugins MAY implement this set of RPCs. NOT REQUIRED for all Plugin endpoints.
+* **Node Service**: All Plugin distributions MUST provide this set of RPCs.
```
jdef: Maybe add "NOT REQUIRED for all Plugin endpoints" here too

spec.md
```diff
@@ -363,6 +360,23 @@ The general flow of the success case is as follows (protos illustrated in YAML f
         baz: qaz
 ```
 
+3. CO queries available capabilities of the plugin.
+
+```
+   # CO --(GetPluginInfo)--> Plugin
```
jdef: s/GetPluginInfo/GetPluginCapabilities/`

spec.md
```diff
@@ -1017,7 +1044,7 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
 2. Execute the `ListVolumes` RPC to possibly obtain a volume ID that may be used to execute a `DeleteVolume` RPC; upon success execute `DeleteVolume`.
 3. The CO takes no further action regarding the timed out RPC, a volume is possibly leaked and the operator/user is expected to clean up.
 
-It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capability if it supports the `CREATE_DELETE_VOLUME` capability: the onus is upon the CO to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
+It is NOT REQUIRED for a controller plugin to implement the `CONTROLLER_LIST_VOLUMES` capability if it supports the `CONTROLLER_CREATE_DELETE_VOLUME` capability: the onus is upon the CO to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
```
jdef: s/controller plugin/controller service/

spec.md
```diff
@@ -794,12 +862,12 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
-| Call not implemented | 12 UNIMPLEMENTED | ControllerUnpublishVolume call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` or `NodeGetCapabilities` to discover Plugin capabilities. |
+| Call not implemented | 12 UNIMPLEMENTED | ControllerUnpublishVolume call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `GetPluginCapabilities` to discover Plugin capabilities. |
 
 
 #### `ValidateVolumeCapabilities`
 
-A Controller Plugin MUST implement this RPC call.
+A Plugin MUST implement this RPC call.
```
julian-hj: s/Plugin/Controller service/ ?

I think it is confusing to claim that all plugins must implement this if it is still in the controller service, which is optional.

spec.md
```diff
@@ -494,6 +503,39 @@ message PluginCapability {
...
+```protobuf
+message ProbeRequest {
+  // The API version assumed by the CO. This is a REQUIRED field.
+  Version version = 1;
+}
+
+message ProbeResponse {}
+```
+
+##### Probe Errors
```
jdef: Wondering if we should have use a RESOURCE_EXHAUSTED code if the plugin thinks that Probe() is being called too frequently?

spec.md
jdef: done

csi.proto
```diff
@@ -99,6 +93,58 @@ message GetPluginInfoResponse {
...
+      UNKNOWN = 0;
+      CONTROLLER_SERVICE = 1;
+    }
+  }
+
+  oneof type {
+    // RPC that the plugin supports.
+    RPC rpc = 1;
+    // Service that the plugin supports.
+    Service service = 2;
```
saad-ali: It looks like you're trying to solve the problem of having a way to enable/disable the entire `CONTROLLER` service and have a way to control individual features of the `CONTROLLER` service.

But I'm not quite sure how this will work. If I want to enable the controller service but only have some of the controller capabilities, how would I do that?

I do like having a single PluginCapability instead of each service having it's own. But I'm not sure of a cleaner way to solve the problem above.

jdef: But I'm not quite sure how this will work. If I want to enable the controller service but only have some of the controller capabilities, how would I do that?

see https://github.com/container-storage-interface/spec/pull/184/files#diff-958e7270f96f5407d7d980f500805b1bR360

csi.proto
```diff
@@ -99,6 +93,58 @@ message GetPluginInfoResponse {
...
+  // service capability as well.
+  repeated PluginCapability capabilities = 2;
+}
+
+// Specifies a capability of the plugin.
+message PluginCapability {
+  message RPC {
+    enum Type {
+      UNKNOWN = 0;
+      CONTROLLER_CREATE_DELETE_VOLUME = 1;
```
akutz: I was under the impression that the `PluginCapability` was **only** going to enumerate whether or not the `Controller` or `Node` service was available and that the existing capability RPCs for the eponymous services were going to continue to be used for now. That's what I suggested and @saad-ali agreed in deference to keeping things simple and reducing the required changes at this time. I do not think it's a good idea, nor a necessary one, to pollute a single RPC with capabilities from both Node and Controller services.

jdef:
superseded by #189 and #190

PR #189, opened 2018-02-10
spec: Identity.GetPluginCapabilities RPC

### LINKED ISSUE #162: Controller service should be optional

jdef:
* Introduce "service" capabilities.
* CONTROLLER_SERVICE is an optional service capability.

- supersedes `GetPluginCapabilities` from #184
- fixes #162


saad-ali:
LGTM

akutz:
Ugh, I wish I had seen this. I think the `NodeService` should absolutely be a plug-in capability. Otherwise the implication is that the node service is always available, which is not the case. There are controller-only hosts. Why the omission?

saad-ali:
Node service *IS* always a *plugin capability*. Note the difference between *instance capabilities* and *plugin capabilities*. A particular instance may or may not support node service, but the plugin must support node service (where as a plugin may choose to not implement controller service).

akutz:
> A particular instance may or may not support node service, but the plugin must support node service

Isn't this in contradiction? I'd argue there is zero reason for a centralized controller deployment to support the node service. Wouldn't you agree?

saad-ali:
A plugin is composed of one or more instances. For example, a plugin may be composed of two instances one running in node mode, one running in controller mode. The plugin as a whole implements both node and controller. Hence, "a particular instance may or may not support node service, but the plugin must support node service". 

> I'd argue there is zero reason for a centralized controller deployment to support the node service. Wouldn't you agree?

I can not imagine a use case for a plugin that does not implement node at all and only implements controller.

akutz:
Hi @saad-ali,

> I can not imagine a use case for a plugin that does not implement node at all and only implements controller.

I can. REX-Ray today has a very popular mode where a Controller host is isolated with the controller service configuration necessary to perform volume creation / removal / attach / detach and the node hosts have only the node services. We call this our centralized architecture. It's very similar to what is outlined in the CSI centralized architecture. 

So yes, there are real-world, working examples of Controller-only hosts with no expectation that any Node services (integration with the host) are performed on the host.

cc @codenrhoden @clintkitson

saad-ali:
Ah I think we're both saying the same thing using different words hence the misunderstanding. 

> REX-Ray today has a very popular mode where a Controller host is isolated with the controller service configuration necessary to perform volume creation / removal / attach / detach and the node hosts have only the node services.

I would consider this a plugin composed of two instances one running in node mode, one running in controller mode. So the plugin still implements both a node and a controller service. It's just that on a specific *instance* one or the other is not implemented. That's the key, `PluginCapabilities` is not advertising what the ability of a particular instance is, it is advertising the ability of the plugin as a hole.

akutz:
Hi @saad-ali,

> That's the key, PluginCapabilities is not advertising what the ability of a particular instance is, it is advertising the ability of the plugin as a whole.

Okay. Hmm. I wonder if there is any difference though? 

From the CO perspective that is. If an instance is configured as Controller-only, even though the deployed instance binary may include the Node bits, to the CO that is a **Controller-only** SP. 

I guess my question then becomes what's the difference between `PluginCapabilities` for an instance and the SP as a whole? The RPC cannot be queried if the instance is not running, and therefore it is by default an instance-scoped answer. Otherwise where is the value? 

If the SP is configured to advertise a Controller plug-in capability but the deployed instance is set to `NodeOnly` mode (an SP setting), then the result of querying plug-in capabilities is incorrect. The SP tells the CO that the controller service is available. If your answer is that the SP should detect that the controller service is *not* available and adjust the RPC result then my response is "that is an instance-related result."

Which brings me back to why not have the Node capability as well to indicate that something can be either:

* Node+Controller
* Node
* Controller

(with Identity always implied)

saad-ali:
From the CO perspective, we know weather a given instance is deployed as node or controller, we know a node component will always exist, what we don't know is if a plugin implements a controller or not (is it missing because the plugin doesn't implement it or is it just not deployed when it is supposed to be. We can use this to discover that.

akutz:
Except you don’t because it is t codified into the spec and are implicitly relying on a deployment package that has a relationship to the included services. What if a deployment package can be deployed as Controller, Node, or both at the same time? That has to be configured.

akutz:
*it isn’t codified

akutz:
And I vehemently disagree that a Node component should always exist Saad. In practice I can show you so many examples of controller only deployments.

akutz:
Sorry for the choppy replies. On my phone. I will clean these up later.

saad-ali:
> Except you don’t because it is t codified into the spec and are implicitly relying on a deployment package that has a relationship to the included services. What if a deployment package can be deployed as Controller, Node, or both at the same time? That has to be configured.

And that is perfectly fine. Deployment will take care of ensuring the correct components are deployed to the correct places with the correct initialization configuration information. As a CO, I don't have to worry about that, so it doesn't need to be part of the spec.

> And I vehemently disagree that a Node component should always exist Saad. In practice I can show you so many examples of controller only deployments.

I think this is a repeat of the "we're both saying the same thing using different words". A single component of the plugin instance may be deployed as controller only or node only or both. The plugin as a whole must be either node only or both controller and node, but never controller only -- a CSI plugin must have a node component deployed on every node to do the last mile binding. And that's the bit that I, as the CO, care about, does this plugin as whole implement a controller or not. I will never ask it if the plugin as a whole implements a node or not, because not implementing a node is invalid.

akutz:
Hi @saad-ali,

I don't think we're saying the same thing. It also doesn't matter any more.

Hi @jdef,

There is some problem with the way the protobuf was written for adding this feature. Some how the enum never got added or wired to the result. See the following:

```go
func (s *service) GetPluginCapabilities(
	ctx context.Context,
	req *csi.GetPluginCapabilitiesRequest) (
	*csi.GetPluginCapabilitiesResponse, error) {

	return &csi.GetPluginCapabilitiesResponse{
		Capabilities: []*csi.PluginCapability{
			&csi.PluginCapability{
				Type: &csi.PluginCapability_Service_{
					Service: &csi.PluginCapability_Service{
						
					},
				},
			},
		},
	}, nil
}
```

One would expect `csi.PluginCapability_Service` to have the ability to set the enum, right? From `csi.pb.go`:

```go
type PluginCapability_Service struct {
}
```

Not so much. Just thought you should know. Right now this new RPC is broken as it cannot return the enum that indicates the capability.

akutz:
Hi @jdef,

Also, just search `csi.pb.go` for the enum type alias `PluginCapability_Service_Type`. It's not the type of any field associated with any response.

akutz:
Hi @jdef,

By the way, my suspicion? You used a reserved keyword in Go for the `oneof` field name:

```proto
// Specifies a capability of the plugin.
message PluginCapability {
  message Service {
    enum Type {
      UNKNOWN = 0;

      // CONTROLLER_SERVICE indicates that the Plugin provides RPCs for
      // the ControllerService. Plugins SHOULD provide this capability.
      // In rare cases certain plugins may wish to omit the
      // ControllerService entirely from their implementation, but such
      // SHOULD NOT be the common case.
      // The presence of this capability determines whether the CO will
      // attempt to invoke the REQUIRED ControllerService RPCs, as well
      // as specific RPCs as indicated by ControllerGetCapabilities.
      CONTROLLER_SERVICE = 1;
    }
  }

  oneof type {
    Service service = 1;
  }
}
```

Change that to `oneof service_type` and I bet everything works.

PR #192, opened 2018-02-15
Remove all SP Implementation Info

akutz:
This patch removes all of the SP implementation information previously detailed in the specification. With respect to issue #171 and comments made by @saad-ali, @childsb, @jdef, and others, it has become apparent that the specification should not supply any suggestions or restrictions on matters related to an SP's packaging, deployment, or runtime configuration.

Removed information includes, but is not limited to the following:

| Removed | Description |
|----------|-------------|
| `CSI_ENDPOINT` | SPs are now responsible for defining the manner by which they grok the gRPC endpoint on which they serve the CSI services. CO admins are now responsible for deploying and configuring SPs using that SP's documentation. |
| Environment variable names | The spec no longer claims ownership of the `CSI_` prefix for environment variables. |
| Permissions | The spec no longer asserts any recommendations with respect to container permissions that an SP might require. |
| Logging | All logging details are the responsibility of the SP and may vary SP to SP. |

As a reminder, the above list is not exhaustive or complete. Please see the patch for the full picture.

# Guide to Developing CSI Storage Plug-ins
The removed information may very well be suitable to live in a separate document that provides guidelines for SP creation and configuration, such as common environment variable names, logging and packaging details, etc.

cc @codenrhoden

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -653,12 +653,12 @@ message VolumeCapability {
 message CapacityRange {
   // Volume must be at least this big. This field is OPTIONAL.
   // A value of 0 is equal to an unspecified field value.
-  // The value of this field MUST NOT be negative. 
+  // The value of this field MUST NOT be negative.
   int64 required_bytes = 1;
 
   // Volume must not be bigger than this. This field is OPTIONAL.
   // A value of 0 is equal to an unspecified field value.
-  // The value of this field MUST NOT be negative. 
+  // The value of this field MUST NOT be negative.
   int64 limit_bytes = 2;
 }
 
@@ -667,7 +667,7 @@ message Volume {
   // The capacity of the volume in bytes. This field is OPTIONAL. If not
   // set (value of 0), it indicates that the capacity of the volume is
   // unknown (e.g., NFS share).
-  // The value of this field MUST NOT be negative. 
+  // The value of this field MUST NOT be negative.
   int64 capacity_bytes = 1;
 
   // Contains identity information for the created volume. This field is
@@ -952,7 +952,7 @@ message ListVolumesRequest {
   // in the subsequent `ListVolumes` call. This field is OPTIONAL. If
   // not specified (zero value), it means there is no restriction on the
   // number of entries that can be returned.
-  // The value of this field MUST NOT be negative. 
+  // The value of this field MUST NOT be negative.
   int32 max_entries = 2;
 
   // A token to specify where to start paginating. Set this field to
@@ -1020,7 +1020,7 @@ message GetCapacityResponse {
   // specified in the request, the Plugin SHALL take those into
   // consideration when calculating the available capacity of the
   // storage. This field is REQUIRED.
-  // The value of this field MUST NOT be negative. 
+  // The value of this field MUST NOT be negative.
   int64 available_capacity = 1;
 }
 ```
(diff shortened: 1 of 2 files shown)


saad-ali:
I really wish we could make this change. Decoupling the spec completely from packaging and deployment would make it cleaner and more flexible. But as mentioned in https://github.com/container-storage-interface/spec/issues/171#issuecomment-366117760, Mesos has a dependency on `CSI_ENDPOINT`, so we can't completely do away with all of it. Maybe we can whittle it down some.

akutz:
Hi @saad-ali,

If we're going to bend the API specification to include implementation details for a **single** CO then it's clear that the API specification includes implementation details. For example, `CSI_ENDPOINT` is going to be the way **all** SPs and COs communicate the endpoint. It's not a de facto standard, it's _de jure_. If this is the case then the specification should absolutely define other configuration properties almost guaranteed to be common across all SPs, such as `CSI_MODE`.

When the specification defines a single SP environment variable the specification now owns a prefix that is for what? For nothing else? If the spec defines anything to do with implementation details then it should define the rest of the common ones too, such as `CSI_MODE`. Because just as Mesos deploys the SPs with `CSI_ENDPOINT`, some SPs will also need the equivalent of `CSI_MODE`, so why not make it a standard?

saad-ali:
> If we're going to bend the API specification to include implementation details for a single CO then it's clear that the API specification includes implementation details. For example, CSI_ENDPOINT is going to be the way all SPs and COs communicate the endpoint. It's not a de facto standard, it's de jure. If this is the case then the specification should absolutely define other configuration properties almost guaranteed to be common across all SPs, such as CSI_MODE.

Defining endpoint as an env var is an exception not the rule.

> When the specification defines a single SP environment variable the specification now owns a prefix that is for what? For nothing else? If the spec defines anything to do with implementation details then it should define the rest of the common ones too, such as CSI_MODE. Because just as Mesos deploys the SPs with CSI_ENDPOINT, some SPs will also need the equivalent of CSI_MODE, so why not make it a standard?

AFAIK endpoint is critical for CSI to work with Mesos. But a single exception should not be justification to open the flood gates and abandon the principle altogether. Separating packaging and deployment from the interface is important and we should strive to do so as we continue to expand and revise the spec.

akutz:
Hi @saad-ali,

If `CSI_ENDPOINT` is an exception then at least remove the language from the spec that declares `CSI_` off-limits. Otherwise a very useful and obvious namespace is completely unused except for a single environment variable.

saad-ali:
> If CSI_ENDPOINT is an exception then at least remove the language from the spec that declares CSI_ off-limits. Otherwise a very useful and obvious namespace is completely unused except for a single environment variable.

That seems reasonable to me. @jieyu?

jdef:
IMO as long as the spec mandates the use of a CSI_xxx envvar then we should
keep the namespace reserved.

akutz:
This is having your cake and eating it too. If `CSI_ENDPOINT` is an exception then the spec should not declare the namespace reserved unless the intent is to provide additional _de jure_ environment variable names. And if that's the case, then provide `CSI_MODE` for starters.

jdef:
We'd like to keep the envvar surface area small. `CSI_ENDPOINT` is the only envvar for which the spec currently defines meaning. If we open up the `CSI_.*` envvar namespace in general, and then later need to add `CSI_FOO` at some point to address a critical use case, how can we reliably do that w/o possibly stomping on 3rd parties?

Just because I think the spec should continue to reserve the CSI_ envvar namespace doesn't imply that the spec **needs** to define any additional envvars other than what's already in the spec. The bar for introducing any additional environment requirements should be **very** high .. and CSI_MODE doesn't clear that bar for me.

spec.md
```diff
@@ -1365,116 +1365,18 @@ If the plugin is unable to complete the NodeGetCapabilities call successfully, i
   * proto3 SHOULD be used with gRPC, as per the [official recommendations](http://www.grpc.io/docs/guides/#protocol-buffer-versions).
   * All Plugins SHALL implement the REQUIRED Identity service RPCs.
     Support for OPTIONAL RPCs is reported by the `ControllerGetCapabilities` and `NodeGetCapabilities` RPC calls.
-* The CO SHALL provide the listen-address for the Plugin by way of the `CSI_ENDPOINT` environment variable.
-  Plugin components SHALL create, bind, and listen for RPCs on the specified listen address.
-  * Only UNIX Domain Sockets may be used as endpoints.
-    This will likely change in a future version of this specification to support non-UNIX platforms.
-* All supported RPC services MUST be available at the listen address of the Plugin.
+* All Plugins MUST support UNIX Domain Sockets as a gRPC endpoint.
```
jdef: I like this statement.

spec.md
```diff
@@ -1365,116 +1365,18 @@ If the plugin is unable to complete the NodeGetCapabilities call successfully, i
...
-#### Operational Recommendations
-
-The Plugin Supervisor expects that a Plugin SHALL act as a long-running service vs. an on-demand, CLI-driven process.
-
-Supervised plugins MAY be isolated and/or resource-bounded.
-
-##### Logging
-
-* Plugins SHOULD generate log messages to ONLY standard output and/or standard error.
-  * In this case the Plugin Supervisor SHALL assume responsibility for all log lifecycle management.
```
jdef: instead of removing this block, we could change SHALL to SHOULD

spec.md
```diff
@@ -1365,116 +1365,18 @@ If the plugin is unable to complete the NodeGetCapabilities call successfully, i
...
-* Plugins SHOULD NOT write potentially sensitive data to logs (e.g. `Credentials`).
-
-##### Available Services
-
-* Plugin Packages MAY support all or a subset of CSI services; service combinations MAY be configurable at runtime by the Plugin Supervisor.
-* Misconfigured plugin software SHOULD fail-fast with an OS-appropriate error code.
-
-##### Linux Capabilities
-
-* Plugin Supervisor SHALL guarantee that plugins will have `CAP_SYS_ADMIN` capability on Linux when running on Nodes.
```
jdef: we need this so that plugins can execute mount operations. i don't think it's wise to remove from the spec.

cpuguy83: Agree it's ok to remove, this is a packaging concern and not related to how the CO and the SP can interact.

saad-ali:

saad-ali:

Issue #193, opened 2018-02-16
Consider adding vMAJOR version (e.g. csi.v1.) to either the API bindings or package names


saad-ali:
Based on @akutz's proposal in: https://github.com/container-storage-interface/spec/issues/171#issuecomment-364543778

While this is not a 0.2 blocker, it should be considered before 1.0.

saad-ali:
I would like to propose cutting a 0.2 release by end of next week so we can pick up 0.2 for Kubernetes. Can we get this in time for that? If so, I have no problem with it. @jdef @jieyu @julian-hj @cpuguy83 for their thoughts.

akutz:
Hi @saad-ali,

You're a eager beaver, aren't ya? Man, I'm just getting SPs up to snuff with 0.1.0 and here you go with 0.2.0 :)

akutz:
Hi @saad-ali,

I can get a PR submitted for people to look at. At least that will spur some discussion.

saad-ali:
Ha, yes, I'd like Kubernetes to ship with 0.2 since there are so many breaking changes between 0.1 and 0.2.

cpuguy83:
Do we need to version each pre-release package? My inclination would be to just keep it at v1 and v1 won't be stable until we actually hit the v1 release, or alternatively v1-dev and bump to v1 when the protos are finalized.

I doubt CO's will suport multiple pre-release versions at the same time.
I suppose it can make it easier for SP's where different CO's may be supporting a different version, though.

jdef:
I'd rather not see versioned pre-release packages. AFAIK the goal was to
provide a mechanism for dealing with breaking API changes post-v1. Pre-v1
we're all working under the assumption that there will be periodic API
breakage (though we will try very hard to void breakage after v0.2)

akutz:
I prefer we version each pre release because some of us do have to keep up with dozens of instances of this stuff. A compromise would be to treat the prerelease versions like release versions so 0.1 is v1, and just understand that upon release everything starts over.

jdef:
Strongly disagree that v0.1 should be modeled as v1

akutz:
Hi @jdef,

Which is why in my PR I allowed for a leading zero if the MAJOR component was zero. For example:

> `csi.vMAJOR(MINOR)?` where `MINOR` is added if `MAJOR` is `0`. For example:
> 
> | Version | Package Name |
> |--------|----------------|
> | `v0.2.0` | `csi.v02` |
[...]

PR #194, opened 2018-02-16
Version the CSI package name

akutz:
This patch fixes issue #193 and adds changes the CSI package name from `csi` to `csi.vMAJOR(MINOR)?` where `MINOR` is added if `MAJOR` is `0`. For example:

| Version | Package Name |
|--------|----------------|
| `v0.2.0` | `csi.v02` |
| `v2.1.0` | `csi.v2` |

The leading zero and subsequent `MINOR` component is necessary to prevent collision with the eventual releases `v1.0`, `v2.0`, etc.

## Accessing Older Spec and Language Binding Files
Please note this patch does not address accessing older versions of the specification. The generated files such as `csi.proto` and `lib/go/csi/csi.pb.go` are not in versioned paths or have versioned file names. This is on purpose. Maintaining all previous versions at the `HEAD` of the repository would result in a convoluted generation process as well as questions about whether `lib/go/csi/v2` is `v2.0`, `v2.1.5`, `v2.3`, etc.

People that wish to access older versions of the specification can do so by using `git` or GitHub's UI to fetch a specific tag or commit and manually copying the protobuf or language binding locally. 

At least that's my thought on the matter. I elected to keep it simple (for now).


akutz:
Please note that this affects all languages, not just Go, since the protobuf type names now include the version. For example, `csi.v02.CreateVolumeRequest`.

akutz:
Replaced by #195.

PR #195, opened 2018-02-16
Versioned proto and golang package structure

### LINKED ISSUE #193: Consider adding vMAJOR version (e.g. csi.v1.) to either the API bindings or package names

jdef:
* `GetSupportedVersions` RPC moves from `Identity` to a new `Versioned` gRPC service
* fixes #193 
  * spec major version name is incorporated into the generated proto package
* generated protobuf files are as follows:
  * csi.proto: contains versioned gRPC services and protobuf messages
  * versioned.proto: contains `Versioned` service and related proto messages. SHOULD never change across CSI revisions. MUST never change in ways that break backward/forward across any releases of CSI
* generated Go bindings are packaged in a `csi/v1` subdirectory
  * importing code may still refer to bindings using `csi` package prefix by default

alternative to #194


akutz:
Maybe I’m confused, but why submit your own PR instead of supplying feedback to the one I already submitted?

jdef:
I wrote my changes before seeing yours. It was less effort for me to simply
push an alternate PR for the code I had already written vs. rewrite my PR
as a list of suggestions. Feel free to incorporate changes from this PR to
maintain authorship.

akutz:
Hi @jdef,

Let's just go with yours :)

csi.proto
```diff
@@ -1,6 +1,8 @@
 ////////
 syntax = "proto3";
-package csi;
+package csi.v1;
+
+option go_package = "csi";
```
akutz: Hi @jdef,

Please consider adding the following after `option go_package`:

```proto
option java_package="csi";
option csharp_namespace="csi";
```

akutz: Hi @jdef,

Please consider adding the following after `option go_package`:

```proto
option java_package="csi";
option csharp_namespace="csi";
```

lib/go/Makefile
```diff
@@ -76,10 +76,11 @@ export PATH := $(shell pwd):$(PATH)
 ########################################################################
 ##                              BUILD                                 ##
 ########################################################################
-CSI_GO := csi/csi.pb.go
-CSI_A := csi.a
 CSI_PROTO := ../../csi.proto
```
akutz: Hi @jdef,

If the language binding's path/package structure is going to be versioned then shouldn't `csi.proto` as well?

akutz:
Hi @julian-hj,

You make a good point regarding the `GetSupportedVersions` call. My suggestion? Create two protos and generate the Identity service without a versioned component and ensure that all RPCs and messages remain as static as possible. It could be overkill, but it's an idea. If nothing else, I see no reason that it would be difficult to answer to multiple versions of a `GetSupportedVersions` call on the server side.

lib/go/Makefile
jieyu: I believe this is necessary for C++ to version the `.proto` file.

jdef:
> I am a bit troubled by the interaction between this change and the GetSupportedVersions RPC. To me it seems very awkward to have to call v1.GetSupportedVersions to find out whether or not v1 is supported.

AFAIK the intent is that GetSupportedVersions SHALL NOT ever change in a way that's breaking/incompatible with prior or future versions of the spec. This allows a CO to query *any* plugin for its supported versions, regardless of the spec version that the plugin has implemented. If needed, we can break out GetSupportedVersions into a separate service but I'm not convinced that we're there yet, and I don't see the harm in leaving it where it is for the moment.

That said, other message types are certainly expected to break across major spec revisions and namespacing the messages by major version gives us the freedom to both eliminate deprecated/obsolete things as well as restructure the RPCs to better fit the needs of the time.

> As I addressed in #194, I am not sure that placing the generated language bindings in a versioned package path is the best solution.

I decided to place the generated bindings in a versioned package path because it allows CO and SP implementations to easily support multiple variants of CSI at the same time. Relying upon ONLY GH-versioning/repo-tagging for this leads to wacky solutions that don't necessarily scale well over time.

Currently thinking about the best way to version the `.proto` and `spec.md` files: if we need to generate all versions with every `make` then we'll need something more flexible than the scheme currently in place.

julian-hj:
Maybe I am missing something, but to me, we lose the benefit of not changing GetSupportedVersions once we put it into a versioned namespace.

Say theoretically, we have a Plugin that supports V2, and a CO that can consume either V1 or V2.  The CO can't just call GetSupportedVersions anymore to find out which version of the client to use.  It instead has to try to invoke "/csi.v1.Identity/GetSupportedVersions", and when that fails, it has to try to use the V2 service version.  So *most* of the version support information will be determined by the success or failure of the RPC.

I suppose a plugin author could choose to implement multiple identity service versions, just to be callable from incompatible COs, but I doubt any plugin author would bother.

Does this make sense, or am I working from wrong assumptions?

jdef:
If GetSupportedVersions is wire compatible for prior and future versions of
the spec, what stops a v1 GetSupportedVersions call from working against a
server that was implemented against a v2 spec?

julian-hj:
@jdef Isn't the whole point of namespacing that we will create a separate set of RPCs?  My read is that it wouldn't work because it is a different RPC.

See the following diff in the grpc client code:
https://github.com/container-storage-interface/spec/pull/195/files#diff-9f98c96192effffd9a3d33831d159befR1984

saad-ali:
> @jdef Isn't the whole point of namespacing that we will create a separate set of RPCs? My read is that it wouldn't work because it is a different RPC.

I believe that since `option go_package = "csi";` is specified the go package will still be `csi` not `csi.v1`. +1 to @akutz comment on doing the same for other languages. The code you linked to is just registering the proto.

That said, I do have questions about this:

> I decided to place the generated bindings in a versioned package path because it allows CO and SP implementations to easily support multiple variants of CSI at the same time. Relying upon ONLY GH-versioning/repo-tagging for this leads to wacky solutions that don't necessarily scale well over time.
> 
> Currently thinking about the best way to version the .proto and spec.md files: if we need to generate all versions with every make then we'll need something more flexible than the scheme currently in place.

In order for this to work HEAD needs to be to maintain all previous versions of the generated bindings in separate folders. What is the plan for managing that? Would we cut to

jdef:
> I believe that since option go_package = "csi"; is specified the go package will still be csi not csi.v1. +1 to @akutz comment on doing the same for other languages. The code you linked to is just registering the proto.

I added the option for Go package generation because we include Go bindings in this repo. C++ doesn't have an option like this, instead the C++ namespace is derived from the protobuf package name. No other languages have build files in this repo. Does it make sense to include protoc options for langs that we don't have a way to test/build in CI?

jdef:
> Couldn't we leave GetSupportedVersions in v1 and use it to determine what features are available from v 1.x? Then when v2 comes around we can do the same for v2.

We *could* but that kind of defeats the purpose of `GetSupportedVersions` as mentioned by @julian-hj . The original intent of the RPC was to support version discovery/negotiation across breaking versions of the spec. My thinking is that we should keep it that way.

csi.proto
jdef: As mentioned later in this thread, I think this only makes sense to do once we're generating bindings within the spec repo for these langs, otherwise it may be difficult to predict the full impact of changes on bindings generated from the spec.

julian-hj:
The new `Versioned` service answers my main objection to this PR, so I am OK with it now as-is, provided that others are also OK with it.  But I am sort of starting to think we might be better off just removing the GetSupportedVersions rpc altogether.  Between this call and the Probe call and the GetCapabilities calls, we are building up a fair amount of cruft that's not really related to the actual work the plugin has to do, and could be reasonably relegated to documentation.

cpuguy83:
I would prefer to not take version negotiation out of the spec, otherwise this will be adding to what the SP needs to figure out for each CO.

saad-ali:
> I would prefer to not take version negotiation out of the spec, otherwise this will be adding to what the SP needs to figure out for each CO.

Having a SP expose a single version per endpoint does simplify implementation, but it also limits SP flexibility.

julian-hj:
> I would prefer to not take version negotiation out of the spec, otherwise this will be adding to what the SP needs to figure out for each CO.

I wasn't suggesting removal of version negotiation entirely. I was merely questioning whether or not GetSupportedVersions offers the CO any useful insights about version support that it cannot get by just calling Probe() with the version it will actually use.  From the SP side, I don't think removing GetSupportedVersions() would really change anything except to reduce the number of RPCs (and now services) they have to implement by 1.  Anyway, I hope we can all discuss it in a few minutes.

versioned.proto
```diff
@@ -0,0 +1,24 @@
...
+  rpc GetSupportedVersions (GetSupportedVersionsRequest)
+    returns (GetSupportedVersionsResponse) {}
+}
+message GetSupportedVersionsRequest {
+}
+
+message GetSupportedVersionsResponse {
+  // All the CSI versions that the Plugin supports. This field is
+  // REQUIRED.
+  repeated Version supported_versions = 1;
```
cpuguy83: Since we're changing this, we've found that we need to know what the min and max supported version so that this can be negotiated.

jieyu:
Resolved by #204

PR #196, opened 2018-02-16
Fix missing Type field from PluginCapability

akutz:
This patch fixes a missing field associated with the GetPluginCapabilities response from #189.

There is a problem with the way the protobuf was written. Somehow the enum never got added or wired to the result. See the following:

```go
func (s *service) GetPluginCapabilities(
	ctx context.Context,
	req *csi.GetPluginCapabilitiesRequest) (
	*csi.GetPluginCapabilitiesResponse, error) {

	return &csi.GetPluginCapabilitiesResponse{
		Capabilities: []*csi.PluginCapability{
			&csi.PluginCapability{
				Type: &csi.PluginCapability_Service_{
					Service: &csi.PluginCapability_Service{
						
					},
				},
			},
		},
	}, nil
}
```

One would expect `csi.PluginCapability_Service` to have the ability to set the enum, right? From `csi.pb.go`:

```go
type PluginCapability_Service struct {
}
```

This patch adds the service type to the service message.


akutz:
Since this has approvals from two COs please merge?

saad-ali:
LGTM merging

PR #200, opened 2018-02-22
Simplify credentials from 8 to 4

### LINKED ISSUE #198: Credentials difficult for COs to expose

saad-ali:
Fixes #198 

* Simplify by having one secret per operation pair (reducing 8 possibilities to 4):
    1. `provisioner_secrets`
        * Grants ability to create and delete volume requests.
        * Passed to `CreateVolume()` and `DeleteVolume()`.
    2. `controller_publish_secrets`
        * Secrets required to complete controller publish and unpublish calls.
        * Example: an access key for volume.
        * Passed to `ControllerPublishVolume()` and `ControllerUnpublishVolume()`
    3. `node_stage_secrets`
        * Secrets required to complete node stage call.
        * Example: a decryption key for volume.
        * Passed to `NodeStageVolume()` only.
        * Not passed to `NodeUnstageVolume()` which we can consider adding in the future, if needed.
    3. `node_publish_secrets`
        * Secrets required to complete node publish call.
        * Example: end user credentials.
        * Passed to `NodePublishVolume()` only
        * Not passed to `NodeUnpublishVolume()` which we can consider adding in the future, if needed.


csi.proto
```diff
@@ -405,18 +415,24 @@ message ControllerUnpublishVolumeRequest {
...
-  // This field contains credential data, for example username and
-  // password. Each key must consist of alphanumeric characters, '-',
-  // '_' or '.'. Each value MUST contain a valid string. An SP MAY
-  // choose to accept binary (non-string) data by using a binary-to-text
-  // encoding scheme, like base64. An SP SHALL advertise the
-  // requirements for credentials in documentation. COs SHALL permit
-  // passing through the required credentials. This information is
-  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
+  // Secrets required by plugin to complete controller unpublish volume
+  // request. This should be the same secrets passed to the
```
jdef: nit: s/should/SHOULD/

csi.proto
```diff
@@ -196,18 +196,21 @@ message CreateVolumeRequest {
...
-  // choose to accept binary (non-string) data by using a binary-to-text
-  // encoding scheme, like base64. An SP SHALL advertise the
-  // requirements for credentials in documentation. COs SHALL permit
-  // passing through the required credentials. This information is
-  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
+  // Secrets required by plugin to complete volume creation request.
+  // A secret is a string to string map where the key identifies the
+  // name of the secret (e.g. "username" or "password"), and the value
+  // contains the secret data (e.g. "bob" or "abc123").
+  // Each key must consist of alphanumeric characters, '-', '_' or '.'.
```
jdef: nit: s/must/MUST/ (here and throughout changeset)

spec.md
```diff
@@ -1636,7 +1631,7 @@ Supervised plugins MAY be isolated and/or resource-bounded.
   * Logging configuration flags and/or variables, including working sample configurations.
   * Default log destination(s) (where do the logs go if no configuration is specified?)
   * Log lifecycle management ownership and related guidance (size limits, rate limits, rolling, archiving, expunging, etc.) applicable to the logging mechanism embedded within the Plugin.
-* Plugins SHOULD NOT write potentially sensitive data to logs (e.g. `Credentials`).
+* Plugins SHOULD NOT write potentially sensitive data to logs (e.g. `Secrets`).
```
jdef: nit: there are no types or fields called `Secrets`. maybe just replace "`Secrets`" with "secrets"?

csi.proto
saad-ali: Done.

csi.proto
saad-ali: Done.

spec.md
saad-ali: Done.

saad-ali:
Feedback addressed. PTAL

PR #204, opened 2018-02-23
Drop GetSupportedVersions in favor of versioned proto package

### LINKED ISSUE #193: Consider adding vMAJOR version (e.g. csi.v1.) to either the API bindings or package names

jieyu:
As per the CSI discussion today regarding versioning:
* alternative to #195 
  * we can add a `Versioned` service later if needed; doing it now isn't critical and it will be hard to remove later on if we decide we don't need it.
* fixes #193 
  * allows v0.1 and v0.2 bindings to co-exist (we're aiming for v0.2 to be the last breaking API change set prior to v1.0)
  * used "v0" for the proto package name
  * removed GetSupportedVersions RPC as well as per-RPC request `Version version = 1` field specs
* small Makefile changes to clean up generated code

based on and supersedes #202


jieyu:
@saad-ali @julian-hj @cpuguy83 PTAL

I squashed and rebased. Squash is because there are too many conflicts. cc @jdef

saad-ali:
LGTM

akutz:
Wow, this seems drastic and potentially harmful.

Why remove this RPC that is in fact quite useful? API changes cannot be completely dictated by the potential language bindings. While I agree that changes to support bindings are good, removing a feature because a specific binding may not utilize is in fact removing that utility for all languages that **can** make use of it.

A CSI endpoint could be in any language and so could a client. A client, for example, that supports multiple API versions. How is a client now able to query the endpoint about what versions are supported? That is completely gone? This is not a good change. Even post 1.0 there will be breaking changes going forward, and now there is simply no way to know what an endpoint supports.

While it's a good idea to version the proto package, the ability to query what versions are supported was fundamental. I'm gone for one meeting and this happens? *sigh* :)

Also, I've mentioned this several times to no avail, but I'll try again. Why was the following omitted?

```proto
option go_package = "csi";
option java_package="csi";
option csharp_namespace="csi";
```

This is frustrating. I recommended this change in #195 and it was ignored. It was recommended and ignored offline as well. Recommendations deserve at least an acknowledgement, even if they are ultimately ignored.

akutz:
Hi @jieyu,

Also, please simplify the packaging path from `lib/go/csi/v0` to `lib/go/v0`. The `csi` component is vestigial and should be removed.

akutz:
Hi @jieyu,

I am really concerned that you neglected to separate the CSI proto/packaging changes and the RPC/Version message changes into multiple commits. I would see the latter completely reverted as it serves no purpose that I'm able to see. However it also requires reverting the changes of value in this PR. I think everyone in this project could do well to start paying more attention to good Git practice. There have been a few PRs lately where commits should be separate.

cpuguy83:
@akutz we removed it with the intention of working towards version negotiation. We didn't feel the existing rpc was good enough and wanted to lump the breaking change into 0.2 and add something else (or potentially even the same thing, who knows) as an additive/non-breaking change before 1.0.

akutz:
Hi @cpuguy83,

Ah. That aspect of your discussion is not represented clearly in the PR description. Thank you very much. That makes sense.

akutz:
I just want to add that all other reductive changes to strip the spec down should be managed ASAP. The fact is that these changes have broad, downstream impact. Entire types are removed, rendering libraries and programs unable to even compile until refactoring occurs.

akutz:
Hmm, no what would make lives easier? Reintroduce the version message type. At least then middleware, helper funcs, etc. that rely on csi.Version don’t immediatelt break.

akutz:
Hi @cpuguy83,

One quick update on your comment regarding _version negotiation_. I just wanted to point out that very early in the CSI version discussion last year I [recommended moving](https://github.com/container-storage-interface/spec/issues/65#issuecomment-321085582) the version negotiation to the gRPC metadata for __this very reason__:

> Hi @jsafrane and @jieyu,
> 
> The gRPC folks expect that your code should always be backwards compatible. Besides, gRPC itself can largely handle additive changes. A message with a single field will parse data that has two fields by ignoring the second field.
> 
> To @jsafrane's point about breaking changes, perhaps we should consider relocating the version field into the gRPC message metadata, which can then be parsed out via a gRPC interceptor.
> 
> That way the message is approved or rejected before it is ever routed to the correct (or rather intended) handler.
> 
> Please see the GoCSI client for an example on injecting a message with metadata and the server that extracts said data.

This change would allow multiple versions of the API to still handle version negotiation. There is a reason that this method is quite common in HTTP-based APIs. Headers are not bound to the API model and are useful for handling contract negotiation.

cc @saad-ali

jdef:
I thought that I explained why I hadn't included the option package name
declarations for java,  etc. Maybe that was on another thread. Reason was
because we don't host bindings for those langs yet and can't see/test the
impact of such lang specific options in the repo.

That said, I'm open to the idea of the spec repo hosting bindings for other
langs.

akutz:
Hi @jdef,

If you did then I missed it. It’s likely you did, and I just overlooked your response. Thank you, and while I agree with the reasoning, if you add the suggested options then people consuming the proto in its raw form won’t have to modify it to get the equivalent package name. However, I see your viewpoint as well.

akutz:
Let’s rememeber that the proto isn’t bound to the rules/results of this repo or what it hosts, but rather that the proto should assume anyone may wish to consume it and generate language bindings for any language.

jdef:
Also, java package names use reverse domain name notation, I'm not sure
that "csi" is appropriate.

And .NET has its own set of rules for namespaces:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces
-- again, I'm not sure that "csi" is appropriate there either.

jdef:
Additionally, Java doesn't support package aliasing (and neither do all
.NET langs) so I'm not convinced that we should use any package naming
options for those platforms.

akutz:
I'm aware of each. i figured it was just the final path component like Go since Java's notation is filesystem based as well as .NET's.

jdef:
I'm also a bit sad that the commit stream was squashed for this. Let's try to do better going forward.

akutz:
Hi @jdef,

So I'm blaming lack of sleep, because I have six years of experience with .NET (C#) and five-ish with Java, and I apparently completely forgot both required the full namespace path at the top of their respective source files.

I even double checked my own sources to verify. 

Damn.

(take a step away from the laptop Andrew)

Thanks for doing the due diligence on this I neglected to do.

PR #188, opened 2018-02-09
Introduce concept of topology to CSI spec.

saad-ali:
Volumes made available by an arbitrary storage provider (SP) may be partitioned such that the volumes are not equally available throughout the cluster the CO is managing. Instead volumes may be constrained to some subset of node(s) in the cluster. We will refer to the segmenting of these clusters (into “racks” in on premise clusters or into “regions” and “zones” in cloud environments, for example) as topology.

This PR introduces the concept of topology to the CSI spec.

Modern Container Orchestration Systems (COs) enable dynamic provisioning of volumes on demand when a workload is scheduled, and also enable pre-provisioned volumes to be used by workloads. In order to make effective scheduling and provisioning decisions, COs need to be able to both influence and act on topology information for each volume.

If volumes exposed by an SP are equally available from every where in the cluster that the CO is managing, then the CO doesn’t need to influence where a workload is scheduled or where a volume is provisioned, and no topology information needs to be shared between SP and CO.

An SP may also have a cluster topology different from (independent of) the topology of the cluster where work is being scheduled. In that case, because the volumes are still available everywhere in the cluster that the CO is managing, the CO doesn’t need to influence where a workload is scheduled or where a volume is provisioned. Regardless, the CO or end user may still want to influence where a volume is provisioned. To enable this, the SP can simply expose a mechanism via an opaque parameter in the `CreateVolume` call (the Container Storage Interface enable SPs to expose an arbitrary set of configuration options during dynamic provisioning by allowing opaque parameters to be passed from cluster admins to the storage plugins, to control things like `diskType` or in this case, maybe some storage system specific topology).

For many storage systems, volumes are not equally accessible from all nodes in the cluster the CO is managing. And therefore, the CO must be able to retrieve the topological location of a volume and use it to influence where the workload referencing it can be scheduled or, vice versa: use the location of a workload to influence where a volume for that workload should be provisioned. Therefore, a standardized mechanism for sharing topological information between COs and SPs must be defined.

There exists an arbitrary number of topologies that may differ from cluster to cluster, and a single cluster may contain multiple topologies, so this solution must be able to handle topology generically (not hard code implementation specific concepts like regions, zones, etc. as proposed in https://github.com/container-storage-interface/spec/issues/7)

Closes https://github.com/container-storage-interface/spec/issues/43

PR #207, opened 2018-03-01
Add Snapshot Support in CSI Spec

### LINKED ISSUE #208: Add Snapshot Support in the CSI Spec
Many storage systems (cloud providers and other storage backends) support the volume snapshot feature as a way to protect data stored on the persistent volumes.  In the case when the original volume gets corrupted,  a volume can be created with the data restored from the point-in-time snapshot to prevent data loss.  Currently CSI does not have support for snapshot.  This issue is raised to add snapshot functionalities in the CSI spec.

A PR is submitted here: #224

xing-yang:
This PR proposes to add snapshot support in CSI specification.
It proposes to add 3 new controller service RPCs, including
CreateSnapshot, DeleteSnapshot, and ListSnapshots,
and modify the existing ControllerGetCapabilitiesResponse and
CreateVolumeRequest.

There are several cases for using the snapshot data in a volume:

1. Create volume from snapshot: In this case, a new volume will be
created on the storage backend and the data in the snapshot will
be copied to the new volume. This case will be addressed by this
proposal.

2. Revert snapshot: In this case, data in the original volume
will be erased and replaced with data in the snapshot. This case
is out of scope for this proposal and should be handled by a
function of its own (not to be mixed with create volume from
snapshot) if it is proposed in the future.

The following changes will be made in the Service Controller:
- Add CreateSnapshot
- Add DeleteSnapshot
- Add ListSnapshots
- Modify CreateVolumeRequest to support creating a new volume from
  snapshot.
- Modify ControllerGetCapabilitiesResponse to return
  CREATE_DELETE_SNAPSHOT, LIST_SNAPSHOTS, 
  and CREATE_VOLUME_FROM_SNAPSHOT support capabilities.

This proposal was drafted here:
https://docs.google.com/document/d/1oXVuDTNhXerr1UJ48tQrxF9J6vbm2LHVVC3z28655Qo/edit#

Closes #208


csi.proto
```diff
@@ -502,6 +529,179 @@ message ControllerServiceCapability {
     RPC rpc = 1;
   }
 }
+message CreateSnapshotRequest {
```
lichuqiang: source to create the snapshot (i.e. volume_id) missing here?

xing-yang: Oops:).  Good catch!  "volume_id" was in the draft design doc, but it looks like I removed it here by accident when removing "version".  Thanks!

xing-yang: Fixed.

csi.proto
```diff
@@ -172,6 +184,10 @@ message CreateVolumeRequest {
   // logged, etc.) by the CO.
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // The ID of the source snapshot to create volume from.
+  // This field is OPTIONAL.
+  string snapshot_id = 6;
```
gouthampacha: I suggest renaming this field as `source_snapshot_id` so that it is less unambiguous, list like the response field on L301. If we support more source types in the future, the `source_*` prefix will help clarify what is being requested as the source.

Alternatively, adding a struct `source_snapshot_request` would allow more flexibility, in case we need more that the ID as Saad suggested in the meeting today.

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+  // If not set (value of 0), it indicates this size is unknown.
+  // The value of this field MUST NOT be negative.
+  int64 restored_size = 1;
+
+  // The size of the snapshot in bytes. This indicates the given size of
+  // the snapshot when it was created. This may be equal to or greater
+  // than the restored_size. This field is OPTIONAL. If not set (value
+  // of 0), it indicates this size is unknown.
+  // The value of this field MUST NOT be negative.
+  int64 size = 2;
```
jieyu: Is there a way to let the CO know about that size before calling `CreateSnapshot`? In other words, is there a way for the CO to tell that there's not enough space for creating a snapshot for a given volume, so that the CO will avoid calling `CreateSnapshot`.

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+
+##### CreateSnapshot Errors
+
+If the plugin is unable to complete the CreateSnapshot call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Snapshot already exists but is incompatible | 6 ALREADY_EXISTS | Indicates that a snapshot corresponding to the specified snapshot `name` already exists but is incompatible with the specified `volume_id`. | Caller MUST fix the arguments or use a different `name` before retrying. |
```
jieyu: Is there a specific error code for "not enough space for a snapshot"

jdef: grpc resource-exhausted ?

spec.md
saad-ali: This may be 0 until "upload" is complete. So this number may change over time. Not immutable, some storage systems can't estimate it.

User wants to know how much they are paying for, so it is important to surface to the end user.
May be useful for CO to do resource management. This will be complicated based on implementation. In some cases it is impossible to know so it can't be required.

CO may not be able to rely on this to calculate capacity.

May make sense as an opaque field.

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+  //    snapshot exists. The Plugin should ensure that multiple
+  //    `CreateSnapshot` calls for the same name do not result in more
+  //    than one snapshot created corresponding to that name. If a
+  //    Plugin is unable to enforce idempotency, the CO's error recovery
+  //    logic could result in multiple (unused) snapshots being created.
+  // 2) Suggested name - Some storage systems allow callers to specify
+  //    an identifier by which to refer to the newly created snapshot.
+  //    If a storage system supports this, it can optionally use this
+  string name = 2;
+
```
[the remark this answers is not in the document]
e4jet: A few more use cases for opaque ```parameters```:
* Create a Read-only or. Read/Write Snapshot
* Should the Snapshot be replicated some place?
* Should the Snapshot have an expiration?

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+  // subsequent calls to refer to the provisioned snapshot.
+  string id = 3;
+
+  // Description for the created snapshot. This field is OPTIONAL.
+  string description = 4;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string volume_id = 5;
```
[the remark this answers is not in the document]
e4jet: Just to expand on this request - The storage system is ultimately responsible for the creation of the snapshot.  There could be some delay between when the CO requests the snapshot and when the snapshot is actually taken.  In order to prevent inconsistency, a storage system driven "create time/date" would be useful.

csi.proto
```diff
@@ -502,6 +529,183 @@ message ControllerServiceCapability {
...
+  // encoding scheme, like base64. An SP SHALL advertise the
+  // requirements for credentials in documentation. COs SHALL permit
+  // users to pass through the required credentials. This information is
+  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_delete_credentials = 2;
+}
+
+message DeleteSnapshotResponse {}
+message ListSnapshotsRequest {
```
e4jet: The design in the google doc states:
>The Plugin SHALL return the information about all the snapshots that it knows about.

Does this mean all those snapshots created using this interface, or all snapshots regardless of how they were created?  Perhaps this should be stated in the comments as well?

spec.md
jdef: this value may change over time, depending upon how snapshots are implemented on the backend. I don't like having something variable coming back in the Snapshot object. Maybe the thing that this field is supposed to represent should come back in ListSnapshots?

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+`GetSnapshot` should be called right after `CreateSnapshot`.
+
+`GetSnapshot` is not required. However if the `REQUIRES_GET_SNAPSHOT` capability is supported, CO must call `GetSnapshot` after calling `CreateSnapshot`.
+
+From CSI plugin’s point of view, both calls are synchronous.
+All CSI functions have been designed to be synchronous, so the snapshot functions are designed to follow the convention.
+It is important to separate these two functions so that a post-action such as `thaw` can be performed on the application or filesystem right after `CreateSnapshot` is complete.
+Since the uploading process could take hours, we don’t want to wait for the snapshot to be uploaded before performing `thaw` because the application should be made available as soon as possible.
+
+#### `GetSnapshot`
```
jdef: I don't see a need for this (GetSnapshot) RPC. ListSnapshots provides the same (and more) functionality. If there's concern about iterating to pages of snapshot objects just to get details for a single snapshot, then perhaps amend the ListSnapshots RPC with some kind of filter-by-snapshot-id?

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+  // users to pass through the required credentials. This information is
+  // sensitive and MUST be treated as such (not logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_list_credentials = 4;
+}
+
+message ListSnapshotsResponse {
+  message Entry {
+    Snapshot snapshot = 1;
+  }
```
jdef: maybe a field of Entry could be "used_capacity_bytes" - which indicates the physical space **currently** consumed by the snapshot, with the understanding that this metric MAY be variable over time (replaces the `size` field of `Snapshot`). A CO could periodically poll `ListSnapshots` to calculate the space consumed by snapshots in the system.

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+
+message CreateSnapshotResponse {
+  // Contains all attributes of the newly created snapshot that are
+  // relevant to the CO along with information required by the Plugin
+  // to uniquely identify the snapshot. This field is REQUIRED.
+  Snapshot snapshot = 1;
+}
+
+// The information about a provisioned snapshot.
+message Snapshot {
```
jdef: I like the idea of `Snapshot` containing ONLY those fields that won't change over time.

spec.md
```diff
@@ -1090,6 +1119,287 @@ message ControllerServiceCapability {
...
+| Call not implemented | 12 UNIMPLEMENTED | CreateSnapshot call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` to discover Plugin capabilities. |
+
+##### CreateSnapshot vs GetSnapshot
+
+In addition to `CreateSnapshot`, a `GetSnapshot` function is also introduced.
+
+For some cloud providers, a snapshot will be uploaded to a location in the cloud (i.e., an object store) after the snapshot is cut (created).
```
jdef: If CreateSnapshotResponse included guidance for a CO (e.g. "I'm a fully created snapshot AND ready for use" vs. "I'm a create snapshot, but not ready for use yet") then this flow becomes simpler. Also, we should get rid of GetSnapshots for this flow and use ListSnapshots instead

jdef: Also, if CreateSnapshotResponse included such guidance, a CO could simply issue repeated (idempotent) calls to CreateSnapahotResponse and monitor the response until it indicates "created AND ready for use". No need for GetSnapshot or ListSnapshots in that case

csi.proto
```diff
@@ -172,6 +189,36 @@ message CreateVolumeRequest {
...
+
+  // Contains identity information for the created snapshot. This field
+  // is REQUIRED. The identity information will be used by the CO in
+  // subsequent calls to refer to the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string volume_id = 3;
```
govint: Is this more for validation that the volume is being created from a snapshot of a volume? Else why is it needed when creating a volume from a snapshot.

Creating a snap of a snap is not supported which is fine.

csi.proto
```diff
@@ -502,6 +556,213 @@ message ControllerServiceCapability {
...
+  // An SP SHALL advertise the requirements for required secret keys and
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
+  // logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_create_secrets = 4;
```
govint: Can these be updated when the plugin is running, say more key-value pairs are added? Or should the plugin be stopped and then restarted with an updated secret.

csi.proto
```diff
@@ -502,6 +556,213 @@ message ControllerServiceCapability {
...
+
+  // Plugin specific parameters passed in as opaque key-value pairs.
+  // This field is OPTIONAL. The Plugin is responsible for parsing and
+  // validating these parameters. COs will treat these as opaque.
+  // Use cases for opaque parameters:
+  // - Specify a policy to automatically clean up the snapshot.
+  // - Specify an expiration date for the snapshot.
+  // - Specify whether the snapshot is readonly or read/write.
+  // - Specify if the snapshot should be replicated to some place.
```
govint: Can this be instead policy based where we create a policy profile and attach that to the volume at create time or later (via an update API) that specifies a snapshot schedule prescribing the interval between snaps, the lifecyle of snaps, how many snaps to maintain at a time (or snapshot depth)., etc.

govint: Will the snapshot be just crash consistent or application consistent or not-consistent at all (which is btw useless).

The proposal must handle the mechanism to be able to take an app consistent snap (VSS?), for this feature to be useful.

csi.proto
govint: Or can the secret be associated with the volume vs. with a running instance of a vendor volume plugin. I'd suggest keeping the secret as meta-data for the volume just so it doesn't become vendor plugin side meta-data needing the vendor plugin to interpret the secret data and its format as it changes.

Keeping the secret attached to a volume and an API to get() KV pairs from the secret is better IMO

csi.proto
```diff
@@ -502,6 +556,213 @@ message ControllerServiceCapability {
...
+  // Uploading is a long process that could take hours. At this phase,
+  // since the snapshot is already cut, a `thaw` operation can be
+  // performed so application can start running again (assuming a
+  // `freeze` operation was done before taking the snapshot). The
+  // status of the snapshot will become `READY` after the upload is
+  // complete.
+  // For cloud providers and storage systems that don't have the
+  // uploading process, the status should be `READY` after `CREATING`.
+  // `thaw` can be done when the status is `READY` in this case.
+  SnapshotStatus status = 5;
```
govint: How does user track that the snapshot is in creating state or stuck in creating? Is there something like a query snapshot state so that the CO can figure that either the snap completed or is stuck and issues a cancel op

csi.proto
```diff
@@ -502,6 +556,213 @@ message ControllerServiceCapability {
...
+  // Each value MUST contain a valid string. An SP MAY choose to accept
+  // binary (non-string) data by using a binary-to-text encoding scheme,
+  // like base64.
+  // An SP SHALL advertise the requirements for required secret keys and
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
```
govint: both for the create and the delete it seems a little weird that a user is going to give their username and password (aka sensitive data) to a third party vendor plugin. What prevents a malicious plugin from passing the user-name and secret to other entities or caching the username/password at locations that aren't secure?

There is no relation between the volume user (end user thats owning the data and the app) and the volume plugin that's facilitating the allocation of storage.

Passing in secrets to the volume provider seems like a very insecure thing to do and an additional burden on the volume plugin to not misuse this data.

csi.proto
```diff
@@ -172,6 +189,36 @@ message CreateVolumeRequest {
...
+  SourceSnapshot source_snapshot = 6;
+}
+
+// The information about a source snapshot.
+message SourceSnapshot {
+  // SnapshotSize contains `used_capacity_byes` which is the physical
+  // space consumed by the snapshot in bytes. `used_capacity_byes` may
+  // be variable over time. This field is OPTIONAL.
+  // If SnapshotSize is not set, it indicates this size is unknown.
+  SnapshotSize size = 1;
```
cpuguy83: Why is this needed?

csi.proto
```diff
@@ -172,6 +189,36 @@ message CreateVolumeRequest {
   // logged, etc.) by the CO.
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // Contains all attributes of the source snapshot to create volume
+  // from. This field is OPTIONAL.
+  SourceSnapshot source_snapshot = 6;
```
cpuguy83: What do you think about using a `oneof` instead and calling this something like `ContentSource` where one of the possibilities is a snapshot reference?

csi.proto
```diff
@@ -37,6 +37,23 @@ service Controller {
...
+  // functionality not every storage system can support and therefore
+  // is out of scope for this proposal. It should be added as a separate
+  // function (not in CreateVolume) when proposed in the future.
+  rpc CreateSnapshot (CreateSnapshotRequest)
+    returns (CreateSnapshotResponse) {}
+
+  rpc DeleteSnapshot (DeleteSnapshotRequest)
+    returns (DeleteSnapshotResponse) {}
+
+  rpc ListSnapshots (ListSnapshotsRequest)
```
cpuguy83: Was there some feedback that a `ListSnapshots` rpc was needed? I'm not sure when an orchestrator would use this.

csi.proto
```diff
@@ -502,6 +556,213 @@ message ControllerServiceCapability {
...
+  // Use cases for opaque parameters:
+  // - Specify a policy to automatically clean up the snapshot.
+  // - Specify an expiration date for the snapshot.
+  // - Specify whether the snapshot is readonly or read/write.
+  // - Specify if the snapshot should be replicated to some place.
+  // - Specify primary or secondary for replication systems that
+  //   support snapshotting only on primary.
+  map<string, string> parameters = 5;
+
+  // NOTE: In the future, we may want to add “tags” to identify a
```
cpuguy83: I don't think we need the note in here? Sounds like this should be an issue on the repo.

csi.proto
cpuguy83: Just noting this essentially makes snapshot creation an async operation, which is a different API than the rest of the interface.

csi.proto
```diff
@@ -172,6 +189,36 @@ message CreateVolumeRequest {
...
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // Contains all attributes of the source snapshot to create volume
+  // from. This field is OPTIONAL.
+  SourceSnapshot source_snapshot = 6;
+}
+
+// The information about a source snapshot.
+message SourceSnapshot {
```
cpuguy83: suggestion: `SnapshotReference`

csi.proto
e4jet: A key workflow that this should support is creating a volume from a snapshot.  I believe that the idea here is that a CO would ask for the list of snapshots so that the user can pick one to clone from.

csi.proto
e4jet: I think its up to the individual CO to decide how these treated (ie: policy based).  While its fine to provide these at volume create time (ie: for schedules etc.), I believe they should be available at snapshot create time as well.  This enables the user to create a "one off" snapshot with a different set of parameters.

xing-yang:
Added GetSnapshot back in the latest PR. CreateSnapshot is synchronous and must block until the snapshot is cut. GetSnapshot is also synchronous and must block until the snapshot is uploaded. ListSnapshots does not block for a particular snapshot as it just returns snapshots with the current information.  So I think a GetSnapshot is still needed and serves a different purpose from CreateSnapshot.

xing-yang:
Updated the section on SnapshotSize.

From this doc (https://cloud.google.com/compute/docs/reference/rest/v1/disks/createSnapshot), the output of createSnapshot has two sizes:
snapshot.diskSizeGb - "Size of the snapshot in GB."
snapshot.storageBytes - "A size of the storage used by the snapshot. As snapshots share storage, this number is expected to change with snapshot creation/deletion."

From here (https://cloud.google.com/compute/docs/reference/rest/v1/disks), there is a sizeGb in create volume:
sizeGb - "Size of the persistent disk, specified in GB. You can specify this field when creating a persistent disk using the sourceSnapshot parameter. The value of sizeGb must not be less than the size of the snapshot."

Based on above, we need to specify snapshot.diskSizeGb (not snapshot.storageBytes) when creating a volume from the snapshot.  Therefore I kept the “size” field in SnapshotSize.

iliastsi:
Hi @xing-yang,

I can see from this proposal that the `CreateSnapshot` call is implemented by the Service Controller. However, it seems that limiting the functionality of snapshots on the Service Controller and not allowing it to reach the Node Plugin, goes against PR #188 by @saad-ali. To put it differently, there may be volumes that are not equally available throughout the cluster the CO is managing, so the Node Plugin should have a chance to participate in `CreateSnapshot`. Maybe not every SP will need to implement `CreateSnapshot` functionality at the node level, but it is certainly beneficial for -- but not limited to -- local volumes.

iliastsi:
We are also interested in group consistent snapshots. Is this something that the SP should provide (e.g., `CreateSnapshot` should be updated to accept more than one source volumes, and snapshot them as a group) or is this something that will be implemented on the CO side (i.e., the CO will freeze our group of pods, then issue individual `CreateSnapshot` requests, then thaw the group)?

jdef:
I like the suggestion from @iliastsi. Perhaps the snapshot RPCs should be wrapped in a different gRPC service `Snapshot`. In addition the `Snapshot` service could provide `SnapshotGetCapabilities` RPC. 

One problem with this approach is that the CO has no way to discover whether the snapshot service is accessible from controller deployments, node deployments, or both. To resolve this we could add service co-location metadata to the PluginCapability message. This isn't quite as clean as I'd like, maybe someone has a better idea.

```protobuf
service Snapshot {
  rpc SnapshotCreate (SnapshotCreateRequest) returns (SnapshotCreateResponse) {}
  rpc SnapshotDelete (SnapshotDeleteRequest) returns (SnapshotDeleteResponse) {}
  rpc SnapshotList (SnapshotListRequest) returns (SnapshotListResponse) {}
  rpc SnapshotGetCapabilities(SnapshotGetCapabilitiesRequest) returns (SnapshotGetCapabilitiesResponse) {}
}
...
message PluginCapability {
  message Service {
    enum Type {
      UNKNOWN = 0;
      CONTROLLER_SERVICE = 1; // ... already part of the spec
      SNAPSHOT_SERVICE = 2;   // ... new: Snapshot service support
    }
    Type type = 1;               // ... already part of the spec
    repeated Type colocated = 2; // new: co-located with the service `type` above
  }

  oneof type {
    Service service = 1;
  }
}
```

csi.proto
```diff
@@ -37,6 +37,40 @@ service Controller {
...
+  // For this reason, a GetSnapshot RPC is introduced.
+  // CreateSnapshot is a synchronous call and it must block until
+  // the snapshot is cut. GetSnapshot is also synchronous and it
+  // must block until the upload process is complete.
+  // Status of the snapshot will be returned by CreateSnapshot and
+  // GetSnapshot to provide guidance for CO to make decisions.
+  // GetSnapshot must be called after the call to CreateSnapshot
+  // if the controller plugin has the REQUIRES_GET_SNAPSHOT capability.
+  rpc GetSnapshot (GetSnapshotRequest)
+    returns (GetSnapshotResponse) {}
```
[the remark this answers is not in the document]
xing-yang: CO can use ListSnapshots with snapshot_id as the filter to get status of a snapshot.  ListSnapshots should not block and it should return with the current info about the snapshot(s).

xing-yang:
@iliastsi, I’m also very interested in consistent group snapshot, but I think that should be handled by a different proposal due to the complexity it will bring to the architecture.

In K8S, we can probably use StatefulSet which allows us to handle multiple volumes together.  We can do freeze before taking snapshots of volumes in the StatefulSet, take snapshots, and unfreeze after snapshots are cut for all the volumes in the StatefulSet.  This approach, however, will not allow us to use the consistent group snapshot feature provided by the Storage Systems directly.  To be able to take a consistent group snapshot using the functionality provided by the Storage Systems directly, we can introduce a volume group construct first and then introduce consistent group snapshot on top of that.  If the CSI community is interested in this feature, we can have a discussion about it later.

jdef:
An alternative to the messy "colocated service" proposal would be to explicitly support controller- and node-level snapshotting via different services. This cleanly separates the capabilities, APIs, and is easier for a CO to understand.

```protobuf
service ControllerSnapshotter {
  rpc ControllerSnapshotCreate(...) returns (...) {}
  rpc ControllerSnapshotDelete(...) returns (...) {}
  rpc ControllerSnapshotList(...) returns (...) {}
  rpc ControllerSnapshotGetCapabilities(...) returns (...) {}
}

service NodeSnapshotter {
  rpc NodeSnapshotCreate(...) returns (...) {}
  rpc NodeSnapshotDelete(...) returns (...) {}
  rpc NodeSnapshotList(...) returns (...) {}
  rpc NodeSnapshotGetCapabilities(...) returns (...) {}
}

message PluginCapability {
  message Service {
    enum Type {
      UNKNOWN = 0;
      CONTROLLER_SERVICE = 1; // ... already part of the spec

      // CONTROLLER_SNAPSHOTTER_SERVICE and NODE_SNAPSHOTTER_SERVICE
      // are mutually exclusive: a plugin SHALL NOT advertise both services.

      CONTROLLER_SNAPSHOTTER_SERVICE = 2; // new: OPTIONAL, controller-level snapshotting
      NODE_SNAPSHOTTER_SERVICE = 3; // new: OPTIONAL, node-level snapshotting
    }
...
```

xing-yang:
Hi @jdef and @iliastsi , very interesting idea regarding making snapshot both a controller and node service.  I’m reading PR #188.  In that PR about topology, TopologyRequirement is added to CreateVolumeRequest, but Create Volume is still only implemented as a Controller RPC.

Once a volume is created and topology requirement is fulfilled, I don’t think the snapshot needs additional topology requirement.  For cloud providers and storage systems that don’t need to upload the snapshot after the snapshot is cut, the snapshot will be residing on the same storage backend where the volume is provisioned.  For cloud providers that will upload the snapshot after it is cut, my understanding is that the snapshot will become available anywhere in the cloud after the upload is complete so this also doesn’t need additional topology requirement.

Let’s discuss more about this at the meeting.  Thanks!

jdef:
@xing-yang thanks for the feedback. I've thought about things a bit more and I think I've added some confusion to the mix w/ my proposals. I started thinking through concrete use cases of local snapshotting and couldn't actually come up with a reason to differentiate between controller + node snapshots. So feel free to ignore both my colocation proposal and the split-service proposal.

xing-yang:
@jdef Sounds good.  Thanks!

iliastsi:
Apologies for missing today's meeting. I was caught up in a last minute call. In fact, I tried joining the meeting at 9:15 PDT, but it had already finished. Here are the key points I was going to discuss on the meeting.

I believe the `CreateSnapshot` call should be able to reach the node where a volume is currently bound, in order for the Node Plugin to perform additional needed steps, which cannot be made remotely. It is important to allow SPs the possibility to perform actions at the node level (especially for volumes which are in use) even if these actions are going to be NOOP for some of them.

As an example, think of a CSI plugin which utilizes caches on the node where a volume is bound. The Node Plugin should be informed about the Snapshot operation in order to flush the caches prior to taking the snapshot.

Looking forward to your comments.

spec.md
```diff
@@ -1090,6 +1180,342 @@ message ControllerServiceCapability {
...
+  // An SP SHALL advertise the requirements for required secret keys and
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
+  // logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_create_secrets = 4;
```
jieyu: can we name this to `create_snapshot_secrets`?

too bad that we have `controller_create_secrets` in `CreateVolumeRequest` (should have been `create_volume_secrets`).

spec.md
jieyu: Does that mean `status` shouldn't be in `Snapshot`? (re: `Snapshot` containing ONLY those fields that won't change over time).

jieyu: If we were to separate immutable fields from mutable fields, maybe  we should rename `GetSnapshot` to `GetSnapshotStatus`?

spec.md
```diff
@@ -1090,6 +1180,342 @@ message ControllerServiceCapability {
...
+     // A snapshot is cut and is now being uploaded.
+     UPLOADING = 1;
+     // A snapshot is ready for use.
+     READY = 2;
+  }
+  Type type = 1;
+}
+```
+
+In the future, we may want to add `tags` (map<string, string> tags) in `CreateSnapshotRequest` to identify a snapshot.
```
jieyu: I'd remove this paragraph given it's not there.

spec.md
```diff
@@ -1090,6 +1180,342 @@ message ControllerServiceCapability {
...
+  // An SP SHALL advertise the requirements for required secret keys and
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
+  // logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_delete_secrets = 2;
```
jieyu: This should be `delete_snapshot_secrets`

spec.md
```diff
@@ -1090,6 +1180,342 @@ message ControllerServiceCapability {
...
+  // An SP SHALL advertise the requirements for required secret keys and
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
+  // logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> controller_get_secrets = 2;
```
jieyu: This should be `get_snapshot_secrets`

xing-yang:
@iliastsi Thanks.  I think that should not be handled by the out-of-tree CSI plugin which is defined by this CSI spec.  In this google doc (https://docs.google.com/document/d/1oXVuDTNhXerr1UJ48tQrxF9J6vbm2LHVVC3z28655Qo/edit#) where we drafted this snapshot proposal, we mentioned about having a SnapshottableVolumePlugin in K8S to facilitate the freeze action before taking a snapshot when the volume is attached and mounted on a pod.

csi.proto
```diff
@@ -172,6 +206,50 @@ message CreateVolumeRequest {
...
+// a supported ContentSource at the moment.
+message ContentSource {
+  oneof source_type {
+    SnapshotReference snapshot_ref = 1;
+  }
+}
+
+// The information about a snapshot reference as a source to create
+// a volume from.
+message SnapshotReference {
```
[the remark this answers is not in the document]
xing-yang: Ok, I'll create a nested type to avoid confusion.

So CO wants to create a volumeB from this SnapshotReference.  CO needs to know how much space is needed to create volumeB from this snapshot so "size" is needed.  Regarding "volume_id", that is the original volume where this snapshot is created from.

spec.md
```diff
@@ -49,6 +49,8 @@ The Container Storage Interface (CSI) will
 * Define plugin protocol RECOMMENDATIONS.
   * Describe a process by which a Supervisor configures a Plugin.
   * Container deployment considerations (`CAP_SYS_ADMIN`, mount namespace, etc.).
+  * Creating and deleting a snapshot (source of the snapshot is a volume).
```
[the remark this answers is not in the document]
xing-yang: Good point.  Will move it.

spec.md
```diff
@@ -49,6 +49,8 @@ The Container Storage Interface (CSI) will
...
+  * Provisioning a new volume from a snapshot (reverting snapshot, where data in the original volume is erased and replaced with data in the snapshot, is out of scope).
```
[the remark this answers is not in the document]
xing-yang: Will do.

spec.md
```diff
@@ -317,6 +319,40 @@ service Controller {
 
   rpc ControllerGetCapabilities (ControllerGetCapabilitiesRequest)
     returns (ControllerGetCapabilitiesResponse) {}  
+
+  // Snapshot can be used as the source to provision a new volume.
```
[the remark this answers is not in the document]
xing-yang: I think I got comments in the previous meeting to add comments here so that these comments will be shown in the generated csi.proto:).  I can certainly move them to the relevant section below.

spec.md
```diff
@@ -592,6 +628,50 @@ message CreateVolumeRequest {
...
+  // Contains all attributes of the source that is used to create
+  // the volume from. This field is OPTIONAL.
+  ContentSource content_source = 6;
+}
+
+// Specifies what source the volume will be created from. One of the
+// source_type fields MUST be specified. Only SnapshotReference is
+// a supported ContentSource at the moment.
+message ContentSource {
+  oneof source_type {
```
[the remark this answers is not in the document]
xing-yang: Sure.

spec.md
```diff
@@ -592,6 +628,50 @@ message CreateVolumeRequest {
...
+    SnapshotReference snapshot_ref = 1;
```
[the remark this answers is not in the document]
xing-yang: Sure.

spec.md
```diff
@@ -592,6 +628,50 @@ message CreateVolumeRequest {
...
+  }
+}
+
+// The information about a snapshot reference as a source to create
+// a volume from.
+message SnapshotReference {
+  // SnapshotSize is the disk size of the snapshot in bytes. The purpose
```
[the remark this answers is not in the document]
xing-yang: Sure.

spec.md
```diff
@@ -1090,6 +1180,342 @@ message ControllerServiceCapability {
...
+  //    Plugin is unable to enforce idempotency, the CO's error recovery
+  //    logic could result in multiple (unused) snapshots being created.
+  // 2) Suggested name - Some storage systems allow callers to specify
+  //    an identifier by which to refer to the newly created snapshot.
+  //    If a storage system supports this, it can optionally use this
+  string name = 2;
+
+  // The description of the snapshot.
+  // This field is OPTIONAL.
+  string description = 3;
```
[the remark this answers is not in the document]
xing-yang: Sure, I can remove this.

spec.md
xing-yang: Yes.

spec.md
xing-yang: Ok.

spec.md
xing-yang: Sure.

spec.md
xing-yang: Ok.

spec.md
xing-yang: How about we remove Status from the Snapshot response from CreateSnapshot and GetSnapshot and keep the Status only in the response from ListSnapshots?

1. CO calls CreateSnapshot. CreateSnapshot will block until snapshot is cut.
2. CO checks if REQUIRES_GET_SNAPSHOT is a capability supported by the plugin.  If yes, CO will call GetSnapshot.
3a. CO calls GetSnapshot. GetSnapshot will block until snapshot is uploaded.  When GetSnapshot returns, CO is done with snapshot creation and uploading.
3b. If REQUIRES_GET_SNAPSHOT is not supported, CO is done with snapshot creation without calling GetSnapshot.

CO can always call ListSnapshots with snapshot_id as the filter to find out the Status of a particular snapshot.   ListSnapshots does not block and will return with the current snapshot info.

jdef:
I still don't think that we need GetSnapshot. We don't have a GetVolume.
I'd rather have CreateShapshot return some kind of Status object in the
response. Keep ListSnapshots for a CO that's interested in querying the
status of 1 or more snapshots. Having List and Get is redundant.

iliastsi:
@xing-yang, I understand that the CO will be responsible for freezing/thawing the application before/after taking a snapshot when the volume is attached and mounted, but that is not enough given my previous example. Let me re-iterate.

Let's say I have implemented a CSI plugin for Google Compute Engine Persistent Disks, but when the volume is mounted on the node, our plugin uses dm-cache over the persistent disk. In that case, merely freezing the application before taking a snapshot is not enough, since the cache device should be flashed first. This is an action specific to the above CSI plugin that only the Node Service could perform.

Since the CSI plugin (and not Kubernetes or any other CO for that matter) is responsible for binding the block volume on the host, we cannot rule out the possibility for vendor specific actions that may need to take place on the node before and/or after the snapshot is cut.

The above is not a theoretical example. This is a problem we are facing with our own CSI plugin, where we cache part of the volume locally at the block level. By not allowing the Snapshot functionality to reach the Node Plugin, our SP cannot support snapshots.

CSI should support this kind of flexibility, so it can be used by Storage Providers that wish to do advanced storage-specific actions at the node level.

xing-yang:
@jdef @jieyu Summarizing the comments regarding GetSnapshot, there are a couple of options:

1. Keep GetSnapshot, and keep Status in CreateSnapshot, GetSnapshot, and ListSnapshots

a) CO calls CreateSnapshot. CreateSnapshot will block until snapshot is cut.  CreateSnapshot returns Status of either “ready” or “uploading” depending on whether the snapshot needs to be uploaded after it is cut.
b) CO checks if REQUIRES_GET_SNAPSHOT is a capability supported by the plugin. If yes, CO will call GetSnapshot.
c1) CO calls GetSnapshot. GetSnapshot will block until snapshot is uploaded. When GetSnapshot returns, CO is done with snapshot creation and uploading.  GetSnapshot returns a Status of “ready” when complete.
c2) If REQUIRES_GET_SNAPSHOT is not supported, CO is done with snapshot creation without calling GetSnapshot.
d) CO can always call ListSnapshots (if LIST_SNAPSHOTS is supported) with snapshot_id as the filter to find out the Status of a particular snapshot. ListSnapshots does not block and will return with the current snapshot info.

2. Keep GetSnapshot, remove Status from CreateSnapshot, GetSnapshot, and keep status in ListSnapshots

a) CO calls CreateSnapshot. CreateSnapshot will block until snapshot is cut.  CreateSnapshot does not return status.
b) CO checks if REQUIRES_GET_SNAPSHOT is a capability supported by the plugin. If yes, CO will call GetSnapshot.
c1) CO calls GetSnapshot. GetSnapshot will block until snapshot is uploaded. When GetSnapshot returns, CO is done with snapshot creation and uploading.
c2) If REQUIRES_GET_SNAPSHOT is not supported, CO is done with snapshot creation without calling GetSnapshot.
d) CO can always call ListSnapshots (if LIST_SNAPSHOTS is supported) with snapshot_id as the filter to find out the Status of a particular snapshot. ListSnapshots does not block and will return with the current snapshot info.

3. Remove GetSnapshot, and keep Status in CreateSnapshot and ListSnapshots.

a) CO calls CreateSnapshot. CreateSnapshot will block until snapshot is cut.  CreateSnapshot returns a status of either “ready” or “uploading” depending on whether the snapshot needs to be uploaded after it is cut.
b) If the status is “ready”, snapshot creation is complete.
c) If the status is “uploading”, the snapshot is being uploaded after it is cut.  For storage systems that need to upload the snapshot, LIST_SNAPSHOTS must be supported. CO will call ListSnapshots with snapshot_id as the filter to find out the Status of a particular snapshot. ListSnapshots does not block and will return with the current snapshot info.  CO checks the status from ListSnapshots to determine whether the upload is complete.  The upload is complete when the status becomes “ready”.

Can we choose one of these options?

xing-yang:
@iliastsi I see.  In your CSI plugin, have you implemented the controller RPCs such as create/delete volume?  For example the NFS plugin has not implemented create/delete volume.  Just wondering if your plugin is similar to that.  For your CSI plugin, is the node RPC enough for creating snapshot or do you need both controller and node RPCs.  Do you need 1 or 2 in the following?
1) CreateSnapshot in Controller service and PreSnapshotAction/PostSnapshotAction in Node service
Or
2) CreateSnapshot and PreSnapshotAction/PostSnapshotAction all in Node service

iliastsi:
@xing-yang, since our storage backend supports dynamic volume provisioning, we have implemented all of the Controller RPCs.

I have been thinking more about this and more specifically how we are going to allow COs to freeze running applications before taking a snapshot. 

As you have also mentioned, the draft of the Snapshot proposal in Google docs talks about a SnapshottableVolumePlugin in K8S to facilitate the freeze action before taking a snapshot, when the volume is attached and mounted on a pod. Here is how I imagine the workflow in this case (please correct me wherever I am wrong):

  1. A user creates a VolumeSnapshot from a PersistentVolumeClaim
  2. An in-tree CSI plugin (SnapshottableVolumePlugin) finds the volume attached/mounted and freezes the running application. It annotates the VolumeSnapshot to indicate the application is now frozen. This plugin runs on the kubelet.
  3. The external snapshot controller for the CSI driver (CSI snapshotter) sees the annotated VolumeSnapshot object so it starts creating the snapshot.
  4. Once the snapshot has been created, the snapshot controller annotates the VolumeSnapshot object.
  5. The SnapshottableVolumePlugin notices the above annotation and thaws the application.

Given the pull/watch pattern on which K8S has been designed, the delay between between steps 2 and 3, and steps 4 and 5 may be several seconds (or even minutes). If this is the case, this will make the freeze times completely unsuitable from production workloads.

My point is that the desired workflow should minimize the delay introduced between freezing the application and calling `CreateSnapshot` in the external CSI plugin to take the actual snapshot. I think this calls the feasibility of having an external Snapshot Controller into question.

To summarize, since the kubelet is responsible for freezing the pod, the only viable option for keeping the application frozen for as short as possible is for the kubelet itself to immediately request the snapshot through an RPC to the CSI Node Plugin.

I do not have all the details available, but it seems that having a `PrepareSnapshot` call in the Controller Service, and then the actual `CreateSnapshot` call in the Node Service is the way to go. This would be similar to your proposal number 2.

xing-yang:
@iliastsi Do you have a website that tells us more about your storage?  Is your CSI plugin posted somewhere on GitHub?  I'd like to understand more about your storage.  Thanks.

iliastsi:
@xing-yang All content that is currently publicly available is on our website, www.arrikto.com. We have not posted our CSI plugin on GitHub yet. We have worked extensively on supporting snapshots and clones for Persistent Volumes on Kubernetes, so my questions arise mostly from our experience and the workarounds we had to do, to make it work end-to-end.

However, because CSI should not be influenced by the needs of a specific plugin, and it shouldn't matter what our specific plugin requires, we don't want to tie the implantation of CSI to our own needs. That's why in my previous message I have tried to make a generic argument, as much as possible, and not focus on our own needs.

It seems to me that if the CreateSnapshot RPC does not reach the CSI Node Plugin, there is no obvious way to not incur very long latencies from the time the kubelet freezes the application, then the external plugin does the snapshot, then the CO thaws it. This is a showstopper for being able to snapshot production workloads.

spec.md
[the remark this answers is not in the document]
xing-yang: Moved comments down to the relevant section below.

spec.md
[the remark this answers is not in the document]
xing-yang: Done.

spec.md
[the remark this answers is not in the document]
xing-yang: Since I renamed SnapshotReference to Snapshot and defined it as a nested type inside VolumeContentSource, I can just call it snapshot instead of snapshot_reference?

spec.md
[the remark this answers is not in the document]
xing-yang: Removed.

spec.md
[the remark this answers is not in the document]
xing-yang: Removed "description" field.

spec.md
xing-yang: Done.

spec.md
xing-yang: Removed.

spec.md
xing-yang: Done.

spec.md
xing-yang: Done.

xing-yang:
@iliastsi Based on your information, I think we can have a CreateSnapshot RPC in both Controller and Node services.  So this brings us back to the proposals from @jdef.  I'd suggest that we keep it as simple as possible and not to add another controller/node service.  We can add CREATE_DELETE_SNAPSHOT capability in both ControllerServiceCapability and NodeServiceCapability and a plugin should support only one of them.  In K8S, we can add a flag to let kubelet know whether snapshot functions will be invoked by kubelet or a snapshot controller.  I think we should not add PreSnapshotAction/PostSnapshotAction as CSI driver interfaces as they should be handled by orchestration. If you implement CreateSnapshot in the node service, you can do flush cache there.

iliastsi:
@xing-yang Having the ability to implement CreateSnapshot in the Node Service will work for us. Let's move forward with having a CreateSnapshot RPC in Node Services as well.

I wonder whether having a CreateSnapshot RPC both in Controller and Node services and letting SPs choose which one to implement is going to put the burden on the COs to support both of them, but maybe CO vendors have more to comment on this.

xing-yang:
@iliastsi For most Storage Providers, the CreateSnapshot RPC should be implemented in the Controller Service, together with Create/DeleteVolume and ControllerPublish/UnpublishVolume.  In the case you mentioned where Node is part of your storage, it makes sense to implement CreateSnapshot as a Node RPC.  I don’t think we should require all SPs to implement CreateSnapshot in the Node service.

iliastsi:
@xing-yang Given my previous [comment](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375695027), I believe most Storage Providers will implement the CreateSnapshot RPC in the Node Service, because that is the only way to support freezing/thawing running applications.

Let's talk more about this during the next meeting. Thanks!

cpuguy83:
@iliastsi I don't think it would be the the plugin's responsibility to freeze/thaw applications.

tsmetana:
Hello. We did have this conversation before. There is no single correct way to freeze/thaw an application. Every app is different and the CO knows nothing about it. If the call to CreateSnapshot implies also the appliction using the volume is correctly paused, there has to be a way for the user to tell the CO how to do this and it has to be also standardized.

xing-yang:
Agree with @cpuguy83 and @tsmetana. Freeze/thaw is part of the orchestration and should not be part of the storage plugin interfaces.  We should separate that from the storage plugin interfaces.  Every app is different and needs a different way to freeze/thaw.  For a complete solution, we’ll need an agent for each application to quiesce that particular application properly.  Each agent tool is different because different applications such as Oracle or SQL provide different APIs to quiesce the apps.  I think CreateSnapshot in the storage plugin should just perform the storage specific task and create a point-in-time copy of the source volume.  In the CO, we should provide hooks for freeze/thaw to be performed in the orchestration work flow.

iliastsi:
@xing-yang, @cpuguy83, @tsmetana I agree. As @xing-yang said, freeze/thaw is part of the orchestration and should not be part of the storage plugin interfaces. In the CO, we should provide hooks for freeze/thaw to be performed in the orchestration workflow.

However I don't think this is part of our current discussion. **Given** that the CO will have its own user-specified way freezing/thawing the application, it is imperative to have the CSI CreateSnapshot RPC at the node level.

Please see my previous [comment](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375695027) for a detailed description of why the CreateSnapshot call cannot be an RPC made by an external snapshot controller towards the CSI Controller Service but rather, it should be a call from the CO to the Node Service Plugin.

spec.md
```diff
@@ -1090,6 +1156,346 @@ message ControllerServiceCapability {
...
+This RPC will be called by the CO to get a snapshot.
+GetSnapshot function is synchronous and it should block until the upload is complete and the snapshot is ready.
+GetSnapshot should be called right after CreateSnapshot.
+
+Some cloud providers will upload the snapshot to a location in the cloud (i.e., an object store) after the snapshot is cut.
+Uploading is a long process that could take hours.
+At this phase, since the snapshot is already cut, a `thaw` operation can be performed so application can start running again (assuming a `freeze` operation was done before taking the snapshot).
+For this reason, a GetSnapshot RPC is introduced.
+CreateSnapshot is a synchronous call and it must block until the snapshot is cut.
+GetSnapshot is also synchronous and it must block until the upload process is complete.
```
yasker: It's still not clear to me why `GetSnapshot` should be blocked until the upload is completed. Does that mean `GetSnapshot` will never know if a snapshot is `Uploading` or not? Even `Uploading` is a valid state in the snapshot status.

If instead GetSnapshot only reports the current status of the snapshot (potentially during uploading), the caller can simply choose to use GetSnapshot to poll the status of the snapshot, probably we can introduce a percentage of progress in the status as well. Blocking for potentially hours is more error-prone IMO.

If so, we may need a Error state in the snapshot object then.

yasker: OK, just saw the ListSnapshots call with snapshot ID, guess that answer the question.

Though in that case, without an Error state in the snapshot object, it seems hard for the caller to know if one snapshot goes wrong. Or the caller always need to depend on `GetSnapshot` to know if a snapshot goes wrong?  But `GetSnapshot` is optional by itself.

spec.md
yasker: I think the different behaviors of `GetSnapshot` and `ListSnapshots` are quite unexpected. Normally GetSnapshot is the singular form of ListSnapshots. I don't expect it to function completely different (blocking vs non-blocking, reporting the error etc).

How about we rename `GetSnapshot` to e.g. `WaitForSnapshotComplete`?

xing-yang: Agree that the naming is confusing.  There's still debate on whether GetSnapshot is needed.  Once that decision is made, we can decide whether renaming is needed.

spec.md
xing-yang: GetSnapshot is only required if the storage system needs to do an upload after the snapshot is cut.  That's why it is optional.  Not every storage system do an upload after the snapshot is cut.

Error status is not there because the original thought was that there will be an gRPC error code returned when an error occurs, but I agree that it is better to add the error status there.

iliastsi:
@xing-yang I cannot speak for the rest of the SPs. Let's go with a CreateSnapshot RPC in both Controller and Node services as you already suggested.

cpuguy83:
We can always add a node service later, burner can't take it away.
I would say let's opt for simple here and expand later as needed.

iliastsi:
@cpuguy83  We have already established that CreateSnapshot call should be able to reach the Node Plugin for Storage Providers to perform advanced storage-specific actions prior to taking a snapshot (see comments above for an example of using caches at the block level).

We need this **today**, and a simple approach just won't cut it.

cpuguy83:
@iliastsi Can you link to the specific comment(s)? There's a lot of discussion here.

iliastsi:
@cpuguy83 Of course, please see [here](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375376740), [here](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375422289), [here](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375695027), [here](https://github.com/container-storage-interface/spec/pull/207#issuecomment-376012713) and [here](https://github.com/container-storage-interface/spec/pull/207#issuecomment-377224279).

spec.md
xing-yang: Added an error status.

csi.proto
```diff
@@ -172,6 +184,48 @@ message CreateVolumeRequest {
...
+// a supported VolumeContentSource at the moment.
+message VolumeContentSource {
+  message Snapshot {
+    // SnapshotSize is the disk size of the snapshot. The purpose of
+    // this field is to give CO guidance on how much space is needed
+    // to create a volume from this snapshot. The size of the volume
+    // must not be less than the size of the source snapshot. This
+    // field is OPTIONAL. If SnapshotSize is not set, it indicates
+    // that this size is unknown.
+    SnapshotSize size = 1;
```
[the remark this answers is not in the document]
xing-yang: That's a good point. There's a CapacityRange in CreateVolumeRequest already.  CO can call ListSnapshots with snapshot_id to find out the size of the snapshot first, and then use that as a guidance to determine CapacityRange.  SnapshotSize can be removed from CreateVolumeRequest.content_source.

csi.proto
```diff
@@ -502,6 +566,244 @@ message ControllerServiceCapability {
...
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. This field is REQUIRED.
+  string timestamp = 4;
+
```
Akrog: Shouldn't we have snapshot attributes like we have the volume ones?

csi.proto
```diff
@@ -502,6 +566,244 @@ message ControllerServiceCapability {
...
+// be called to wait for the upload to complete. GetSnapshot is also
+// synchronous and it will block until the upload process is complete.
+// The status returned by GetSnapshot should be `READY`.
+// An `ERROR` status is not included because there should be an gRPC
+// error code if an error happens.
+message SnapshotStatus {
+  enum Type {
+     UNKNOWN = 0;
+     // A snapshot is cut and is now being uploaded.
+     UPLOADING = 1;
```
Akrog: Is uploading the right word?  Wouldn't it be better to use CREATING?

csi.proto
```diff
@@ -502,6 +566,244 @@ message ControllerServiceCapability {
...
+  // A token to specify where to start paginating. Set this field to
+  // `next_token` returned by a previous `ListSnapshots` call to get the
+  // next page of entries. This field is OPTIONAL.
+  // An empty string is equal to an unspecified field value.
+  string starting_token = 2;
+
+  // Identity information for the source volume. This field is OPTIONAL.
+  // It can be used to list snapshots by volume.
+  string source_volume_id = 3;
+
```
Akrog: Missing volume attributes

csi.proto
```diff
@@ -502,6 +566,244 @@ message ControllerServiceCapability {
...
+     // A snapshot is in error status.
+     ERROR = 3;
+  }
+  Type type = 1;
+}
+message GetSnapshotRequest {
+  // The ID of the snapshot to be retrieved.
+  // This field is REQUIRED.
+  string snapshot_id = 1;
+
```
Akrog: If we add snapshot attributes we would have to pass them here as well.

csi.proto
xing-yang: We discussed about it but decided to start without attributes and add later if needed.

csi.proto
xing-yang: This is actually "uploading".  At this point, the snapshot is already cut, the cloud provider is uploading the snapshot somewhere in the cloud.

There was a "creating" status initially in the google doc where we drafted this, but we decided to remove "creating" because CreateSnapshot is a blocking call and when it returns it is in either ready, uploading, or error status.

csi.proto
```diff
@@ -492,6 +550,12 @@ message ControllerServiceCapability {
       PUBLISH_UNPUBLISH_VOLUME = 2;
       LIST_VOLUMES = 3;
       GET_CAPACITY = 4;
+      CREATE_DELETE_SNAPSHOT = 5;
+      // CO must call GetSnapshot after CreateSnapshot to verify
+      // upload is complete if REQUIRES_GET_SNAPSHOT is true.
+      REQUIRES_GET_SNAPSHOT = 6;
+      LIST_SNAPSHOTS = 7;
+      CREATE_VOLUME_FROM_SNAPSHOT = 8;
```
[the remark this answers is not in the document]
xing-yang: Will do.

csi.proto
```diff
@@ -172,6 +184,48 @@ message CreateVolumeRequest {
...
+
+    // Contains identity information for the created snapshot. This
+    // field is REQUIRED. The identity information will be used by
+    // the CO in subsequent calls to refer to the provisioned snapshot.
+    string id = 2;
+
+    // Identity information for the source volume. Note that create
+    // snapshot from snapshot is not supported here so the source has
+    // to be a volume. This field is REQUIRED.
+    string volume_id = 3;
```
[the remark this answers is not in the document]
xing-yang: Need to discuss about this.

csi.proto
Akrog: I would need them now... Like I need some additional data in other RPC calls ( https://github.com/container-storage-interface/spec/issues/216 )

csi.proto
Akrog: I see, I didn't get what it was meant to do until you explained it in the meeting, thanks.

lpabon:
Just a link to the [three options](https://github.com/container-storage-interface/spec/pull/207#issuecomment-375421743)

j-griffith:
What if you led with option #3 
    "Remove GetSnapshot, and keep Status in CreateSnapshot and ListSnapshots."

It seems that you could overload the status field in the Snapshot response to include *just enough* detail here.  Something like:
    * Creating
    * Created
    * Uploading
    * Complete
An "Error" prefix in front of each of these would indicate where and if something went wrong.  So if the upload failed you'd have:
    "Failed-Uploading"

If we derive an agreed upon set of status listings here it's fairly well understood where something may have failed.

csi.proto
```diff
@@ -502,6 +549,218 @@ message ControllerServiceCapability {
...
+  //    `CreateSnapshot` again, with the same name, to ensure the
+  //    snapshot exists. The Plugin should ensure that multiple
+  //    `CreateSnapshot` calls for the same name do not result in more
+  //    than one snapshot created corresponding to that name. If a
+  //    Plugin is unable to enforce idempotency, the CO's error recovery
+  //    logic could result in multiple (unused) snapshots being created.
+  // 2) Suggested name - Some storage systems allow callers to specify
+  //    an identifier by which to refer to the newly created snapshot.
+  //    If a storage system supports this, it can optionally use this
+  string name = 2;
```
julian-hj: I feel like the above comment is too verbose and could be shortened to say "The suggested name for the snapshot. This field is REQUIRED for idempotency."

I realize that this comment is just a copy of the long comment in CreateVolume, but I feel that one is also too long, and we definitely don't need a second copy of the definition of idempotency plus commentary on possible SP implementation

csi.proto
```diff
@@ -502,6 +549,218 @@ message ControllerServiceCapability {
...
+  // values in documentation.
+  // CO SHALL permit passing through the required secrets.
+  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
+  // all unique secrets that an SP expects must be unique across all CSI
+  // operations.
+  // This information is sensitive and MUST be treated as such (not
+  // logged, etc.) by the CO.
+  // This field is OPTIONAL.
+  map<string, string> create_snapshot_secrets = 3;
+
```
julian-hj: NIT: It would be great if we could consolidate secrets requirements into a single section in the spec rather than having multiple long comment blocks that are repreated but could diverge over time

csi.proto
```diff
@@ -492,6 +468,15 @@ message ControllerServiceCapability {
...
       GET_CAPACITY = 4;
+      CREATE_DELETE_SNAPSHOT = 5;
+      // For storage systems that need to upload the snapshot after
+      // it is cut, LIST_SNAPSHOTS is a required RPC because CO needs
+      // to call it to find out whether the upload is complete.
+      LIST_SNAPSHOTS = 6;
+      // Currently the only way to consume a snapshot is to create
+      // a volume from it. In the future, a revert snapshot function
+      // may be added as another way to consume the snapshot.
+      CREATE_VOLUME_FROM_SNAPSHOT = 7;
```
cpuguy83: Do we have a solid case for drivers that support snapshots that wouldn't support this option?

csi.proto
```diff
@@ -502,6 +487,176 @@ message ControllerServiceCapability {
...
+  SnapshotStatus status = 5;
+}
+
+message SnapshotSize {
+  // The disk size of the snapshot. The purpose of this field is to
+  // give CO guidance on how much space is needed to create a volume
+  // from this snapshot. The size of the volume must not be less than
+  // the size of the source snapshot. This field is OPTIONAL. The value
+  // of this field MUST NOT be negative.
+  int64 size = 1;
```
cpuguy83: Maybe `size_bytes`? Does this need it's own message spec?

csi.proto
```diff
@@ -502,6 +487,176 @@ message ControllerServiceCapability {
...
+// on what action can be taken in the process of snapshotting. Based on
+// this information, CO can issue repeated (idemponent) calls to
+// CreateSnapshot, monitor the response, and make decisions.
+// Note that CreateSnapshot is a synchronous call and it must block
+// until the snapshot is cut. If the cloud provider or storage system
+// does not need to upload the snapshot after it is cut, the status
+// returned by CreateSnapshot should be `READY`.
+// If the cloud provider or storage system needs to upload the snapshot
+// after the snapshot is cut, the status returned by CreateSnapshot
+// should be `UPLOADINDG`. In this case, another RPC ListSnapshots
```
cpuguy83: I'd prefer to take out this recommendation.
Mostly it feels like using `ListSnapshots` to get the status of a single snapshot is strange.
Did we settle on relying on the impotency of `CreateSnapshot` for this?

csi.proto
j-griffith: @cpuguy83 Do you mean the "revert" case?  I do think there are a number of backends that have trouble with this; Older versions of Cinder (and a number of drivers in Cinder still) have trouble with this one.  

I also think the common provider model is create-from-xxxx, during the Snapshot discussions way back we had a lot of discussion around this and I think we had some pretty solid concensus on the revert being a unique call.  

Of course if that's not what you mean here.. then.. well... never mind :)

xing-yang: @cpuguy83 I don't have a solid case for drivers that support snapshots that wouldn't support "CREATE_VOLUME_FROM_SNAPSHOT".  Should we require drivers supporting CREATE_DELETE_SNAPSHOT to support create volume from snapshot?

csi.proto
xing-yang: Sure, I'll rename this to size_bytes.  The reason we use a message spec is because some storage systems do not know the size of volume or snapshot and therefore will leave this field as null.  Using a message spec makes it easier to differentiate between null and a size of 0.

csi.proto
julian-hj: minor typo `UPLOADINDG`

xing-yang: Fixed.

csi.proto
xing-yang: Changed to size_bytes.

csi.proto
xing-yang: @cpuguy83 During last meeting, I believe the decision was not to rely on CreateSnapshot to find out the uploading status, but to use ListSnapshots with a filtering of a single snapshot and get rid of GetSnapshot.

spec.md
```diff
@@ -575,23 +597,33 @@ message CreateVolumeRequest {
...
+  message Snapshot {
+    // Contains identity information for the created snapshot. This
+    // field is REQUIRED. The identity information will be used by
+    // the CO in subsequent calls to refer to the provisioned snapshot.
+    string id = 1;
+
+    // Identity information for the source volume. Note that create
+    // snapshot from snapshot is not supported here so the source has
+    // to be a volume. This field is REQUIRED.
+    string volume_id = 2;
```
cpuguy83: How come we need volume ID in content source? I don't think we'd have this information on `CreateVolume`

spec.md
```diff
@@ -1090,6 +1089,250 @@ message ControllerServiceCapability {
...
+
+  // Contains identity information for the created snapshot. This field
+  // is REQUIRED. The identity information will be used by the CO in
+  // subsequent calls to refer to the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string volume_id = 3;
```
cpuguy83: source_volume_id

spec.md
```diff
@@ -1090,6 +1089,250 @@ message ControllerServiceCapability {
...
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. This field is REQUIRED.
+  string timestamp = 4;
+
+  // The status of a snapshot. A snapshot could have the following
+  // statuses:
+  // UPLOADING - A snapshot is cut and is now being uploaded.
```
cpuguy83: Can we move these descriptions to the relevant message? This makes sure the doc doesn't diverge from the actual message.

spec.md
```diff
@@ -1103,6 +1346,16 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
 
 It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capability if it supports the `CREATE_DELETE_VOLUME` capability: the onus is upon the CO to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
 
+##### `CreateSnapshot`, `DeleteSnapshot`, `ListSnapshots`
+
+The plugin-generated `snapshot_id` is a REQUIRED field for the `DeleteSnapshot` RPC, as opposed to the CO-generated snapshot `name` that is REQUIRED for the `CreateSnapshot` RPC: these fields MAY NOT contain the same value.
```
cpuguy83: s/MAY NOT/MUST NOT/ ?

spec.md
```diff
@@ -1103,6 +1346,16 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
...
+If a `CreateSnapshot` operation times out, leaving the CO without an ID with which to reference a snapshot, and the CO also decides that it no longer needs/wants the snapshot in question then the CO MAY choose one of the following paths:
+
+1. Execute the `ListSnapshots` RPC to possibly obtain a snapshot ID that may be used to execute a `DeleteSnapshot` RPC; upon success execute `DeleteSnapshot`.
```
cpuguy83: Limiting to these two options is problematic.
I do not recall any discussion on the last call that an idempotent `CreateSnapshot` is problematic. Unless I'm mistaken, it should be an expectation that a timed out `CreateSnapshot` will result in a subsequent `CreateSnapshot` expecting the same resource.

spec.md
```diff
@@ -1090,6 +1089,250 @@ message ControllerServiceCapability {
...
+
+A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` controller capability.
+This RPC will be called by the CO to create a new snapshot from a source volume on behalf of a user.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot `name` already exists and is compatible with the specified `volume_id` in the `CreateSnapshotRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateSnapshotResponse`.
+
+Snapshot can be used as the source to provision a new volume.
+CreateVolumeRequest is modified to take an OPTIONAL parameter of a source snapshot.
+Reverting snapshot, where data in the original volume is erased and replaced with data in the snapshot, is an advanced functionality not every storage system can support and therefore is out of scope for the current proposal.
```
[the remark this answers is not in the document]
xing-yang: Do you want me to remove the mentioning of "reverting snapshot?  I added it because people kept asking whether "reverting snapshot" is supported by this proposal.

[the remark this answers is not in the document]
xing-yang: Reading your comments again.  It seems that you just want me to modify the text.  Will do.

spec.md
```diff
@@ -1090,6 +1089,250 @@ message ControllerServiceCapability {
...
+It can be added as a separate function (not in CreateVolume) when proposed in the future.
```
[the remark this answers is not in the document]
xing-yang: Will remove this line.

spec.md
xing-yang: Probably not needed as we already have snapshot id here.  I'll remove volume_id.

spec.md

spec.md
xing-yang: Sure.

spec.md
xing-yang: "MAY NOT" may be better because we don't have a way to ask storage systems not to use the CO-generated "snapshot_id" as the plugin's "snapshot_id".

spec.md
cpuguy83: Ah ok, I re-read spec here and it makes sense. Thanks!

spec.md
```diff
@@ -1090,6 +1084,248 @@ message ControllerServiceCapability {
...
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. This field is REQUIRED.
+  string timestamp = 4;
```
cpuguy83: `created_at`
Also thinking this should be more formal than a string.
A unix nano encoded as an int64 could work
If it is a string then we should make sure to call out the exact format in the spec.

xing-yang: Sure.  Added a description on the format.  Let me know if it looks okay.

spec.md
```diff
@@ -1090,6 +1084,249 @@ message ControllerServiceCapability {
...
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. The format of this field should be a string in UnixNano
+  // time encoded as an int64. This field is REQUIRED.
+  string timestamp = 4;
```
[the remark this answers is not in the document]
cpuguy83: Just to disambiguate some more... I mean the field type could be an int64.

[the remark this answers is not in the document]
xing-yang: Sure.  Changed to int64.  Thanks.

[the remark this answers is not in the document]
cpuguy83: Sorry, can you change the name to something more descriptive, e.g. `created_at`?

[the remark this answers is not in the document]
xing-yang: Sure, changed to "created_at".

xing-yang:
@cpuguy83 Thanks for your reviews!

spec.md
```diff
@@ -1090,6 +1084,249 @@ message ControllerServiceCapability {
 
 If the plugin is unable to complete the ControllerGetCapabilities call successfully, it MUST return a non-ok gRPC code in the gRPC status.
 
+#### `CreateSnapshot`
```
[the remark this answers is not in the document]
xing-yang: @humblec We don't have that in this PR.  Let's start without it and consider adding it later if needed.  Thanks.

xing-yang:
Added descriptions on SnapshotStatus returned by CreateSnapshot and ListSnapshots with filter.

cpuguy83:
This seems like we've gotten pretty close to consensus here. This LGTM. Would be great for others to chime in. Thanks!

jdef:
I still can't review this PR on GH, and instead see an angry unicorn when I
try loading the page.

spec.md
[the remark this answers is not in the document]
humblec: Sure, SGTM.

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+
+A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` controller capability.
+This RPC will be called by the CO to create a new snapshot from a source volume on behalf of a user.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot `name` already exists and is compatible with the specified `source_volume_id` and `parameters` in the `CreateSnapshotRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateSnapshotResponse`.
+
+Snapshot can be used as the source to provision a new volume.
```
humblec: @xing-yang while we promote the snap to a new volume do we have a mechanism to say the new volume should be `READONLY` or ` READWRITE` ? or an implicit mention in snapshot response itself the new volumes will/can only be `READONLY` or `READWRITE` from this snap? may be I missed it.

xing-yang: @humblec In CreateVolumeRequest, there is a VolumeCapability where you can define AccessMode.  Create volume from a snapshot and create a blank volume should work the same way there.

humblec: @xing-yang cool, I missed that field. thanks!

xing-yang:
@jdef Do you still have problem viewing this PR?  Thanks.

jdef:
Thanks for asking, I still cannot view it

jdef:
Is it worth hooking this repo up to https://reviewable.io/ (or something
similar) to get past this GH problem?

xing-yang:
@jdef I'm submitting a new PR and referencing this PR.  Let's see if that solves this GH problem.

xing-yang:
@jdef I opened a new PR here: https://github.com/container-storage-interface/spec/pull/224.  Can you check to see if you can open that one?  Thanks.

xing-yang:
Just want to drop a note here that PR #224 (which I submitted to replace this original PR because some people can no longer open and review this PR) was merged today.  Thank you all for reviewing and providing valuable feedback!

Issue #213, opened 2018-03-30
Clarify Operation Pending Errors


davidz627:
Currently the "Operation pending for volume" errors for each of the RPCs are a little ambiguous. They should be clarified to say that the driver should only return aborted when there is an operation [of another type] pending for the volume. A pending operation of the same type should be treated as an idempotent operation and return success when the original operation completes.

PR #214, opened 2018-03-30
Clarifies Operation Pending Error Description

davidz627:
Fixes: #213

Fix operation pending description to clarify that it should only return `ABORTED` when operation is pending for a different RPC Call Type


davidz627:
@saad-ali @jieyu

davidz627:
added "for a different RPC type " to descriptions of the error

spec.md
```diff
@@ -848,7 +848,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+| Operation pending for volume | 10 ABORTED | Indicates that there is a already an operation pending for a different RPC type for the specified volume. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified volume, and then retry with exponential back off. |
```
julian-hj: Here it is not clear whether this error is really only about RPCs of a different type.  What about multiple controller publish calls for different nodes, in cases where that is allowed?  

If we are really standing behind the statement that the CO should ensure that `there is no more than one call "in-flight" per volume at a given time` then concurrent ControllerPublishVolume calls could also return `ABORTED` here.

spec.md
```diff
@@ -914,7 +914,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
```
julian-hj: Same issue as `ControllerPublishVolume`

spec.md
```diff
@@ -711,7 +711,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
```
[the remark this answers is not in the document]
davidz627: I guess in theory any of the operations could be pending/running on a "volume" with the same name  even if the volume doesn't currently exist. The way you determine whether a volume exists or not could be long-running? This is way over in theory though and seems like an unlikely scenario.

Are you recommending something like "a different RPC type [such as DeleteVolume]...." ?

spec.md
davidz627: Thats a really good point. However, I believe the current wording does allow for multiple controller publish calls for different nodes since that is not a `different RPC type`. It would just require a response of `ABORTED` if say you called `DeleteVolume` while a `ControllerPublishVolume` was still pending.

And I don't think `there is no more than one call "in-flight" per volume at a given time` is a rule at all - it seems to me more like a description of what kind of behavior we may expect or strive for, as its prefaced by `In general` and followed by examples of how it may not be true.

spec.md
davidz627: See answer for `ControllerPublishVolume`

PR #215, opened 2018-03-30
Removed redundant 'call' from all 'RPC call' text

davidz627:
nit: Remote Procedure Call [call]

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -490,7 +490,7 @@ If the plugin is unable to complete the GetPluginCapabilities call successfully,
 
 #### `Probe`
 
-A Plugin MUST implement this RPC call.
+A Plugin MUST implement this RPC.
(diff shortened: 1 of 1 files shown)


davidz627:
@saad-ali @jieyu PTAL

davidz627:
nvm. I have been informed that this could actually be the correct nomenclature. Could one of you confirm

julian-hj:
Hmm, yeah, hard to know what is exactly "correct" but since "RPC call" appears all over the RPC RFC it is probably not "incorrect"

https://tools.ietf.org/html/rfc5531

Personally I think this is a bummer since "RPC call" seems totally redundant to me, but then again so do "crochet hook" and "Please RSVP" and people say those things all the time in polite company too.

Probably we can close this one.

PR #218, opened 2018-04-12
spec: use MUST when describing CapacityRange field reqs

jdef:


csi.proto
```diff
@@ -243,15 +243,15 @@ message VolumeCapability {
 }
 
 // The capacity of the storage space in bytes. To specify an exact size,
-// `required_bytes` and `limit_bytes` can be set to the same value. At
+// `required_bytes` and `limit_bytes` MUST be set to the same value. At
```
cpuguy83: Minor nit here, but a `MUST` here might be overkill.

jdef:
i'd rather err on the side of **clarity** w/ respect to expectations here.
a CO that's doing precise resource accounting and wants to specify the
EXACT size of the desired volume MUST use the same value for required_bytes
and limit_bytes. lower case words "must" or "can" are not the right
terminology for requirements. if a CSI plugin cannot honor the field rules
specified here (read: cannot provide a volume of the exact size requested)
then it SHALL be considered an error- use of MUST makes it clear that the
descriptions for these fields dictate **requirements** vs suggestions.

jdef:
SHOULD is an official recommendation.
MUST and SHALL are official requirements.

shall I replace MUST with SHALL?

PR #219, opened 2018-04-12
travis: bump golang to 1.9.5

jdef:


cpuguy83:
Not 1.10.1?

jdef:
I could have bumped it higher. My goal was to move to a non EOL golang
release.

Issue #217, opened 2018-04-10
Allow CSI drivers to specify how many volumes can be attached to single node


gnufied:
As part of fixing - https://github.com/kubernetes/features/issues/554 , I believe we need to extend the specs to add support for max number of volumes that can be attached to a single node. For sake of discussion lets call it - `max_volume_count`.

There are couple of places we can do this:

1. `GetPluginInfoResponse` seems like one of the easiest places. The problem of course is - `manifest` field inside is a `map[string]string`, so we will have to convert string to int before using the field.  But there may be bigger problem that - same driver running on 2 different nodes may have different values of `max_volume_count`. For example - EBS driver on node A may support 40 while it may only support 20 volumes on node B. 

2. Another option is something like `NodeGetCapabilitiesResponse`.  

Does anyone have any other thoughts on this?

msau42:
The max count should be reported per node.  Each node could support a different number of volumes depending on its size.

cpuguy83:
I'm not sure I understand the design of this. Can you give the details of when/why this would be needed?

msau42:
Many cloud provider's storage have limits on how many of those devices can be attached to a node, and the limit can be dependent on node size.

gnufied:
Yes for example - only 40 EBS volumes (it can vary sometimes depending on node size) can be attached to a EC2 node.

The CSI driver needs to provide hints to CO, so as CO can stop scheduling new containers to the node if node is already at maximum capacity.

cpuguy83:
Ah, ok. This makes sense.

chhsia0:
In the community meeting today we discussed that this could be used for making the scheduler smarter, where the scheduler keeps track of the current count of volumes "available" to a node and get the max count from the node to infer the remaining capacity, and use this information to help making scheduling decisions.

The problem of not directly providing the current capacity is that this is a dynamic property and there might be a race between querying the current capacity from a node and scheduling some other workloads. This makes me think that if the available capacity can be obtained at the controller side given a node id. If yes, then we could have a controller API to return the "available_publish_capacity" instead, and CO could potentially synchronize its interaction with the controller plugin to avoid races.

Thoughts?

msau42:
The problem is synchronizing the scheduler with "current available" introduces signficant latency and still doesn't completely solve race conditions.  The scheduler has to wait until the controller updates the "current count" before scheduling the next workload.  And you still have the race where after the scheduler reads the value, some out-of-band attach could happen to modify the count.

[...]

PR #221, opened 2018-04-17
Add node volume capacity to NodeGetCapability

gnufied:
Fixes https://github.com/container-storage-interface/spec/issues/217

cc @saad-ali


saad-ali:
I'm not sure if `NodeGetCapabilitiesResponse` is the best home for this. I was thinking that since the `ControllerGetCapabilitiesResponse` contains the `PUBLISH_UNPUBLISH_VOLUME` response (indicating that a plugin can do attach or not), that maybe a better home for this. But, then I realized the limit may not be the same for every node, so ability to define it per node could be useful. So I don't have a better suggestion.

That said I would call it something other then "capacity", `MaxVolumesPerNode` seems more accurate to me.

e4jet:
+1 for MaxVolumesPerNode, NodeCapacity could be misconstrued.

gnufied:
If a volume plugin returns non-zero value of this field, it can be safely assumed that - that is the Maximum number of volumes that can be published/staged/mounted to the node of that type. 

So volume types that come with this limit - apply the limit to all stages of CSI lifecycle. So, that is why I  like a simpler and broader name on purpose.  So +1 to `MaxVolumesPerNode` from me, because once a CSI driver returns non-zero value for this field, it doesn't matter - you can only have that many maximum volume of that type on the node. The distinction that you can only stage `MaxVolumesPerNode` becomes immaterial and using a generic name will allow us to reuse the field name for volume types where distinction between stage/publish is not so clear cut.

jdef:
If I'm running multiple CSI plugins on a node, and each plugin is reporting
its own separate limit, how does that solve the root problem? For example,
let's say that my node's OS allows me to attach a max of X devices and that
I have multiple CSI plugin types installed on said node: the sum of each
plugin's MaxVolumesPerNode may be greater than X.

AFAICT the use case this is trying to solve is to avoid exceeding the
threshold of "max devices attached to a OS/VM". It's not clear to me that
this API would accomplish that: see
https://github.com/container-storage-interface/spec/issues/217#issuecomment-382089755

gnufied:
We are not trying to solve problem of "total number of devices of all types that can be attached to the node". We are trying to solve problem of "total number of device of particular volume type that can be attached to the node". 

The limitation is not imposed by OS or Node itself, but is imposed by storage vendor. Often the reason for imposing such a limit is - because volume is on  network, only so many volumes can be attached to a node before we saturate the network.   It is possible that, other types of volumes can still be attached to node because they may or may not share the transport mechanism.

Using EBS as an example - a node with higher network throughput can support more number of devices and even though we reach maximum number of EBS devices, local devices can still be attached to the node and not count towards limit of EBS devices.

jdef:
Thanks for the clarification. Please update the docs for the field to
include a similar explanation.

jdef:
Also, CSI doesn't have the concept of "attach" - there are only "stage" and
"publish". What's still not clear to me is how a CO can reason about when
attachment occurs: without this, I don't understand how this "max attached
volumes" field is useful. Should a CO assume that if a plugin supports
controller-publish that this translates to "attachment"? If a plugin does
not support controller-publish, what other OP indicates attachment,
node-staging?

cpuguy83:
It's also asking the CO to make a pretty strong assumption that it is the only thing that would cause this max number to be reached.

gnufied:
@jdef using Kubernetes as an example - when CSI driver is registered to the node(https://github.com/kubernetes-csi/driver-registrar) , it will for example modify Kubernetes's node object with a field that will indicate, how many volumes of this type can be attached to the node.  Kubernetes scheduler can then use that information to pick nodes for scheduling new pods. Any node which is nearing the limit will be skipped.

I know that CSI doesn't uses "attach" word - but that does not matter, I am using the word loosely.  For most volume types that don't support controller-publish (For example - NFS, GlusterFS, Portworx), I would assume that the CSI driver for these volume plugins will return default value (`0`) and hence no such limits will apply to that volume type, but I don't know if it will hold true in future. 

The way CO will count, how many volumes of that type are being used on the node is by counting number of containers running on the node using that volume type (and removing duplicates).  It does not matter if volumes were controller-published or directly node-published. It just happens so that - this limit will mostly affect controller-published (or attachable) volume types but how this field will be used depends on author of CSI driver. 

@cpuguy83 it is possible that volumes of that type were published to a node, outside purview of CO or CSI driver but we have to make best case assumption and leave it to the CO to decide - how to count the actual number of published volumes.

jdef:
> The way CO will count, how many volumes of that type are being used on the node is by counting number of containers running on the node using that volume type (and removing duplicates).

CSI does not require a container to be running on the node in order to publish/stage a volume. If no containers are running on the node, how does a CO count "attached" volumes?

> It just happens so that - this limit will mostly affect controller-published (or attachable) volume types but how this field will be used depends on author of CSI driver.

"... depends on the author of the CSI driver" feels pretty fuzzy, and not in a good way. The spec should be clear w/ respect to how this number is determined by the plugin, and also how a CO may reason about whether it's reached some threshold.

cpuguy83:
I don't think there is any way for the CO to decide that. The CO relies on the SP to deal with this kind of info.

gnufied:
>I don't think there is any way for the CO to decide that. The CO relies on the SP to deal with this kind of info.

hmm - okay. You mean CO can not determine how many volumes of particular type is in use on a node and it needs SP to deal with this kind of stuff? 

So - would it help if we had a separate rpc call that returns list of volumes published/staged to the node? A bit like current `ListVolumes` but scoped to a node.  If yes - can we add that as a follow up change or must it be done within this PR?

gnufied:
>So just thinking about what we want to accomplish... we want the SP to be able to give hints to the scheduler so that it doesn't needlessly schedule a workload to a node that the node can't handle. We want one of those hints to be storage capacity.
Is that accurate?

Yeah - that is correct.

childsb:
i think there's too many variables in determining a 'max' to add this as API... memory, CPU, network could all cause a ceiling on the number of attachable devices long before the actual limit of the storage subsystem are reached... And what if the "max" changes as resources become consumed.... 

for 'generic' type transports like iscsi, nfs, fibre channel the 'max' may depend on what the storage system exporting the volumes is or the other infrastructure around the node.  these things are something the admin should plan and not burden the storage layer with.  i think this function belongs to the CO & scheduler.

cpuguy83:
There is no way for the CO to know this information. The reason for this
discussion is so the SP can provide hints to the CO scheduler so as to not
waste time scheduling to a node that can't handle the workload because the
workload needs something that the SP can't provide.

childsb:
What storage platform provides device connect limit as API?  For AWS/ebs its buried in documentation-- and even the "upper bound" isn't guaranteed..  i dont see how the storage layer could know its own limit better then the admin. EBS is a good thing to pick on.... It's node connection limit has changed over the last year.. This would mean the csi-ebs driver would either need this as a configurable value, or its forever hard coded at a certain value.

I'm not saying the CO should "know" this, but the CO is the configuration point for the admin to set it.

> There is no way for the CO to know this information. The reason for this
discussion is so the SP can provide hints to the CO scheduler so as to not
waste time scheduling to a node that can't handle the workload because the
workload needs something that the SP can't provide.

rootfs:
while there is a good use case, I concur with @childsb that this is probably not exact science. Per AWS doc [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html). Any upfront hints from the node would then be misleading.

```text
Important

Attaching more than 40 volumes to a Linux instance is supported on a best effort basis only and is not guaranteed.
```

cpuguy83:
So yes, it won't be exact. But we can make a best effort to give the
scheduler *some* information. If ultimately the scheduler decides to put a
workload on a node that can't handle it, then the error return should at
least be sufficient that the CO knows why it failed can try another node.

gnufied:
I agree with @cpuguy83 . It is true that volume limits for different volume types is not exact science or some API call away.  But shipping production CSI drivers without any limits is problematic too. Without reasonable defaults most users will be forced to configure individual limit on each node and without that configuration most users will *only* notice the error when suddenly workloads will refuse to start. 

If anything our experience with Kubernetes hardcoded(that exists right now) default teaches us - driver provided defaults works for majority of cases.  Yes it is not perfect (because source/SP does not provide perfect answer) but it reduces need of tuning a parameter from get-go.

gnufied:
> Yet there is another counter argument. CSI driver is not the sole volume manager on the node. There could well be the case another out-of-band volume manager there that attaches some volumes to the instance. In such case, the max volumes reported by the node won't be accurate.

Strictly speaking max. volume on a node is a invariant and does not depend on how many volumes were attached out of band. Yes there could be volumes attached to the node outside CSI and that is why we have a ongoing discussion about adding "list of total volumes attached to the node".

> A good error handling than error avoidance makes better sense.

Can you elaborate more?

childsb:
A couple questions--

> For EBS all node types except M5/C5 will return limit of 40. For C5/M5 instance types the limit will be 28. 

AWS will let you pay for higher attach rates up to an undocumented MAX.. there are also instances where the MAX limitation may be less then those numbers due to the OS..  In those scenarios how would the admin adjust the actual 'max' ? ((also dont we set this value to 50 in kubernetes?))

Similar to questions earlier in the thread, for things like NFS or iSCSI where the MAX depends more on the environment or resource usage, what would be the MAX there, or how would it be determined? And is the MAX value expected to be fixed for the lifetime of the node, or can it change as resources free up?

> A good error handling than error avoidance makes better sense.
I read this as-- If COs expect this to be determinate, they may skimp on error recovery for over scheduling a node.  There's no good way to convey that the MAX is indeterminate.  Calling it a "hint" might be a good start.  The real work is how the CO handles scheduling failure that may result from this being wrong... And if the CO assumes the MAX value is indeterminate and implements the correct recovery API, why have the call at all.

gnufied:
@saad-ali @jdef @childsb PTAL. Introduced new `NodeGetInfo` call and deprecated `NodeGetId` call. Also - added special mention of the fact that `MaxVolumesPerNode` applies mainly to Controller published volumes.

spec.md
```diff
@@ -316,7 +316,7 @@ service Controller {
     returns (GetCapacityResponse) {}
 
   rpc ControllerGetCapabilities (ControllerGetCapabilitiesRequest)
-    returns (ControllerGetCapabilitiesResponse) {}  
+    returns (ControllerGetCapabilitiesResponse) {}
```
jdef: Unrelated whitespace change?

spec.md
```diff
@@ -333,10 +333,15 @@ service Node {
     returns (NodeUnpublishVolumeResponse) {}
 
   rpc NodeGetId (NodeGetIdRequest)
-    returns (NodeGetIdResponse) {}
+    returns (NodeGetIdResponse) {
+    option deprecated = true;
```
jdef: Please include a comment directing the reader to the new API. Also, this PR schedules this RPC for removal prior to publishing CSI v1.0... should include such comment

jdef: We should make it clear that prior to v1.0 plugins that implement NodeGetInfo MUST implement this as well

spec.md
```diff
@@ -1386,6 +1391,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 #### `NodeGetId`
 
+`NodeGetId` RPC call is deprecated. Users of this RPC call are recommended to use `NodeGetInfo`.
```
jdef: s/ are recommended to/SHOULD/

spec.md
```diff
@@ -1450,6 +1457,45 @@ message NodeServiceCapability {
...
+
+message NodeGetInfoResponse {
+  // The ID of the node as understood by the SP which SHALL be used by
+  // CO in subsequent `ControllerPublishVolume`.
+  // This is a REQUIRED field.
+  string node_id = 1;
+
+  // Maximum number of volumes that can be ControllerPublished
+  // to the node. If a Plugin does not define a value here
+  // it is upto CO to decide how many volumes of this type
```
jdef: Upto is not a word

csi.proto
```diff
@@ -672,3 +677,19 @@ message NodeServiceCapability {
...
+message NodeGetInfoResponse {
+  // The ID of the node as understood by the SP which SHALL be used by
+  // CO in subsequent `ControllerPublishVolume`.
+  // This is a REQUIRED field.
+  string node_id = 1;
+
+  // Maximum number of volumes that can be ControllerPublished
+  // to the node. If a Plugin does not define a value here
+  // it is upto CO to decide how many volumes of this type
+  // can be ControllerPublished to the node.
```
jdef: ControllerPublished is not the name of the RPC. Please rephrase

csi.proto
```diff
@@ -672,3 +677,19 @@ message NodeServiceCapability {
...
+  // This field is OPTIONAL
+  int64 MaxVolumesPerNode = 2;
```
jdef: Lower snake case for field names

spec.md
```diff
@@ -1450,6 +1457,45 @@ message NodeServiceCapability {
...
+}
+
+message NodeGetInfoResponse {
+  // The ID of the node as understood by the SP which SHALL be used by
+  // CO in subsequent `ControllerPublishVolume`.
+  // This is a REQUIRED field.
+  string node_id = 1;
+
+  // Maximum number of volumes that can be ControllerPublished
+  // to the node. If a Plugin does not define a value here
```
jdef: Zero is considered undefined. Let's make that clear

spec.md
gnufied: Yep but may be it is a good thing? This line had a trailing empty whitespace

jdef: the "quality" of the change is fine. if this is to be included, please do so as a separate commit

gnufied:
@jdef fixed all your review comments. PTAL. Failure here seems unrelated to changes in the PR.  I have filed https://github.com/container-storage-interface/spec/issues/226 to track that one down

spec.md
```diff
@@ -332,11 +332,23 @@ service Node {
   rpc NodeUnpublishVolume (NodeUnpublishVolumeRequest)
     returns (NodeUnpublishVolumeResponse) {}
 
+  // NodeGetId is being deprecated in favor of
+  // NodeGetInfo and will be removed in CSI 1.0.
+  // Existing drivers however may depend on this RPC call
```
jdef: nit: missing commas "Existing drivers, however, may depend on this..."

spec.md
```diff
@@ -1386,6 +1398,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 #### `NodeGetId`
 
+`NodeGetId` RPC call is deprecated. Users of this RPC call SHOULD use `NodeGetInfo`.
```
jdef: nit: different sentences should be on different lines

spec.md
```diff
@@ -1450,6 +1464,45 @@ message NodeServiceCapability {
...
 If the plugin is unable to complete the NodeGetCapabilities call successfully, it MUST return a non-ok gRPC code in the gRPC status.
 
+
+#### `NodeGetInfo`
+
+A Node Plugin MUST implement this RPC call if the plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+The CO SHOULD call this RPC for the node at which it wants to place the workload.
+The result of this call will be used by CO in `ControllerPublishVolume`.
+
```
jdef: nit: remove extra blank line (only 1 is needed between the above paragraph and the protobuf block)

spec.md
```diff
@@ -1450,6 +1464,45 @@ message NodeServiceCapability {
...
+
+```protobuf
+message NodeGetInfoRequest {
+}
+
+message NodeGetInfoResponse {
+  // The ID of the node as understood by the SP which SHALL be used by
+  // CO in subsequent `ControllerPublishVolume`.
```
jdef: nit: grammar. suggest: "... in subsequent calls to `ControllerPublishVolume`."

spec.md
```diff
@@ -1450,6 +1464,45 @@ message NodeServiceCapability {
...
+  // This is a REQUIRED field.
+  string node_id = 1;
+
+  // Maximum number of volumes that controller can publish
+  // to the node. If value is not set or zero CO SHALL
+  // decide how many volumes of this type can be published
+  // by the controller to the node.
+  // This field is OPTIONAL
```
jdef: nit: please end w/ a period.

jdef: also, maybe include a statement along the lines that negative values are not allowed here

spec.md
gnufied: In spec.md - it is all in one line. github is wrapping the line. But still - I do not under why different sentences should start on different lines. We start new sentence on same line all the time when writing a paragraph don't we?

gnufied:
@jdef pushed another round of changes to address your review comments. PTAL. build is still broken because of linked issue.

spec.md
```diff
@@ -332,11 +332,22 @@ service Node {
...
-    returns (NodeGetIdResponse) {}
+    returns (NodeGetIdResponse) {
+    option deprecated = true;
+  }
 
   rpc NodeGetCapabilities (NodeGetCapabilitiesRequest)
     returns (NodeGetCapabilitiesResponse) {}
+
+  // Prior to CSI 1.0 - CSI plugins MUST implement both NodeGetId and
+  // NodeGetInfo RPC calls
```
jdef: please end sentences with a period.

spec.md
jdef: we generally try to follow the rules documented here: https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md

gnufied:
@jdef fixed. thank you.

spec.md
```diff
@@ -1450,6 +1464,44 @@ message NodeServiceCapability {
...
+A Node Plugin MUST implement this RPC call if the plugin has `PUBLISH_UNPUBLISH_VOLUME` controller capability.
+The Plugin SHALL assume that this RPC will be executed on the node where the volume will be used.
+The CO SHOULD call this RPC for the node at which it wants to place the workload.
+The result of this call will be used by CO in `ControllerPublishVolume`.
+
+```protobuf
+message NodeGetInfoRequest {
+}
+
+message NodeGetInfoResponse {
```
bswartz: This change doesn't allow nodes to return any more opaque data to the controller than the old spec. It's still just a single string (node_id). An opaque string map would give more flexibility to implementers.

gnufied: Yeah it was discussed that, `NodeGetInfo` will be extended to cover any immutable/invariant data about the node. So this field has to be extended. We are going to need a follow up PR because this PR just concerns itself with node volume limits.

gnufied:
rebased with master and branch is green now. I think I have addressed all outstanding concerns. @saad-ali @jdef PTAL

spec.md
```diff
@@ -332,11 +332,22 @@ service Node {
   rpc NodeUnpublishVolume (NodeUnpublishVolumeRequest)
     returns (NodeUnpublishVolumeResponse) {}
 
+  // NodeGetId is being deprecated in favor of NodeGetInfo and will be
+  // removed in CSI 1.0. Existing drivers, however may depend on this
```
jdef: still missing a comma after `however`. the text should appear this way `, however,`

spec.md
jdef: some history: I think the issue w/ returning a `map` or `repeated` for `node_id` was that it doesn't hash well: implementations want to use the `node_id` as a dictionary key, which is more difficult unless `node_id` has a primitive type.

spec.md
jdef: thanks

gnufied:
@saad-ali @jdef PTAL. Answered all outstanding concerns.

jdef:
please resolve conflicts, rebase if needed

saad-ali:
LGTM, merging.

spec.md
```diff
@@ -1733,6 +1747,44 @@ message NodeServiceCapability {
 
 If the plugin is unable to complete the NodeGetCapabilities call successfully, it MUST return a non-ok gRPC code in the gRPC status.
 
+
+#### `NodeGetInfo`
```
verult: Is `NodeGetInfo` guaranteed to return the same result every time, i.e. can node info change over time? @saad-ali what if topology is included?

Either way I'd recommend making the answer to this question explicit.

jdef:
I'm pretty sure we decided that a successful response for this RPC should
not change over time. If we haven't stated that directly then we probably
should, for clarity.

spec.md
gnufied: Yeah I think we meant this value to be invariant through life cycle of the node. I will open a PR to make this explicit. 

There may be an exceptional case though - where node is shutdown, modified and rebooted. For example - on EC2 it is possible to change instance type of a instance.  is topology falls under similar case? I am not sure, if we worded explicitly what happens to node-id when a node is modified and rebooted.

PR #224, opened 2018-04-25
Add Snapshot Support in CSI Spec

xing-yang:
This PR proposes to add snapshot support in CSI specification.
It proposes to add 3 new controller service RPCs, including
CreateSnapshot, DeleteSnapshot, and ListSnapshots, and modify
the existing ControllerGetCapabilitiesResponse and
CreateVolumeRequest.

There are several cases for using the snapshot data in a volume:

1. Create volume from snapshot: In this case, a new volume will be
created on the storage backend and the data in the snapshot will
be copied to the new volume. This case will be addressed by this
proposal.

2. Revert snapshot: In this case, data in the original volume
will be erased and replaced with data in the snapshot. This case
is out of scope for this proposal and should be handled by a
function of its own (not to be mixed with create volume from
snapshot) if it is proposed in the future.

The following changes will be made in the Service Controller:
- Add CreateSnapshot
- Add DeleteSnapshot
- Add ListSnapshots
- Modify CreateVolumeRequest to support creating a new volume from
  snapshot.
- Modify ControllerGetCapabilitiesResponse to return
  CREATE_DELETE_SNAPSHOT and LIST_SNAPSHOTS support capabilities.
- Creating volume from snapshot is a required functionality for plugins
  supporting CREATE_DELETE_SNAPSHOT.

This proposal was drafted here:
https://docs.google.com/document/d/1oXVuDTNhXerr1UJ48tQrxF9J6vbm2LHVVC3z28655Qo/edit#

The PR https://github.com/container-storage-interface/spec/pull/207 is getting very big (possibly due to the large number of comments) and some people can't view it on Github, so close that one and submit a new PR here instead.


spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
 
+#### `CreateSnapshot`
+
+A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` controller capability.
+This RPC will be called by the CO to create a new snapshot from a source volume on behalf of a user.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot `name` already exists and is compatible with the specified `source_volume_id` and `parameters` in the `CreateSnapshotRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateSnapshotResponse`.
+
+Snapshot can be used as the source to provision a new volume.
```
jdef: s/Snapshot can/A snapshot MAY/

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+}
+
+// The information about a provisioned snapshot.
+message Snapshot {
+  // SnapshotSize is the disk size of the snapshot. The purpose of this
+  // field is to give CO guidance on how much space is needed to create
+  // a volume from this snapshot. The size of the volume must not be
+  // less than the size of the source snapshot. This field is OPTIONAL.
+  // If SnapshotSize is not set, it indicates that this size is unknown.
+  SnapshotSize size = 1;
```
jdef: why `SnapshotSize` instead of a well-known-type? https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/int64-value

saad-ali:
Woo I can actually open this without unicorn of death, thanks @xing-yang !

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+
+  // Contains identity information for the created snapshot. This field
```
jdef: nit: s/Contains identity information for the create snapshot./Uniquely identifies a snapshot and is generated by the plugin. It will not change over time./

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+  // subsequent calls to refer to the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. The format of this field should be a UnixNano time encoded
```
jdef: is `UnixNano` a standard? it might be a golang API but CSI isn't tied to a specific lang - can we find a more language neutral term (once that's also well-defined)?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+  // If SnapshotSize is not set, it indicates that this size is unknown.
+  SnapshotSize size = 1;
+
+  // Contains identity information for the created snapshot. This field
+  // is REQUIRED. The identity information will be used by the CO in
+  // subsequent calls to refer to the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
```
jdef: we don't support this yet, but might we at some point? Is it worth creating a message type to give us flexibility here (like we did w/ VolumeContentSource)? For example, perhaps a SnapshotSource type? If not, that's OK we can add one later if needed.

spec.md
```diff
@@ -379,6 +390,17 @@ This string MAY be surfaced by CO to end users.
...
+
+Secrets may be required by plugin to complete a RPC request.
+A secret is a string to string map where the key identifies the name of the secret (e.g. "username" or "password"), and the value contains the secret data (e.g. "bob" or "abc123").
+Each key MUST consist of alphanumeric characters, '-', '_' or '.'.
+Each value MUST contain a valid string. An SP MAY choose to accept binary (non-string) data by using a binary-to-text encoding scheme, like base64.
+An SP SHALL advertise the requirements for required secret keys and values in documentation.
+CO SHALL permit passing through the required secrets.
+A CO MAY pass the same secrets to all RPCs, therefore the keys for all unique secrets that an SP expects must be unique across all CSI operations.
+This information is sensitive and MUST be treated as such (not logged, etc.) by the CO.
+
```
saad-ali: nit: I prefer having this information on each field instead of in a common place because it makes it easy for a reader to discover the requirements when it is a comment on the field rather then an independent paragraph that they have to find.

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+  // give CO guidance on how much space is needed to create a volume
+  // from this snapshot. The size of the volume must not be less than
+  // the size of the source snapshot. This field is OPTIONAL. The value
+  // of this field MUST NOT be negative.
+  int64 size_bytes = 1;
+}
+
+// The status of a snapshot. A snapshot could have the following
+// statuses:
+// UPLOADING - A snapshot is cut and is now being uploaded.
```
jdef: UPLOADING and READY have the same definitions in the enum below. do we need to repeat that information here?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// READY - A snapshot is ready for use.
+// Some cloud providers will upload the snapshot to a location in
+// the cloud (i.e., an object store) after the snapshot is cut.
+// Uploading is a long process that could take hours. At this phase,
```
jdef: s/is a/may be a/

spec.md
```diff
@@ -575,23 +597,28 @@ message CreateVolumeRequest {
...
+  // Contains all attributes of the source that is used to create
+  // the volume from. This field is OPTIONAL.
+  VolumeContentSource content_source = 6;
+}
+
+// Specifies what source the volume will be created from. One of the
+// source_type fields MUST be specified. Only Snapshot is
+// a supported VolumeContentSource at the moment.
+message VolumeContentSource {
+  message Snapshot {
```
saad-ali: This is confusing, because there is already a top level (global) `message Snapshot`.

Maybe call it `SnapshotSource` instead of just `Snapshot`?

spec.md
```diff
@@ -575,23 +597,28 @@ message CreateVolumeRequest {
...
-  // A CO MAY pass the same secrets to all RPCs, therefore the keys for
-  // all unique secrets that an SP expects must be unique across all CSI
-  // operations.
-  // This information is sensitive and MUST be treated as such (not
-  // logged, etc.) by the CO.
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // Contains all attributes of the source that is used to create
+  // the volume from. This field is OPTIONAL.
```
saad-ali: nit:

```
// If specified, the new volume will be pre-populated with data from this source.
// This field is OPTIONAL.
```

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+}
+
+// The status of a snapshot. A snapshot could have the following
+// statuses:
+// UPLOADING - A snapshot is cut and is now being uploaded.
+// READY - A snapshot is ready for use.
+// Some cloud providers will upload the snapshot to a location in
+// the cloud (i.e., an object store) after the snapshot is cut.
+// Uploading is a long process that could take hours. At this phase,
+// since the snapshot is already cut, a `thaw` operation can be
```
jdef: `thaw` and `freeze` are CO-side pseudo-operations that execute outside the context of CSI. this is the first time they are mentioned in the spec right? It may be worth elaborating on the expected container state (w/ respect to snapshots) in the markdown section of the spec Then, these more terse sentences will have been properly framed for the reader.

spec.md
```diff
@@ -699,6 +726,10 @@ message Volume {
   // the same attributes. This field is OPTIONAL and when present MUST
   // be passed to volume validation and publishing calls.
   map<string, string> attributes = 3;
+
+  // All information about the source where the volume is created
+  // from. This field is OPTIONAL.
```
saad-ali: nit:

```
// If specified, indicates that the volume is not empty and is
// pre-populated with data from the specified source.
// This field is OPTIONAL.
```

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// READY - A snapshot is ready for use.
+// Some cloud providers will upload the snapshot to a location in
+// the cloud (i.e., an object store) after the snapshot is cut.
+// Uploading is a long process that could take hours. At this phase,
+// since the snapshot is already cut, a `thaw` operation can be
+// performed so application can start running again (assuming a
+// `freeze` operation was done before taking the snapshot). The
+// status of the snapshot will become `READY` after the upload is
+// complete.
+// For cloud providers and storage systems that don't have the
```
jdef: I LOVE the detail in these paragraphs. I think it might be more appropriate to move the bulk of this text into the markdown section of the spec and keep the commentary for the protos a bit more concise. Perhaps a better home would be the "RPC Interactions" section?

spec.md
```diff
@@ -1074,6 +1058,16 @@ message ControllerServiceCapability {
...
+      CREATE_DELETE_SNAPSHOT = 5;
+      // For storage systems that need to upload the snapshot after
+      // it is cut, LIST_SNAPSHOTS is a required RPC because CO needs
+      // to call it to find out whether the upload is complete.
+      LIST_SNAPSHOTS = 6;
+      // Currently the only way to consume a snapshot is to create
+      // a volume from it. Therefore plugins supporting
+      // CREATE_DELETE_SNAPSHOT should also support
+      // CREATE_VOLUME_FROM_SNAPSHOT.
+      CREATE_VOLUME_FROM_SNAPSHOT = 7;
```
saad-ali: If this is a requirement for `CREATE_DELETE_SNAPSHOT` why have it as a separate capability?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// action can be taken in the process of snapshotting. Based on
+// this information, CO can issue repeated (idemponent) calls to
+// CreateSnapshot, monitor the response, and make decisions.
+// Note that CreateSnapshot is a synchronous call and it must block
+// until the snapshot is cut. If the cloud provider or storage system
+// does not need to upload the snapshot after it is cut, the status
+// returned by CreateSnapshot should be `READY`.
+// If the cloud provider or storage system needs to upload the snapshot
+// after the snapshot is cut, the status returned by CreateSnapshot
+// should be `UPLOADING`. CO can continue to call CreateSnapshot while
```
jdef: s/should/SHALL/

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// A `CREATING` status is not included here because CreateSnapshot
+// is synchronous and will block until the snapshot is cut.
+// The SnapshotStatus parameter provides guidance to the CO on what
+// action can be taken in the process of snapshotting. Based on
+// this information, CO can issue repeated (idemponent) calls to
+// CreateSnapshot, monitor the response, and make decisions.
+// Note that CreateSnapshot is a synchronous call and it must block
+// until the snapshot is cut. If the cloud provider or storage system
+// does not need to upload the snapshot after it is cut, the status
+// returned by CreateSnapshot should be `READY`.
```
jdef: s/should/SHALL/

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// If the cloud provider or storage system needs to upload the snapshot
+// after the snapshot is cut, the status returned by CreateSnapshot
+// should be `UPLOADING`. CO can continue to call CreateSnapshot while
+// waiting for the upload to complete until the status becomes `READY`.
+// Note that CreateSnapshot no longer blocks after the snapshot is cut.
+// Alternatively, another RPC ListSnapshots can be called with
```
jdef: remove "RPC" from this line?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+message CreateSnapshotResponse {
+  // Contains all attributes of the newly created snapshot that are
+  // relevant to the CO along with information required by the Plugin
+  // to uniquely identify the snapshot. This field is REQUIRED.
+  Snapshot snapshot = 1;
+}
+
+// The information about a provisioned snapshot.
+message Snapshot {
+  // SnapshotSize is the disk size of the snapshot. The purpose of this
```
saad-ali: nit: `disk size of the snapshot` -> `disk size of the restored snapshot`?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// until the snapshot is cut. If the cloud provider or storage system
+// does not need to upload the snapshot after it is cut, the status
+// returned by CreateSnapshot should be `READY`.
+// If the cloud provider or storage system needs to upload the snapshot
+// after the snapshot is cut, the status returned by CreateSnapshot
+// should be `UPLOADING`. CO can continue to call CreateSnapshot while
+// waiting for the upload to complete until the status becomes `READY`.
+// Note that CreateSnapshot no longer blocks after the snapshot is cut.
+// Alternatively, another RPC ListSnapshots can be called with
+// snapshot_id as filtering repeatedly to wait for the upload to
```
jdef: a/as/and/ ? maybe "called repeatedly" (earlier in the sentence) instead of "filtering repeatedly" here? maybe it's worth using the term "poll" or "polling" to describe the process we're expecting COs to use here?

spec.md
saad-ali: nit: Move these detailed descriptions above each enum below -- having two sets of descriptions means that they may become out of sync over time.

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+// complete. ListSnapshots should return immediately with the current
```
jdef: "... should return immediately..."

I mean, it's a blocking API. We don't know how long it will take. It shouldn't take very long, but I don't know why that's important to call out here.

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+| Operation pending for snapshot | 10 ABORTED | Indicates that there is already an operation pending for the specified snapshot. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per snapshot at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same snapshot. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified snapshot, and then retry with exponential back off. |
+| Call not implemented | 12 UNIMPLEMENTED | DeleteSnapshot call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` to discover Plugin capabilities. |
+
+
+#### `ListSnapshots`
+
+A Controller Plugin MUST implement this RPC call if it has LIST_SNAPSHOTS capability.
+The Plugin SHALL return the information about all the snapshots on the storage system regardless of how they were created.
```
saad-ali: nit: Except if `snapshot_id` is provided? Maybe word it so that like `all snapshots on the storage system within the given parameters regardless of how they were created`?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+
+  // Identity information for a specific snapshot. This field is
+  // OPTIONAL. It can be used to list only a specific snapshot.
+  // ListSnapshots will return with current snapshot information
+  // and will not block if the snapshot is being uploaded.
+  string snapshot_id = 4;
+}
+
+message ListSnapshotsResponse {
+  message Entry {
```
saad-ali: nit: Why wrap `Snapshot` object? I know `ListVolumes` does this as well. I can't remember why? @jdef do you remember?

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+| Call not implemented | 12 UNIMPLEMENTED | CreateSnapshot call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` to discover Plugin capabilities. |
+| Not enough space to create snapshot | 13 RESOURCE_EXHAUSTED | There is not enough space on the storage system to handle the create snapshot request. | Caller should fail this request. Future calls to CreateSnapshot may succeed if space is freed up. |
+
+
+#### `DeleteSnapshot`
+
+A Controller Plugin MUST implement this RPC call if it has CREATE_DELETE_SNAPSHOT capability.
+This RPC will be called by the CO to delete a snapshot.
+If successful, the storage space associated with the snapshot MUST be released and all the data in the snapshot SHALL NOT be accessible anymore.
```
jdef: "... storage space associated with the snapshot MUST be released..." - this probably depends on how the storage backend reclaims storage space. is there a reason that we feel the need to declare this in the spec? for example, some backend systems might tombstone a snapshot, but it might not actually be deleted until some reaper (or compactor, or whatever) comes along and removes the data.

I see that we have similar language for `DeleteVolume`, perhaps we should create a ticket and address this language in a follow-up PR instead of in this one.

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+| Call not implemented | 12 UNIMPLEMENTED | CreateSnapshot call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` to discover Plugin capabilities. |
+| Not enough space to create snapshot | 13 RESOURCE_EXHAUSTED | There is not enough space on the storage system to handle the create snapshot request. | Caller should fail this request. Future calls to CreateSnapshot may succeed if space is freed up. |
+
+
+#### `DeleteSnapshot`
+
+A Controller Plugin MUST implement this RPC call if it has CREATE_DELETE_SNAPSHOT capability.
```
jdef: wrap in backticks? `CREATE_DELETE_SNAPSHOT`

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+This RPC will be called by the CO to delete a snapshot.
+If successful, the storage space associated with the snapshot MUST be released and all the data in the snapshot SHALL NOT be accessible anymore.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot_id does not exist or the artifacts associated with the snapshot do not exist anymore, the Plugin MUST reply `0 OK`.
```
jdef: backticks? `snapshot_id`

spec.md
```diff
@@ -1090,6 +1084,252 @@ message ControllerServiceCapability {
...
+| Operation pending for snapshot | 10 ABORTED | Indicates that there is already an operation pending for the specified snapshot. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per snapshot at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same snapshot. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified snapshot, and then retry with exponential back off. |
+| Call not implemented | 12 UNIMPLEMENTED | DeleteSnapshot call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` to discover Plugin capabilities. |
+
+
+#### `ListSnapshots`
+
+A Controller Plugin MUST implement this RPC call if it has LIST_SNAPSHOTS capability.
```
jdef: backticks? `LIST_SNAPSHOTS`

spec.md
jdef: For volumes, I think it we made it an entry in case we wanted to include additional information later on, like the currently attached node or something.

spec.md
xing-yang: I moved this info to a common place to address comments from @julian-hj.  He said the comments are getting very lengthy for each field and it's better to put the comments in a common place.  I'm fine doing it either way.

spec.md
xing-yang: Sure.

spec.md
xing-yang: Ok.

spec.md
xing-yang: Ok.

spec.md
xing-yang: You are right.  This is not necessary.  I'll remove this one and change the spec accordingly.

spec.md
xing-yang: Actually I changed this to "disk size of the snapshot" based on the Google Cloud API doc.

From this doc (https://cloud.google.com/compute/docs/reference/rest/v1/disks/createSnapshot), the output of createSnapshot has two sizes:
snapshot.diskSizeGb - "Size of the snapshot in GB."
snapshot.storageBytes - "A size of the storage used by the snapshot. As snapshots share storage, this number is expected to change with snapshot creation/deletion."

From here (https://cloud.google.com/compute/docs/reference/rest/v1/disks), there is a sizeGb in create volume:
sizeGb - "Size of the persistent disk, specified in GB. You can specify this field when creating a persistent disk using the sourceSnapshot parameter. The value of sizeGb must not be less than the size of the snapshot."

So our SnapshotSize is the same as snapshot.diskSizeGb - "Size of the snapshot in GB".  I can change this just to "size of the snapshot".

spec.md
xing-yang: Sure.

spec.md
xing-yang: Sure.

spec.md
xing-yang: I'll keep it this way to be consistent with ListVolumes.

spec.md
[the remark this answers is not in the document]
xing-yang: Sure.

spec.md
```diff
@@ -575,23 +597,28 @@ message CreateVolumeRequest {
...
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // Contains all attributes of the source that is used to create
+  // the volume from. This field is OPTIONAL.
+  VolumeContentSource content_source = 6;
+}
+
+// Specifies what source the volume will be created from. One of the
+// source_type fields MUST be specified. Only Snapshot is
```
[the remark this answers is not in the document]
xing-yang: Ok.

spec.md
```diff
@@ -1074,6 +1058,16 @@ message ControllerServiceCapability {
...
       LIST_VOLUMES = 3;
       GET_CAPACITY = 4;
+      CREATE_DELETE_SNAPSHOT = 5;
+      // For storage systems that need to upload the snapshot after
+      // it is cut, LIST_SNAPSHOTS is a required RPC because CO needs
+      // to call it to find out whether the upload is complete.
+      LIST_SNAPSHOTS = 6;
+      // Currently the only way to consume a snapshot is to create
+      // a volume from it. Therefore plugins supporting
+      // CREATE_DELETE_SNAPSHOT should also support
```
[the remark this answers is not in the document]
xing-yang: Sure.

spec.md
```diff
@@ -1074,6 +1058,16 @@ message ControllerServiceCapability {
       PUBLISH_UNPUBLISH_VOLUME = 2;
       LIST_VOLUMES = 3;
       GET_CAPACITY = 4;
+      CREATE_DELETE_SNAPSHOT = 5;
+      // For storage systems that need to upload the snapshot after
+      // it is cut, LIST_SNAPSHOTS is a required RPC because CO needs
```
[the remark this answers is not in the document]
xing-yang: Ok

spec.md
```diff
@@ -379,6 +390,17 @@ This string MAY be surfaced by CO to end users.
 
 The status `details` MUST be empty. In the future, this spec may require `details` to return a machine-parsable protobuf message if the status `code` is not `OK` to enable CO's to implement smarter error handling and fault resolution.
 
+### Secrets Requirements
+
+Secrets may be required by plugin to complete a RPC request.
+A secret is a string to string map where the key identifies the name of the secret (e.g. "username" or "password"), and the value contains the secret data (e.g. "bob" or "abc123").
+Each key MUST consist of alphanumeric characters, '-', '_' or '.'.
+Each value MUST contain a valid string. An SP MAY choose to accept binary (non-string) data by using a binary-to-text encoding scheme, like base64.
```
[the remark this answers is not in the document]
xing-yang: You are right.  I'll split.

spec.md
xing-yang: What I meant was that ListSnapshots should not wait for any snapshot to be created or deleted and will just return the snapshots it can find currently, but you are right that this could still take a long time if there are a large number of snapshots in the storage system.  I'll reword this.

spec.md
xing-yang: Yes, I added this because it is also in DeleteVolume.  I'll create a ticket.

spec.md
xing-yang: Ok.

spec.md
xing-yang: Sure.

spec.md
xing-yang: Sure.

spec.md
xing-yang: sure.

spec.md
xing-yang: We had some discussions on this earlier and decided to make SnapshotSize a message to differentiate between a size of 0 and null (for storage systems that don't know the size).  Let me add a note for this.

spec.md
xing-yang: Sure.

spec.md
xing-yang: I think this is not a common case to support create snapshot from snapshot, so I'd prefer to add it later if needed.

spec.md
xing-yang: I'll simplify the comments.

spec.md
xing-yang: Sure.

jdef:
In csi, capacity is specified in bytes, not mb, gb, tb, etc. I'd like csi
to use consistent units where possible.

jdef:
Well Known Types are used for the same reason. Which is why I raised the
question. WKTs are common message types vs the custom type that we defined.

xing-yang:
@jdef I see.  I'll use Int64Value then.  Regarding capacity, I am aware that volume uses bytes.  I actually specified "bytes" in the spec earlier but got a review comment asking me to remove it.  So right now the spec just says "size of snapshot" without specifying a unit.

jdef:
It might take a little elbow grease to use Int64Value given that way that
WKT are kind of half-integrated into protobuf. If it's easier to do the
Int64Value refactor in a follow-up PR then that's fine, I don't want to
hold this PR up for it.

Regarding bytes: the CSI spec should be consistent. If we're not specifying
bytes for this field, then maybe we should have a general
disclaimer/statement earlier in the spec that says something to effect of
"unless otherwise stated, all fields pertaining to snapshot and/or volume
capacity are quantified in units of bytes".

spec.md
xing-yang: I'll change UnixNano to Unix nanoseconds which can be retrieved by other language as well.

xing-yang:
@jdef Ok, I've added "bytes" in the comments for the size field.  I'll address Int64Value in a followup PR.

xing-yang:
@jdef @saad-ali I have addressed your comments.  Please take look.  Thanks!

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+  // This field is REQUIRED.
+  string source_volume_id = 1;
+
+  // The suggested name for the snapshot. This field is REQUIRED for
+  // idempotency.
+  string name = 2;
+
+  // Secrets required by plugin to complete snapshot creation request.
+  // This field is OPTIONAL.
+  map<string, string> create_snapshot_secrets = 3;
```
dims: guess it's too late to use `map<string, bytes>` like we do in v1.Secret? ( https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/generated.proto#L3654 )

csi.proto
```diff
@@ -503,6 +485,147 @@ message ControllerServiceCapability {
...
   }
 }
+message CreateSnapshotRequest {
+  // The ID of the source volume to be snapshotted.
+  // This field is REQUIRED.
+  string source_volume_id = 1;
+
+  // The suggested name for the snapshot. This field is REQUIRED for
+  // idempotency.
+  string name = 2;
```
dims: So here in CreateSnapshotRequest, we are REQUIRED to pass a `name`. and this better be a fresh one or the snapshot does not get created. Right? 

If that's the case, we should be able to query the `name`(s) we used before .... However, `ListSnapshotsResponse` has a list of `Snapshot` object, BUT these objects do not have the `name` that was used when they were created.

So how do we keep track of the names that were already used?

spec.md
xing-yang: The secret is used in many places in the CSI spec and it has always been map<string, string>.  I don't know the reasoning behind map<string, string> vs map<string, bytes>.  I think you can open an issue with the request to change it to map<string, bytes>.

csi.proto
xing-yang: @dims The name in CreateSnapshotRequest works the same way as the name in CreateVolumeRequest.  In create volume, name is unique and it will be saved in PV in K8S so we can find a mapping between the volume name and the id generated by the plugin in the PV source.  CreateSnapshot will work the same way.  Snapshot will get an ID after it is created by the plugin and when we call ListSnapshots we pass in the snapshot ID to find it.

dims: @xing-yang the code that created the snapshot is the ONLY one who knows which name is mapped to which ID. in create volumes like you mentioned anyone can look in the PV to find the name and ID. that's not the case here

xing-yang: @dims For snapshot, we have VolumeSnapshot which is equivalent to PVC and VolumeSnapshotData which is equivalent to PV.  We also have data source for snapshot, similar to the data source for PV.  However those design info will be in the design proposal for K8S, not here:).  This is a CSI spec that involves other CO's in addition to K8S.

xing-yang: @dims I heard you.  Can you take a look of the "RPC Interactions" section?  It explains that name may be different from resource id.  Maybe you'll change your mind after reading that:)?

xing-yang: From the "RPC Interactions" section, you can see that it was a design decision that the CO-generated volume `name` may be different from the plugin-generated `volume_id`.  This PR for snapshot is consistent with that design.

spec.md
```diff
@@ -1103,6 +1313,46 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
 
 It is NOT REQUIRED for a controller plugin to implement the `LIST_VOLUMES` capability if it supports the `CREATE_DELETE_VOLUME` capability: the onus is upon the CO to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
 
+##### `CreateSnapshot`, `DeleteSnapshot`, `ListSnapshots`
+
+The plugin-generated `snapshot_id` is a REQUIRED field for the `DeleteSnapshot` RPC, as opposed to the CO-generated snapshot `name` that is REQUIRED for the `CreateSnapshot` RPC: these fields MAY NOT contain the same value.
```
dims: based on discussion above. may be we should make it clear

```
CO-generated snapshot `name` specified in CreateSnapshot cannot be looked up by `ListSnapshots`. It is a write only field and is not used anywhere else.
```

xing-yang: Sure.  I added some text to clarify this.  Thanks!

spec.md
xing-yang: Sound good!  I changed it in the latest PR.

spec.md
jdef: I actually like having them here, honestly.

spec.md
```diff
@@ -575,23 +598,29 @@ message CreateVolumeRequest {
...
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // If specified, the new volume will be pre-populated with data from
+  // this source. This field is OPTIONAL.
+  VolumeContentSource content_source = 6;
+}
+
+// Specifies what source the volume will be created from. One of the
+// source_type fields MUST be specified.
```
jdef: nit: this field is now called `type`, right?

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+}
+
+// The information about a provisioned snapshot.
+message Snapshot {
+  // This is the complete size of the snapshot in bytes. The purpose of
+  // this field is to give CO guidance on how much space is needed to
+  // create a volume from this snapshot. The size of the volume must not
+  // be less than the size of the source snapshot. This field is
+  // OPTIONAL. If this field is not set, it indicates that this size is
+  // unknown. The value of this field MUST NOT be negative.
```
jdef: is zero a valid snapshot size? if so, what does it mean?

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+  // the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. The format of this field should be a Unix Nanoseconds time
```
jdef: nit: why is Nanoseconds capitalized?

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+     UNKNOWN = 0;
+     // A snapshot is ready for use.
+     READY = 1;
+     // A snapshot is cut and is now being uploaded.
+     // Some cloud providers and storage systems uploads the snapshot
+     // to the cloud after the snapshot is cut. During this phase,
+     // `thaw` can be done so the application can be running again if
+     // `freeze` was done before taking the snapshot.
+     UPLOADING = 2;
+     // A snapshot is in error status.
```
jdef: is this a terminal status? as in, the snapshot can only be deleted from here?

spec.md
```diff
@@ -1103,6 +1313,48 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
...
+
+It is NOT REQUIRED for a controller plugin to implement the `LIST_SNAPSHOTS` capability if it supports the CREATE_DELETE_SNAPSHOT capability: the CO needs to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
+
+##### Snapshot Statuses
+
+A snapshot could have the following statusus: UPLOADING, READY, and ERROR.
+
+Some cloud providers will upload the snapshot to a location in the cloud (i.e., an object store) after the snapshot is cut.
+Uploading may be a long process that could take hours.
+If a `freeze` operation was done on the application before taking the snapshot, it could be a long time before the application can be running again if we wait until the upload is complete to `thaw` the application.
```
jdef: this is good information. the organization feels a bit clumsy. i might take a stab at crafting a PR against this branch to reorganize this section for better "flow"

spec.md
```diff
@@ -1103,6 +1313,48 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
...
+`thaw` can be done when the status is `READY` in this case.
+
+A `CREATING` status is not included here because CreateSnapshot is synchronous and will block until the snapshot is cut.
+
+The SnapshotStatus parameter provides guidance to the CO on what action can be taken in the process of snapshotting.
+Based on this information, CO can issue repeated (idemponent) calls to CreateSnapshot, monitor the response, and make decisions.
+Note that CreateSnapshot is a synchronous call and it must block until the snapshot is cut.
+If the cloud provider or storage system does not need to upload the snapshot after it is cut, the status returned by CreateSnapshot SHALL be `READY`.
+If the cloud provider or storage system needs to upload the snapshot after the snapshot is cut, the status returned by CreateSnapshot SHALL be `UPLOADING`.
+CO can continue to call CreateSnapshot while waiting for the upload to complete until the status becomes `READY`.
```
jdef: suggest: The CO MAY continue to call...

spec.md
jdef: https://github.com/container-storage-interface/spec/issues/225

spec.md
jieyu: + @saad-ali. IIRC, the rationale of using `map<string, string>` is to optimize the common use case where the secret is actually a string (e.g., passwd). For binary secrets, you can always encode it using base64.

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+  // relevant to the CO along with information required by the Plugin
+  // to uniquely identify the snapshot. This field is REQUIRED.
+  Snapshot snapshot = 1;
+}
+
+// The information about a provisioned snapshot.
+message Snapshot {
+  // This is the complete size of the snapshot in bytes. The purpose of
+  // this field is to give CO guidance on how much space is needed to
+  // create a volume from this snapshot. The size of the volume must not
```
jieyu: nits: Please use `MUST NOT` here.

spec.md
jieyu: nits: Given that this field is of type `SnapshotSize`, please move the comment about negative or not to `SnapshotSize` message.

spec.md
```diff
@@ -1090,6 +1086,220 @@ message ControllerServiceCapability {
...
+
+  // Identity information for the source volume. Note that create
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. The format of this field should be a Unix Nanoseconds time
+  // encoded as an int64. This field is REQUIRED.
+  int64 created_at = 4;
```
jieyu: Can you specify what will be the specification for the timestamp? Is there an RFC that we can reference?

spec.md
xing-yang: Yes!  Fixed.

spec.md
xing-yang: In the earlier version of this spec, we were also considering how much space is used by a snapshot so the used space is variable and can be zero in the beginning.  To differentiate between the case when the snapshot space is not used vs the case when the storage system does not know the size, we decided to use a message to define that.

Now that we have removed the variable part and this snapshot size should be greater than zero if specified.  I wonder if we should just use int64 for the size now and a size of 0 indicates the storage system does not know the size?  That will be consistent with the volume size.

I'll move the comment close to the SnapshotSize message if we are keeping it.

spec.md
xing-yang: I'll make it lower case.

spec.md
xing-yang: If a snapshot ends up in error status, the user will need to find out what has caused the error condition and then determines what to do with it.  If an error happens before the snapshot is cut, it needs to be deleted.  If an error happens during the upload, may be the cloud admin has a way to fix the problem and still allow the snapshot to be used.

spec.md
xing-yang: Sure.  Thanks:).

spec.md

spec.md
```diff
@@ -575,23 +598,29 @@ message CreateVolumeRequest {
...
-  // all unique secrets that an SP expects must be unique across all CSI
-  // operations.
-  // This information is sensitive and MUST be treated as such (not
-  // logged, etc.) by the CO.
   // This field is OPTIONAL.
   map<string, string> controller_create_secrets = 5;
+
+  // If specified, the new volume will be pre-populated with data from
+  // this source. This field is OPTIONAL.
+  VolumeContentSource content_source = 6;
```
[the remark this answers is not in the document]
xing-yang: Sure.  Changed.

spec.md
```diff
@@ -1074,6 +1061,15 @@ message ControllerServiceCapability {
       PUBLISH_UNPUBLISH_VOLUME = 2;
       LIST_VOLUMES = 3;
       GET_CAPACITY = 4;
+      // Currently the only way to consume a snapshot is to create
+      // a volume from it. Therefore plugins supporting
+      // CREATE_DELETE_SNAPSHOT SHOULD support creating volume from
+      // snapshot.
+      CREATE_DELETE_SNAPSHOT = 5;
+      // For storage systems that need to upload the snapshot after
+      // it is cut, LIST_SNAPSHOTS is a REQUIRED RPC because CO needs
```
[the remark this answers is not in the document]
xing-yang: Sure. Reworded a little bit.  Let me know if it looks better.

spec.md

spec.md
xing-yang: On Unix, if you enter “date +%s%N”, it returns the current time in nanoseconds since 1970-01-01 00:00:00 UTC.
$ date +%s%N
1525619388433753983
In Go, it is time.Now().UnixNano().

xing-yang:
@jieyu @jdef @saad-ali I have addressed your comments.  Can you please take a look again?  Thanks!

By the way, does anyone know how to re-trigger the CI?

jdef:
Re-triggering CI won't help, this needs #227 to land, and then to rebase

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
...
+  // field is REQUIRED.
+  int64 created_at = 4;
+
+  // The status of a snapshot.
+  SnapshotStatus status = 5;
+}
+
+message SnapshotSize {
+  // The complete size of the snapshot in bytes.
+  // The value of this field MUST NOT be negative.
```
jdef: It's still unclear to me what a value of "zero" means here. I think we need to either (a) disallow zero; (b) explain the meaning of zero, or else; (c) re-think using a message type for `SnapshotSize` and instead make zero == unspecified.

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
...
+#### `CreateSnapshot`
+
+A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` controller capability.
+This RPC will be called by the CO to create a new snapshot from a source volume on behalf of a user.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot `name` already exists and is compatible with the specified `source_volume_id` and `parameters` in the `CreateSnapshotRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateSnapshotResponse`.
+
+A snapshot MAY be used as the source to provision a new volume.
+CreateVolumeRequest is modified to take an OPTIONAL parameter of a source snapshot.
```
jdef: nit: "is modified" only makes sense in the context of reviewing this PR. Once the PR lands, it sounds funny. Suggest instead: "A `CreateVolumeRequest` message may specify an OPTIONAL source snapshot parameter."

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
...
+Reverting snapshot, where data in the original volume is erased and replaced with data in the snapshot, is an advanced functionality not every storage system can support and therefore is currently out of scope.
```
jdef: suggest: "Reverting a snapshot..."

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
...
+  // unknown.
+  SnapshotSize size = 1;
+
+  // Uniquely identifies a snapshot and is generated by the plugin. It
+  // will not change over time. This field is REQUIRED. The identity
+  // information will be used by the CO in subsequent calls to refer to
+  // the provisioned snapshot.
+  string id = 2;
+
+  // Identity information for the source volume. Note that create
```
jdef: nit: s/create snapshot from snapshot/creating a snapshot from a snapshot/

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
...
+  // snapshot from snapshot is not supported here so the source has to
+  // be a volume. This field is REQUIRED.
+  string source_volume_id = 3;
+
+  // Timestamp when the point-in-time snapshot is taken on the storage
+  // system. The format of this field should be a Unix nanoseconds time
+  // encoded as an int64. On Unix, the command `date +%s%N` returns the
```
jdef: thanks for this clarity

spec.md
jdef: is the user the only actor that can initiate recovery here? for example, are there ways that a CO could gracefully recovery from an ERROR condition and if so, what are they? (read: please document them if there are any)

spec.md
```diff
@@ -1090,6 +1088,223 @@ message ControllerServiceCapability {
 
 If the plugin is unable to complete the ControllerGetCapabilities call successfully, it MUST return a non-ok gRPC code in the gRPC status.
 
+#### `CreateSnapshot`
+
+A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` controller capability.
+This RPC will be called by the CO to create a new snapshot from a source volume on behalf of a user.
+
+This operation MUST be idempotent.
+If a snapshot corresponding to the specified snapshot `name` already exists and is compatible with the specified `source_volume_id` and `parameters` in the `CreateSnapshotRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateSnapshotResponse`.
```
jdef: This is an important sentence: if the snapshot "exists" and the snapshot status is `ERROR` this RPC is still expected to return `OK`. Maybe we could make that more clear in the errors table that follows?

spec.md
jdef: this is a bit misleading. the text "... MAY NOT contain the same value." means this:

(a) the returned `snapshot_id` might be the same as the user-generated snapshot `name`
(b) .. or it could be completely different

to say that the ListSnapshots can't be filtered by the user-generated `name` only applies for snapshots that generate their own `snapshot_id` that's guaranteed to never match user-generated `name`. The sentence that was added below (`SHALL NOT be used...`) to resolve complaints is semantically inconsistent - please rewrite, or else remove it.

jdef:
@xing-yang please stop squashing your commits down to 1 - that can be done as a last step. it's harder for reviewers to see what's changed between revisions when everything is continuously squashed. thanks!

spec.md
xing-yang: I prefer option (c).  That will be consistent with what "zero" means in volume size.

spec.md
xing-yang: Ok

spec.md
xing-yang: Ok

spec.md
xing-yang: Ok

spec.md
xing-yang: I checked with Jing on the upload case.  If an error happens during upload, that is also a terminal status so the snapshot can be deleted.  So that is consistent with the error before the snapshot is cut.

spec.md
xing-yang: If the snapshot "exists" and the snapshot status becomes ERROR, this RPC should be returning a gRPC error code, depending on what the error is.  Currently we have the following error codes specified: 6 ALREADY_EXISTS; 10 ABORTED (Operation pending for snapshot); 12 UNIMPLEMENTED; and 13 RESOURCE_EXHAUSTED.

xing-yang: The sentence you are referring to only applies if the snapshot has been successfully cut and uploaded (if upload is part of the process).

spec.md
```diff
@@ -1103,6 +1313,46 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
...
+If a `CreateSnapshot` operation times out, leaving the CO without an ID with which to reference a snapshot, and the CO also decides that it no longer needs/wants the snapshot in question then the CO MAY choose one of the following paths:
+
+1. Execute the `ListSnapshots` RPC to possibly obtain a snapshot ID that may be used to execute a `DeleteSnapshot` RPC; upon success execute `DeleteSnapshot`.
+2. The CO takes no further action regarding the timed out RPC, a snapshot is possibly leaked and the operator/user is expected to clean up.
+
+It is NOT REQUIRED for a controller plugin to implement the `LIST_SNAPSHOTS` capability if it supports the CREATE_DELETE_SNAPSHOT capability: the CO needs to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
+
+##### Snapshot Statuses
+
+A snapshot could have the following statusus: UPLOADING, READY, and ERROR.
```
jdef: please add a sentence or two in this section that describes `ERROR` as a terminal snapshot status, and that a CO SHOULD explicitly *delete* snapshots in this state.

xing-yang: Sure.  Added.

spec.md
jdef: @saad-ali would it help if the secrets field in the protos referred back to this section for requirements?

xing-yang:
Merge conflicts are resolved now.

spec.md
```diff
@@ -1090,6 +1088,221 @@ message ControllerServiceCapability {
...
+  // This field is REQUIRED.
+  string source_volume_id = 1;
+
+  // The suggested name for the snapshot. This field is REQUIRED for
+  // idempotency.
+  string name = 2;
+
+  // Secrets required by plugin to complete snapshot creation request.
+  // This field is OPTIONAL.
+  map<string, string> create_snapshot_secrets = 3;
```
jdef: nit: wondering if we can come up with a better name for the secrets field here. `create_snapshot_secrets` stutters with the message type name. maybe `controller_create_secrets`?

spec.md
```diff
@@ -1090,6 +1088,221 @@ message ControllerServiceCapability {
...
+
+```protobuf
+message DeleteSnapshotRequest {
+  // The ID of the snapshot to be deleted.
+  // This field is REQUIRED.
+  string snapshot_id = 1;
+
+  // Secrets required by plugin to complete snapshot deletion request.
+  // This field is OPTIONAL.
+  map<string, string> delete_snapshot_secrets = 2;
```
jdef: nit: ditto re: a better name for the secrets field. maybe `controller_delete_secrets`?

spec.md
```diff
@@ -1103,6 +1316,49 @@ If a `CreateVolume` operation times out, leaving the CO without an ID with which
...
 
+##### `CreateSnapshot`, `DeleteSnapshot`, `ListSnapshots`
+
+The plugin-generated `snapshot_id` is a REQUIRED field for the `DeleteSnapshot` RPC, as opposed to the CO-generated snapshot `name` that is REQUIRED for the `CreateSnapshot` RPC.
+If a `CreateSnapshot` operation times out, leaving the CO without an ID with which to reference a snapshot, and the CO also decides that it no longer needs/wants the snapshot in question then the CO MAY choose one of the following paths:
+
+1. Execute the `ListSnapshots` RPC to possibly obtain a snapshot ID that may be used to execute a `DeleteSnapshot` RPC; upon success execute `DeleteSnapshot`.
+2. The CO takes no further action regarding the timed out RPC, a snapshot is possibly leaked and the operator/user is expected to clean up.
+
+It is NOT REQUIRED for a controller plugin to implement the `LIST_SNAPSHOTS` capability if it supports the CREATE_DELETE_SNAPSHOT capability: the CO needs to take into consideration the full range of plugin capabilities before deciding how to proceed in the above scenario.
```
jdef: I'm thinking about this statement more. If a plugin *does not* support LIST_SNAPSHOTS and also implements an "uploading" step, then .. in the event that an upload fails, there's no way for a CO to obtain a `snapshot_id` because the `CreateSnapshot` call with return a gRPC error (instead of a Snapshot w/ a SnapshotStatus). Is that acceptable?

spec.md
```diff
@@ -1090,6 +1088,221 @@ message ControllerServiceCapability {
...
+     // Some cloud providers and storage systems uploads the snapshot
+     // to the cloud after the snapshot is cut. During this phase,
+     // `thaw` can be done so the application can be running again if
+     // `freeze` was done before taking the snapshot.
+     UPLOADING = 2;
+     // A snapshot is in error status.
+     ERROR = 3;
+  }
+  Type type = 1;
+}
```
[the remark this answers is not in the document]
xing-yang: I think having a "message" or "details" string would be very helpful for debugging.  I'll add that.

spec.md
xing-yang: I think someone (maybe Jie) suggested naming this field this way (as secrets for creating a snapshot).  If you and him can agree with a new way of naming, I'll be happy to change it:).

spec.md
xing-yang: In this case, CreateSnapshot can return a Snapshot with a Error status and a gRPC error code.

spec.md
xing-yang: Same as above.  I named it this way because I got an earlier comment to call this one as <operation>_snapshot_secrets.

spec.md
```diff
@@ -1183,6 +1191,10 @@ message SnapshotStatus {
      ERROR = 3;
   }
   Type type = 1;
```
jdef: suggest: // This field is REQUIRED.

spec.md
```diff
@@ -1183,6 +1191,10 @@ message SnapshotStatus {
...
+
+  // Additional information to describe why a snapshot ended up in the
+  // `ERROR` status.
```
jdef: suggest: This field is OPTIONAL.

xing-yang: Done.

spec.md
xing-yang: Done.

spec.md
saad-ali: @jdef That's a fair compromise. I would like the proto file to be self-documenting. If there is a requirement for some field, I want that to be quickly apparent from reading the comment rather then require user to go read the entire document. If we can have a pointer comment saying "see section x of spec for requirements on y", that should be fine -- since the reader will be aware that there are additional requirements that apply.

saad-ali:
@xing-yang We spoke offline, please ping me when ready. I'd like to get this merged.

xing-yang:
Rebased.

xing-yang:
Hi @saad-ali, I have rebased and addressed all the comments.  Please take a look.  Thanks!

xing-yang:
@jdef @saad-ali I changed the status `ERROR` to `ERROR_UPLOADING` and added some clarifications based on what we discussed.  Please take a look.  Thanks!

saad-ali:
LGTM, and merging!!

xing-yang:
Awesome!  Thanks everyone for reviewing the PR and helping us improve the design!!!

PR #227, opened 2018-05-01
Makefile fixes, bump golang protobuf

### LINKED ISSUE #226: Results from make command are not reproducible
I think because make uses `csi.proto.tmp` file and lib files get regenerated from tmp file, the result of make command is not reproducible. Also `tmp` file is gitignored and hence one branch may silently reuse result from another branch making authoring process harder.

cc @jdef @saad-ali

jdef:
fixes #226 

The upstream golang/protobuf project just landed, onto master, a ton of changes from their development branch and it significantly changes the generated bindings. This PR re-generates the bindings using the latest golang protoc plugin. Also included, a fix for the Makefile's so that running `make clobber` works as expected (and allows one to easily regenerate bindings locally, using the latest binaries).


jdef:
PTAL @cpuguy83 @julian-hj @saad-ali @jieyu

PR #228, opened 2018-05-10
Probe also reports readiness, and other cleanup

jdef:
* consistently document empty messages
* fix spelling mistakes (thanks, `aspell`)
* `ProbeResponse` returns additional `readiness` field to distinguish between "I'm not healthy" and "I'm healthy, but I'm not yet done initializing".

Alternatives considered:

1. No API changes, `Probe` blocks until plugin has completed initialization.
2. Return a unique gRPC error code that means "I'm still initializing".

Consequences of (1): The CO is unable to differentiate between a timeout (e.g. underlying I/O issue) vs. initialization-in-progress. The CO may decide that a `Probe` which repeatedly times out indicates an unhealthy plugin and mistakenly decide to restart the plugin, aborting the ongoing initialization process.

Consequences of (2): returning non-OK gRPC status code for a health plugin violates the contract of Probe and is not backwards compatible. Probe returns a non-OK error code when it is NOT healthy. While a plugin is initializing it MAY be perfectly healthy and SHOULD return an OK status code in such case.

Well known type (WKT) `BoolValue` vs `bool` vs simple, 1-value `Readiness` message:

1. `bool readiness` has a default value of `false`, which is not backwards compatible with the existing `Probe` RPC. CSI v0.3 is aiming for full backwards compatibility with plugins written against the v0.2 spec.

2. A simple, 1-value `Readiness` message further clutters the API with a simple type that will only ever have a single bool value. WKT's are neatly integrated into protobuf and this case seems like a great candidate for such use.


jdef:
@julian-hj @jieyu @saad-ali @cpuguy83 PTAL

cpuguy83:
I like the idea, but 👎 (but usually cringe at these)  on the bool for readiness. Would prefer something more descriptive... a `oneof` on state might work.

Note that it feels like returning an OK but not ready is also a breaking change.

jdef:
@cpuguy83 if we used an `enum` to capture "readiness" (or whatever we want to call it) then we'd need to wrap the `enum` in a `message` type because this is an OPTIONAL field for the `ProbeResponse` message. It might look like this:

```protobuf
message ProbeResponse {
  message Condition 
    enum Type {
      UNKNOWN = 0;
      PENDING = 1; // plugin is healthy, but not all RPCs are online yet
      READY = 2;   // plugin is healthy, all RPCs are online
    }
    Type type = 1; // REQUIRED
  }
  Condition condition = 1; // OPTIONAL, assume READY if not present
}
```

personally, this feels a bit over-engineered. but I'm flexible.

jdef:
... or if you wanted something more k8s-style:
```protobuf
message ProbeResponse {
  message Condition 
    enum Type {
      UNKNOWN = 0;
      READINESS = 1;
    }
    Type type = 1; // REQUIRED
    bool value = 2; // REQUIRED
    string message = 3; // OPTIONAL
  }
  Condition condition = 1; // OPTIONAL, assume "healthy and ready" if not present
}
```

gpaul:
I believe these are the relevant states to represent:

1. Plugin reports health or failure unambiguously. This is the trivial case.

2. Plugin confidently reports "wait a bit longer", e.g., it is running `fsck` on a bunch of volumes but has reason to believe this process will terminate and it will become healthy and ready.

3. Plugin reports "waiting" but cannot know that it expects to transition to being healthy, e.g., it is waiting on requests to an SP that is unresponsive. This is a candidate for a separate `WaitingOnSP` state. Does the `readiness` bool handle this case? I believe it does: the plugin should follow well-defined timeout and retry logic particular to its SP. During such a scenario it reports `readiness=false` but once its retry strategy is exhausted without success it should transition to a failed state.

Until we think of more states, I propose we stick to introducing the bool?

csi.proto
```diff
@@ -112,10 +116,31 @@ message PluginCapability {
...
+  //    initialization state.
+  // 3) The plugin has finished initializing and is ready to service
+  //    calls to its Controller and/or Node services. A successful
+  //    response is returned with a readiness value of `true`.
+  //
+  // This field is OPTIONAL. If not present, the caller SHALL assume
+  // that the plugin is in a ready state and is accepting calls to its
+  // Controller and/or Node services (according to the plugin's reported
+  // capabilities).
+  .google.protobuf.BoolValue readiness = 1;
```
gpaul: Could we use a different field name? The word `readiness` does not sounds like it refers to a boolean value. Perhaps `ready` or my personal favourite: `initializating`.

cpuguy83:
Bool works for me.

csi.proto
jdef: done

jdef:
rebased to master

jdef:
@cpuguy83 is that an official LGTM?

cpuguy83:
Yes, LGTM.

PR #232, opened 2018-05-11
spec: clarify field requirements

### LINKED ISSUE #185: Response Message Validation: the Problem & a Solution
Hi all,

As I was working on some CSI-related middleware, a problem occurred to me: is it ever safe for an SP to validate CSI response messages against the specification and return a nil response message with an error if the message is invalid?

The answer, I believe, is **no**.

# The Problem
As I outlined in thecodeteam/gocsi#81:

> An SP is capable of making changes to an underlying storage platform. The results of such changes are reflected in CSI response messages. Dropping these messages could render a client unable to acknowledge a change has occurred, and could potentially lead to further errors, such as data loss. 

Prior to the aforementioned PR, GoCSI would validate SP response messages against the CSI specification, following the same rules as applied to incoming request messages. Some examples, include, but are not limited to, the following:

* Strings longer than `128` bytes would cause an error code of `Internal` to be returned with the details about the field and its size.
* A `GetPluginInfoResponse.Name` field that does not adhere to reverse-domain name notation causes an error to be returned.
* Fields that have non-nil, empty map values cause a response to be marked in error.

I believe the GoCSI spec validation middleware is quite useful for the development of an SP as it lets us know early in the process if request **and** response data is valid. However, as stated at the top of this issue, the process of dropping a response to return an error due to a response's validity against the spec is not a thing SPs should ever do.

# A Solution
Please note the title of this section is ***A** Solution*, not ***The** Solution*. That is because what is outlined below is **an** approach, but certainly not the only one.

The PR thecodeteam/gocsi#81 changes the way GoCSI's spec validation middleware handles response messages. Instead of dropping the response message and returning an error, the response message is now [marshalled](https://github.com/thecodeteam/gocsi/blob/abda6ec9c37dc2e02132d7d6ec33d784e4b3d758/middleware/specvalidator/spec_validator.go#L246-L279) into the gRPC error details:

```go
if s.opts.repValidation {
	// Validate the response against the CSI specification.
	if err := s.validateResponse(ctx, method, rep); err != nil {

		// If an error occurred while validating the response, it is
		// imperative the response not be discarded as it could be
		// important to the client.
		st, ok := status.FromError(err)
		if !ok {
			st = status.New(codes.Internal, err.Error())
		}

		// Add the response to the error details.
		st, err2 := st.WithDetails(rep.(proto.Message))

		// If there is a problem encoding the response into the
		// protobuf details then err on the side of caution, log
		// the encoding error, validation error, and return the
		// original response.
		if err2 != nil {
			log.WithFields(map[string]interface{}{
				"encErr": err2,
				"valErr": err,
			}).Error("failed to encode error details; " +
				"returning invalid response")

			return rep, nil
		}

		// There was no issue encoding the response, so return
		// the gRPC status error with the error message and payload.
		return nil, st.Err()
	}
}
```

The test `With Invalid Plug-in Name Error` in [`./testing/identity_test.go`](https://github.com/thecodeteam/gocsi/blob/abda6ec9c37dc2e02132d7d6ec33d784e4b3d758/testing/identity_test.go#L110-L131) illustrates the error case of a `GetPluginInfoResponse.Name` field not conforming to the reverse-domain naming convention. The response is encoded into the error's details, and the error is returned. The linked example shows how a client may reconstitute the original response:

```go
st, ok := status.FromError(err)
Ω(ok).Should(BeTrue())
Ω(st.Details()).Should(HaveLen(1))
rep, ok := st.Details()[0].(*csi.GetPluginInfoResponse)
Ω(ok).Should(BeTrue())
Ω(rep.Name).Should(Equal("Mock"))
Ω(rep.VendorVersion).Should(Equal("v1.0.0"))
```

As demonstrated above, an error was returned from the SP, and the original response is still available if it is necessary to the client-side workflow.

# Conclusion
I believe it's imperative the CSI specification comment on how to handle response message validation. Either to state that response messages should **never** be dropped, or if they contain invalid data, to return a gRPC error and encode the original response message in the error details. What is important, however, is that response messages should **always** be returned to a client, even if the message data contains errors. It's simply too dangerous to not inform clients when the underlying storage platform has been modified.

# Action Items
In order to handle this issue, the following questions (for starters) must be answered:

* Should CSI response messages be validated against the spec with the same conventions as applied to request messages?
* What is the expected behavior of an SP if a response message contains invalid data?
* What error code should be used when a response message contains invalid data? For example, GoCSI's spec validation middleware currently uses `Internal` for invalid field data in response messages, as opposed to `InvalidArgument` for invalid field data request messages.

jdef:
* fixes #185


jdef:
@cpuguy83 @julian-hj @saad-ali @jieyu PTAL

julian-hj:
:+1:

Issue #206, opened 2018-02-27
Secrets field names


akutz:
# Field Name Length
I just double-checked the spec, and there is never more than a single type of secrets field per RPC. The field names are unnecessarily verbose and result in language bindings and subsequent developer code that is not as clean or succinct as it could be.

Please consider simplifying the secrets fields to `secrets`.

Also, if the argument against this is that RPCs may have multiple secrets packages in the future, then the fields should become more verbose as the future of CSI may also contain snapshots, pools, etc. Then `controller_create_secrets` is no longer unique unless it becomes `controller_create_volume_secrets`.

At the very minimum the field names should drop the service prefix as it's redundant.

However, again, these are not RPC message definitions. The scope of fields is tied to the RPC message itself, and thus information like the service prefix as necessarily unique is quite improbable. To that end, so is everything but `secrets` itself.

# Field Name Permanence
The names of the secrets-related fields should be declared untouchable once decided upon. This is due to their inherent, sensitive nature. Logging middleware and other applications may elect to omit sensitive data when storing or producing output. Changing the names of the fields in the future could result in the leaking of sensitive data.

# Summary
1. Please simplify the secrets field names to `secrets`.
2. Please never change these field names again

cc @saad-ali

saad-ali:
What @jdef mentioned is being handled in PR #233 and issue https://github.com/container-storage-interface/spec/issues/291?

Regarding the crazy secret names? Yes, we should clean them up and just call them `secret` and clarify in the field comments.

saad-ali:
Closed in https://github.com/container-storage-interface/spec/pull/317

PR #233, opened 2018-05-11
Sensitive field option for credential fields

jdef:
Custom protobuf option for fields that MAY contain sensitive, secret information. Allows for introspection of CSI message data using specific protobuf options, which is a more robust approach than the current alternative (testing field names).

* partial fix for #206


jdef:
rebased to master

jdef:
Jie was wondering how I came up with the extension number. I basically just picked a large number that I thought would not collide with anything else in the spec. Well, apparently there's a registry of extension numbers for public projects. So if we want this PR to land, we'll probably need to register an extension number.

See https://github.com/protocolbuffers/protobuf/blob/master/docs/options.md

jdef:
https://github.com/protocolbuffers/protobuf/pull/5319

jdef:
Once we are assigned an official extension range I can bump this PR to use a number from that range.

jdef:
rebased to master

saad-ali:
protocolbuffers/protobuf#5319 has merged. We should be good to use 1059. PR needs rebase.

PR #237, opened 2018-05-22
Snapshot clarifications

Akrog:
The new snapshot feature has a couple of things that I think should be addressed.

There is no reference to expected behavior when the required_bytes in _CreateVolumeRequest_'s _capacity_range_ is larger or smaller than that of the _VolumeContentSource_.  This may lead to different interpretations and implementations, so we should be explicit.

Comments in _SnapshotSource_ can be confusing since they are referencing a "created snapshot", when they should be referencing an existing snapshot instead.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -655,11 +655,10 @@ message CreateVolumeRequest {
 // type fields MUST be specified.
 message VolumeContentSource {
   message SnapshotSource {
-    // Contains identity information for the created snapshot. This
-    // field is REQUIRED. The identity information will be used by
-    // the CO in subsequent calls to refer to the provisioned snapshot.
-    // Plugin is REQUIRED to support creating volume from snapshot if
-    // it supports the capability CREATE_DELETE_SNAPSHOT.
+    // Contains identity information for the existing source snapshot.
+    // This field is REQUIRED. Plugin is REQUIRED to support creating
+    // volume from snapshot if it supports the capability
+    // CREATE_DELETE_SNAPSHOT.
     string id = 1;
   }
 
--- a/spec.md
+++ b/spec.md
@@ -623,7 +623,11 @@ message CreateVolumeRequest {
 
   // This field is OPTIONAL. This allows the CO to specify the capacity
   // requirement of the volume to be provisioned. If not specified, the
-  // Plugin MAY choose an implementation-defined capacity range.
+  // Plugin MAY choose an implementation-defined capacity range. If
+  // specified it MUST always be honored, even when creating volumes
+  // from a source; which may force some backends to internally extend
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -615,7 +615,11 @@ message CreateVolumeRequest {
 
   // This field is OPTIONAL. This allows the CO to specify the capacity
   // requirement of the volume to be provisioned. If not specified, the
-  // Plugin MAY choose an implementation-defined capacity range.
+  // Plugin MAY choose an implementation-defined capacity range. If
+  // specified it must always be honored, even when creating volumes
```
jdef: nit: s/must/MUST/ ... to make it an official requirement

spec.md
```diff
@@ -783,7 +786,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
-| Unsupported `capacity_range` | 11 OUT_OF_RANGE | Indicates that the capacity range is not allowed by the Plugin. More human-readable information MAY be provided in the gRPC `status.message` field. | Caller MUST fix the capacity range before retrying. |
+| Unsupported `capacity_range` | 11 OUT_OF_RANGE | Indicates that the capacity range is not allowed by the Plugin (for example when trying to create a volume smaller than the source snapshot). More human-readable information MAY be provided in the gRPC `status.message` field. | Caller MUST fix the capacity range before retrying. |
```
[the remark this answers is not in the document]
Akrog: That was my initial writing, but then I read in the row above `the CO MAY lose state (for example when the CO crashes and restarts),` and decided to maintain the style.  I'm fine either way.

I'm trying to figure out what's the problem with the csi.pb.go I generated in these patches, and why the Go job failed.  Once I have that figured out I'll submit a new patch.

xing-yang:
LGTM

spec.md
```diff
@@ -655,11 +659,10 @@ message CreateVolumeRequest {
 // type fields MUST be specified.
 message VolumeContentSource {
   message SnapshotSource {
-    // Contains identity information for the created snapshot. This
-    // field is REQUIRED. The identity information will be used by
-    // the CO in subsequent calls to refer to the provisioned snapshot.
-    // Plugin is REQUIRED to support creating volume from snapshot if
-    // it supports the capability CREATE_DELETE_SNAPSHOT.
+    // Contains identity information for the existing source snapshot.
+    // This field is REQUIRED. Plugin is REQUIRED to support creating
```
cpuguy83: Should this be `Plugin MUST support ...`?

cpuguy83: This is fine as is.

Akrog:
Thanks!

PR #238, opened 2018-05-22
Add a new RPC NodeDrain to the spec

Kaixiang:
Further to conversation in the 5/9/2018 community meeting.  We (OSS CF) agreed to submit a PR adding a NodeDrain RPC.

Thanks
CF


spec.md
```diff
@@ -1784,6 +1785,27 @@ Condition | gRPC Code | Description | Recovery Behavior
 | --- | --- | --- | --- |
 | Call not implemented | 12 UNIMPLEMENTED | NodeGetInfo call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` or `NodeGetCapabilities` to discover Plugin capabilities. |
 
+#### `NodeDrain`
+
+A Node Plugin MUST implement this RPC call if the plugin has `DRAIN_VOLUMES` node capability.
+
+If the plugin has DRAIN_VOLUMES capability, the CO SHOULD call NodeDrain prior to terminating the plugin. During normal
```
jdef: Please review the contribution guidelines re: expected format for markdown text (e.g. one sentence per line), a single sentence shall not be split across multiple lines...

Also, this RPC is missing an error table that describes likely errors a CO might encounter and the expected CO behavior in such cases.

spec.md
```diff
@@ -1784,6 +1785,27 @@ Condition | gRPC Code | Description | Recovery Behavior
...
+#### `NodeDrain`
+
+A Node Plugin MUST implement this RPC call if the plugin has `DRAIN_VOLUMES` node capability.
+
+If the plugin has DRAIN_VOLUMES capability, the CO SHOULD call NodeDrain prior to terminating the plugin. During normal 
+operation, the CO will already have called `NodeUnpublishVolume` and `NodeUnstageVolume` on all volumes published or staged 
+on the node, however in some failure cases it is easy for clean-up to fail and mountpoints to be left on the node. 
+NodeDrain provides the plugin an opportunity to clean up.
+
+The CO MUST NOT call other RPCs on the node after NodeDrain is called.  The plugin should treat any other Node RPC after
```
jdef: "...MUST NOT call ... after NodeDrain is called." - because the plan is to reboot the agent?

Why do we need this limitation? Couldn't `NodeDrain` be invoked upon agent startup, to clear away any non-transient cobwebs that may have accumulated in a plugin's persistent store? Or is this too different of a use case?

jdef:
Is this still something that we want to pursue?

julian-hj:
Probably not.  I have a story in our Cloud Foundry backlog to investigate how much cleanup logic can be done at the CO layer when the node shuts down, without interacting with the Node plugin.  If that turns up anything that absolutely has to be done in the plugin, we can open a new PR with a more specific rationale documented.

For now, please feel free to go ahead and close this one.

PR #244, opened 2018-05-31
spec: Add efficient volume cloning support

Akrog:
With the new CSI snapshot feature it's now possible to create volumes
from a snapshots.  This may lead to users abusing it for unintended
functionality, such as having a golden image to source new volumes.

Most storage backends support efficient volume cloning, so it would be
benefitial to expose this functionality in the CSI spec.

By supporting cloning we avoid users creating a volume, and then a
snapshot of that volume, just so it can be used as the source for new
volumes.

The VolumeContentSource, that was added by the snapshots feature, will
be used for this feature.  New VolumeSource message is added to the
VolumeContentSource.


spec.md
```diff
@@ -602,6 +602,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume `name` already exists and is compatible with the specified `capacity_range`, `volume_capabilities` and `parameters` in the `CreateVolumeRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateVolumeResponse`.
 
+Plugins MAY create 3 types of volumes:
+
+- Empty volumes. When plugin supports `CREATE_DELETE_VOLUME` capability.
```
jdef: nit: the next two are "optional capability" but this is simply "capability" -- but it's also optional. why the difference?

spec.md
```diff
@@ -602,6 +602,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
...
+- From an existing snapshot. When plugin supports `CREATE_DELETE_SNAPSHOT` optional capability .
```
jdef: nit: extra space prior to the period

spec.md
```diff
@@ -796,6 +810,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+| Source incompatible or not supported | 3 INVALID_ARGUMENT | Besides the general cases, this code MUST also be used to indicate when plugin supporting CREATE_DELETE_VOLUME cannot create a volume from the requested source (`SnapshotSource` or `VolumeSource`).  Failure may be caused by not supporting the source (CO shouldn't have provided that source) or incompatibility between `parameters` from the source and the ones requested for the new volume. More human-readable information SHOULD be provided in the gRPC `status.message` field if the problem is the source. | On source related issues, caller MUST use different parameters, a different source, or no source at all. |
+| Source does not exist | 5 NOT_FOUND | Indicates that the source specified exist. | Caller MUST verify that the `volume_content_source` is correct, the source is accessible, and has not been deleted, before retrying with exponential back off. |
```
jdef: Typo? suggest: "Indicates that the specified source does not exist."

Also, the comma before "before" seems strange.

spec.md
```diff
@@ -1125,6 +1141,11 @@ message ControllerServiceCapability {
       // with the snapshot_id as the filter to query whether the
       // uploading process is complete or not.
       LIST_SNAPSHOTS = 6;
+      // Plugins supporting efficient volume cloning MAY report this
+      // capability.  Source volume must be managed by the same plugin.
```
jdef: we only use a single space between sentences.

also, suggest "A source volume..."

spec.md
```diff
@@ -1125,6 +1141,11 @@ message ControllerServiceCapability {
...
+      // Not all volume sources and parameters combinations MAY work.
```
jdef: MAY indicates a recommendation, or "soft" requirement - does it make sense in this context? or would "may" (non-RFC verbage) suffice?

spec.md
```diff
@@ -1125,6 +1141,11 @@ message ControllerServiceCapability {
...
+      // CLONE_VOLUME is NOT REQUIRED.
```
jdef: "... is NOT REQUIRED" seems redundant. It's a capability field *because* it's optional. I don't think there's a need to state that here.

spec.md
```diff
@@ -602,6 +602,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume `name` already exists and is compatible with the specified `capacity_range`, `volume_capabilities` and `parameters` in the `CreateVolumeRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateVolumeResponse`.
 
+Plugins MAY create 3 types of volumes:
+
+- Empty volumes. When plugin supports `CREATE_DELETE_VOLUME` capability.
+- From an existing snapshot. When plugin supports `CREATE_DELETE_SNAPSHOT` optional capability .
+- From an existing volume. When plugin supports efficient cloning, and reports the optional capability `CLONE_VOLUME`.
```
cpuguy83: This seems like an implementation detail. Should we support naive copies and leave it up to the plugin to handle it?

spec.md
Akrog: My bad, I'll update it.

spec.md
Akrog: Thanks.

spec.md
Akrog: Sorry, I've been writing too much on RST lately.

spec.md
Akrog: Done.

spec.md
Akrog: Normal may will suffice in my opinion.  Changing it.

spec.md
Akrog: I agree, I just copied from the CREATE_DELETE_SNAPSHOT in case this was something that should be done in the specs.  Removing it.

spec.md
Akrog: I wanted to indicate that the idea is for efficient cloning.  Not an attach vol1, attach vol2, then use dd.
Plugins can ignore it and internally do that, that's why I didn't add any MUST or REQUIRED. 
But plugins using the host assisted cloning mechanism (dd)  would give terrible performance and in my opinion would be frustrating.
Do you think I should remove the "efficient" adjective?

jdef:
Are any COs currently in a position to benefit from this?

cpuguy83:
Not here, but it's something we've discussed. I think it makes sense to support this at the CSI level at least.

Akrog:
Looks like some non CSI providers are currently implementing this via PVC annotations: https://github.com/kubernetes/cloud-provider-openstack/pull/183

aglitke:
In addition to openstack/cinder, I am aware of at least two other kubernetes storage providers which have some implementation: Gluster and NetApp Trident.  On the kubernetes side we are in early discussions with the storage-SIG on how to standardize on an approach.  Once this is done, kubernetes would be able to support this API by converting PVC annotations into a volume_source.

Akrog:
@j-griffith Yes, I'm interested in dusting this one off.  I hadn't noticed the patch was in conflict, fixed now.

spec.md
```diff
@@ -613,6 +613,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume `name` already exists, is accessible from `accessibility_requirements`, and is compatible with the specified `capacity_range`, `volume_capabilities` and `parameters` in the `CreateVolumeRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateVolumeResponse`.
 
+Plugins MAY create 3 types of volumes:
+
+- Empty volumes. When plugin supports `CREATE_DELETE_VOLUME` optional capability.
+- From an existing snapshot. When plugin supports `CREATE_DELETE_SNAPSHOT` optional capability.
+- From an existing volume. When plugin supports efficient cloning, and reports the optional capability `CLONE_VOLUME`.
```
jdef: nit: why is "efficient" a prerequisite for this capability? either a plugin has the cap or it doesn't.

Akrog: A plugin could do host assisted "cloning" by creating a new volume, attaching both volumes, and `dd`-ing contents from source to destination.

I don't think it would be a good idea mixing both concepts (fast storage backend cloning and slow host assisted cloning) under the same capability.  That would probably lead to unfulfilled expectations from the caller side.

We could add two capabilities instead of one, `CLONE_VOLUME` and `EFFICIENT_CLONING`. 
 That way plugins for backends supporting efficient cloning would report them both, and those doing host assisted cloning would just report the `CLONE_VOLUME` capability.  But my initial though is that this would be more confusing/troublesome.

j-griffith: So I think that in discussions we agreed that we didn't need/want to specify here and wanted to keep this generic in terms of the implementation with the caveat that if we need to we could consider additions for host assisted type cloning in the future?  If folks agree it'd be great to remove the specifics here and see if we can move forward with getting this merged.

spec.md
```diff
@@ -1009,6 +1023,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+| Source incompatible or not supported | 3 INVALID_ARGUMENT | Besides the general cases, this code MUST also be used to indicate when plugin supporting CREATE_DELETE_VOLUME cannot create a volume from the requested source (`SnapshotSource` or `VolumeSource`).  Failure may be caused by not supporting the source (CO shouldn't have provided that source) or incompatibility between `parameters` from the source and the ones requested for the new volume. More human-readable information SHOULD be provided in the gRPC `status.message` field if the problem is the source. | On source related issues, caller MUST use different parameters, a different source, or no source at all. |
+| Source does not exist | 5 NOT_FOUND | Indicates that the specified source does not exist. | Caller MUST verify that the `volume_content_source` is correct, the source is accessible, and has not been deleted before retrying with exponential back off. |
```
j-griffith: I think we can leverage this to solve the potential of some devices only allowing RO volumes to be cloned by @humblec on the Kubernetes implementation side.  IE even if we don't add caveats like RO or optimized etc to the spec we can respond with a reasonable error message if that's a requirement and it's not satisfied.  In other words this LGTM and provides what we'd need.

spec.md
Akrog: Thanks, I removed "efficient" from here, and added a comment below on the attaching and copying as agreed in the meeting.

spec.md
```diff
@@ -614,6 +614,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume `name` already exists, is accessible from `accessibility_requirements`, and is compatible with the specified `capacity_range`, `volume_capabilities` and `parameters` in the `CreateVolumeRequest`, the Plugin MUST reply `0 OK` with the corresponding `CreateVolumeResponse`.
 
+Plugins MAY create 3 types of volumes:
+
+- Empty volumes. When plugin supports `CREATE_DELETE_VOLUME` optional capability.
+- From an existing snapshot. When plugin supports `CREATE_DELETE_SNAPSHOT` optional capability.
```
saad-ali: When plugin supports `CREATE_DELETE_VOLUME` and `CREATE_DELETE_SNAPSHOT` optional capabilities.

spec.md
```diff
@@ -614,6 +614,12 @@ This RPC will be called by the CO to provision a new volume on behalf of a user
...
+- From an existing volume. When plugin supports cloning, and reports the optional capability `CLONE_VOLUME`.
```
saad-ali: When plugin supports cloning, and reports the optional capabilities `CREATE_DELETE_VOLUME` and `CLONE_VOLUME`.

spec.md
```diff
@@ -691,8 +697,16 @@ message VolumeContentSource {
     string id = 1;
   }
 
+  message VolumeSource {
+    // Contains identity information for the existing source volume.
+    // This field is REQUIRED.  Plugins reporting CLONE_VOLUME
```
saad-ali: nit: be consistent about one or two spaces after period.

spec.md
```diff
@@ -1012,6 +1026,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
 | Call not implemented | 12 UNIMPLEMENTED | CreateVolume call is not implemented by the plugin or disabled in the Plugin's current mode of operation. | Caller MUST NOT retry. Caller MAY call `ControllerGetCapabilities` or `NodeGetCapabilities` to discover Plugin capabilities. |
+| Source incompatible or not supported | 3 INVALID_ARGUMENT | Besides the general cases, this code MUST also be used to indicate when plugin supporting CREATE_DELETE_VOLUME cannot create a volume from the requested source (`SnapshotSource` or `VolumeSource`).  Failure may be caused by not supporting the source (CO shouldn't have provided that source) or incompatibility between `parameters` from the source and the ones requested for the new volume. More human-readable information SHOULD be provided in the gRPC `status.message` field if the problem is the source. | On source related issues, caller MUST use different parameters, a different source, or no source at all. |
```
jdef: nit: please re-order the new error row entries here so that the error codes flow from lowest to highest

spec.md
```diff
@@ -1354,6 +1370,14 @@ message ControllerServiceCapability {
       // with the snapshot_id as the filter to query whether the
       // uploading process is complete or not.
       LIST_SNAPSHOTS = 6;
+      // Plugins supporting volume cloning at the storage level MAY
+      // report this capability. The source volume must be managed by 
+      // the same plugin.  Not all volume sources and parameters
+      // combinations may work. Plugins MUST NOT implement this feature
+      // by attaching 2 volumes and copying the content, but COs may
+      // implement it that way for plugins that don't support this
+      // feature.
```
saad-ali: We should just drop:

```
      // Plugins MUST NOT implement this feature
      // by attaching 2 volumes and copying the content, but COs may
      // implement it that way for plugins that don't support this
      // feature.
```

Seems like an implementation detail. Good plugins won't do it, we don't need to spell it out.

spec.md
Akrog: Thanks, I failed to notice that errors were ordered.

spec.md
Akrog: Sorry about that.  I always leave two, but I know in this spec one is preferred, so I try to, but muscle memory sometimes gets the best of me. ;-)

spec.md
Akrog: I'll remove it, but after the "efficient cloning" discussion on this review, we decided in last weeks meeting (if I understood it right) to spell it out in the comments to ensure that plugins and COs were clear on this point, and users could expect a consistent behavior across plugins: cloning is a fast operation.

j-griffith: > I'll remove it, but after the "efficient cloning" discussion on this PR, we decided in last weeks meeting (if I understood it right) to spell it out in the comments to ensure that plugins and COs were clear on this point, and users could expect a consistent behavior across plugins: cloning is a fast operation.

We did talk about some comments but I think having it in the form of "MUST NOT" is too strong here, and it is an implementation detail so if someone wants to do it then hey, that's up to them.

Akrog: > We did talk about some comments but I think having it in the form of "MUST NOT" is too strong here, and it is an implementation detail so if someone wants to do it then hey, that's up to them.

That is the situation I wanted to avoid, unmet expectations of clone on some plugins/backends.  If a backend doesn't support efficient cloning and they report they do, then they are not following the specs, and users can call them out on it.

But maybe it's not such a big deal to people as I'm making it to be.

j-griffith: Meh... I don't know.  Anyway, LGTM; I'm still not a contributor so you just get my moral support  in the form of a +1

saad-ali:
Please rebase and we will merge this.

xing-yang:
LGTM but need a rebase.

j-griffith:
LGTM

Issue #127, opened 2017-10-23
Spec: Include an invalid credentials error on all calls that get the credentials to use


govint:
All calls to the plugin that get the credentials as an argument must include an error to handle invalid credentials.

jdef:
error codes were rewritten in #115. is this sufficient, or is there still a need for changes here?

PR #251, opened 2018-06-21
Add explicit note about optional ControllerUnpublish secrets.

jsafrane:
As discusses on community meeting this Wednesday, CO may loose access to secrets that were used in `ControllerPublish` call. Still, CO needs to clean up attached volumes and it must issue `ControllerUnpublish`.

`controller_unpublish_secrets` are already optional and only *SHOULD* be the same as in `ControllerPublish`, I am adding a possibility that CO *MAY* omit them.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1122,6 +1122,8 @@ This RPC is a reverse operation of `ControllerPublishVolume`.
 It MUST be called after all `NodeUnstageVolume` and `NodeUnpublishVolume` on the volume are called and succeed.
 The Plugin SHOULD perform the work that is necessary for making the volume ready to be consumed by a different node.
 The Plugin MUST NOT assume that this RPC will be executed on the node where the volume was previously used.
+The CO MAY omit `controller_unpublish_secrets` in case when it looses access to the secrets after `ControllerPublishVolume` call.
+It is RECOMMENDED for the Plugin not to depend on presence of `controller_unpublish_secrets`.
 
 This RPC is typically called by the CO when the workload using the volume is being moved to a different node, or all the workload using the volume on a node has finished.
 


spec.md
```diff
@@ -1122,6 +1122,7 @@ This RPC is a reverse operation of `ControllerPublishVolume`.
 It MUST be called after all `NodeUnstageVolume` and `NodeUnpublishVolume` on the volume are called and succeed.
 The Plugin SHOULD perform the work that is necessary for making the volume ready to be consumed by a different node.
 The Plugin MUST NOT assume that this RPC will be executed on the node where the volume was previously used.
+The CO MAY omit `controller_unpublish_secrets` in case when it looses access to the secrets after `ControllerPublishVolume` call. It is RECOMMENDED for the Plugin not to depend on presence of `controller_unpublish_secrets`.
```
jieyu: nits: please have one sentence per line.

jsafrane:
> I.e. why only modify this call and not node-unstage-vol or node-unpublish-vol? 

Because NodeUnstage and NodeUnpublish don't have secrets at all.

Secrets are present in:

* CreateVolume, CreateSnapshot, ControllerPublish, NodeStage and NodePublish - I think it's useful to have them there, at least to transfer encryption key.

* DeleteVolume, DeleteSnapshot - I *personally* think it's not necessary to have them there, but I don't mind. CO may loose the secrets, so it won't be able to delete these objects, but the CO I care about has clear error paths here. In addition, backend storage system can typically handle thousands of volumes and few undeletable volumes won't matter, while typically only limited number of volumes can be attached to a node (10 on Digital Ocean!) and every undetachable volume hurts a lot. Again, this is *personal* experience.

spec.md
jsafrane: fixed

jdef:
I agree that undetachable volumes hurt. We've experienced the same problem in our CO/volume-driver integrations. That said, I still have a concern about removing the controller-unpublish-secrets: if an SP requires credentials for such ops .. how can we legitimately state that a plugin shouldn't depend upon them in the RPC request? It seems that by doing so we'd be implicitly requiring *plugins* to cache secrets and that's something that I do NOT think we should force plugins to do.

Maybe what we're missing is a more clear recovery path for CO's that lose credentials. I don't think the spec is very clear w/ respect to errors due to credential problems. For example, there are gRPC error codes specifically mapped to (a) authentication errors, and; (b) authorization errors - but the CSI spec doesn't mention them at all, either in the "common errors" section nor on a per-call basis.

Consider: if a CO issued a call that didn't have the proper secrets, and a plugin returned a clear error code indicating "you're not properly authenticated" - then that seems like a clear signal to the CO re: the proper recovery behavior. It's on the CO and/or admin to ensure that secrets don't go stale and are properly reinstated if temporarily lost.

Also, related to your specific case, if k8s forgets the controller credentials for a controller-unpublish call ... won't those credentials also be missing for subsequent controller-publish calls? In other words, won't the future attachment of volume be blocked anyway because the CO forgot the controller-level credentials?

jsafrane:
>  if a CO issued a call that didn't have the proper secrets, and a plugin returned a clear error code indicating "you're not properly authenticated" - then that seems like a clear signal to the CO re: the proper recovery behavior

That sounds reasonable. It gives us opportunity to start some alarms.

>  if k8s forgets the controller credentials for a controller-unpublish call ... won't those credentials also be missing for subsequent controller-publish calls?

Yes, they will be missing, but failed ControllerPublish won't leave leaked attachment behind. In addition, users will see that their containers won't start and start nagging admin or someone. Users won't notice failed ControllerUnpublish, because their containers finished just fine.

jdef:
It's not clear to me that all plugins will be able to honor this proposal. If attach/detach is governed by controller publish/unpublish and the secrets for unpublish go missing, but are actually required by the storage backend in order for the plugin to actually perform a detach, then we've gained nothing because the plugin can't actually execute the detach anyway (because it's unable to securely communicate w/ the storage backend to exec the operation).

The currently proposed language implies that it's not perilous for plugins to be handed a ControllerUnpublish request w/o secrets (where they are expected) when, in reality, it might be very problematic: the attachment will still be left behind.

Is this actually a problem that CSI can solve? If a storage backend requires secrets for detaching a volume and the CO loses them, then what is a plugin actually going to be able to do about it?

jdef:
#127 talks about adding explicit error codes for authn/authz problems related to invalid credentials

spec.md
```diff
@@ -1122,6 +1122,8 @@ This RPC is a reverse operation of `ControllerPublishVolume`.
 It MUST be called after all `NodeUnstageVolume` and `NodeUnpublishVolume` on the volume are called and succeed.
 The Plugin SHOULD perform the work that is necessary for making the volume ready to be consumed by a different node.
 The Plugin MUST NOT assume that this RPC will be executed on the node where the volume was previously used.
+The CO MAY omit `controller_unpublish_secrets` in case when it looses access to the secrets after `ControllerPublishVolume` call.
```
julian-hj: nit: s/looses/loses/

saad-ali:
It sucks, but I think this is working as intended. If CO may loses access to secrets, that just means that the CO is not permitted to operate on it. The cleanup should fail until the end user restores the COs permission to do cleanup. Being able to cleanup without secrets could be a security issue.

PR #252, opened 2018-06-22
Add workload information to NodePublish.

### LINKED ISSUE #248: CreateVolumeRequest should provide CO related volume resource information
This issue is tightly coupled with k8s. Some storage provisioning mechanism heavily depends on the CO itself. They leverage the architecture of CO and its resources to provision volume. Before CSI the volume representative information in CO(eg. pvc in kubernetes) was passed to the custom provisioner. The onus of extracting the volume related information was offloaded to the provisioner.
The current CSI spec restricts the free flow of CO's volume resource information(eg PVC/SC in K8s) to the driver. It is however possible to pass the information through redundant declaration in the storage class parameters but this becomes hard to maintain.
I propose that the CO's should pass their volume related claims information either as a map or a more defined struct to the CSI driver.

Related issue https://github.com/kubernetes-csi/external-provisioner/issues/93

jsafrane:
Fixes #248 
(see the bug for long discussion)

Idea behind `Workload.co_name` is that every CO will use its own name (say `"kubernetes.io:1.12"`) that specifies what keys + values are present in the `Workload.parameters`, incl. their format. The CO may define several different `Workload.co_name` values (and thus set of parameters/values), depending e.g. on CO version (say `"kubernetes.io:1.13"` may add a new parameter).

The Plugin is allowed to ignore the Workload, if it can complete the request without it. Which covers IMO all plugins with "real" storage backend.

`Workload` will be used primarily in plugins that provide ephemeral volumes with configuration, secrets, identity or similar data for workloads. These are tightly bound to CO and thus must be prepared to `Workload.co_name` changes (and new/changed/deleted `parameters`).


spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
...
+
+// Parameters of a CO workload that lead to CSI request.
+message Workload {
+    // CO name. This name defines content of parameters map. Each CO
+    // SHOULD clearly document which co_name it uses and which parameter
+    // names and format of values it provides. This filed is REQUIRED.
+    string co_name = 1;
+
+    // CO specific parameters of the workload, such as workload name.
+    // Exact format depends on the CO. This filed is REQUIRED.
```
jieyu: I won't make this field "REQUIRED". I think it's OK that CO sets the `co_name` only in some cases.

spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
...
+  // Information about workload that wants to use the volume.
+  // This is an OPTIONAL field. The Plugin MAY choose to ignore
+  // this field if it does not need any workload related information
+  // to fullfill the request.
+  Workload workload = 9;
+}
+
+// Parameters of a CO workload that lead to CSI request.
+message Workload {
+    // CO name. This name defines content of parameters map. Each CO
```
jieyu: nits: fix indentation. the indentation should be 2 here.

spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
   // MUST match the attributes of the Volume identified by
   // `volume_id`.
   map<string, string> volume_attributes = 8;
+
+  // Information about workload that wants to use the volume.
```
jieyu: nits: about *the* workload

spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
...
+  // This is an OPTIONAL field. The Plugin MAY choose to ignore
```
jieyu: I would be more explicit here:
```
This is an OPTIONAL field. Some CO MAY choose not to set this field.
```

spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
...
+  // this field if it does not need any workload related information
+  // to fullfill the request.
```
jdef: Sp: fulfill

spec.md
```diff
@@ -1849,6 +1849,24 @@ message NodePublishVolumeRequest {
...
+  Workload workload = 9;
+}
+
+// Parameters of a CO workload that lead to CSI request.
+message Workload {
+    // CO name. This name defines content of parameters map. Each CO
+    // SHOULD clearly document which co_name it uses and which parameter
+    // names and format of values it provides. This filed is REQUIRED.
```
jdef: s/filed/field/

spec.md
jdef: Ditto: field

spec.md
jsafrane: fixed

spec.md
jsafrane: fixed

spec.md
jsafrane: updated with your suggestion

spec.md
jsafrane: fixed

spec.md
jsafrane: fixed

spec.md
jsafrane: If there is no `co_name`, how the plugin knows what to expect in the parameters? Do we expect that COs will share some parameters names?	I think CSI should be strict on the "sending" side, CO must fill it, and benevolent on "receiving" side, the plugin may choose to ignore it.

jieyu: Sorry, I was meant to say that `parameters` needs to be optional.

jieyu:
Discussed this in the community sync regarding if `parameters` should be optional or not.

The reason for it being optional is for the case in the future that we introduce some first class fields in the `Workload` struct that's common to all COs. If that's the case, I think it's better to make `parameters` optional because the CO might just set other first class fields there.

However, we don't have a first class field there yet. So the argument is kind of weak. I can go either way, though I still prefer it being optional.

cc @saad-ali @jsafrane @jdef

jieyu:
RE @jdef's comments on a big warning, this is what I would suggest:
```
// Information about the workload that wants to use the volume.
// This is an OPTIONAL field. Some CO MAY choose not to set this
// field. The Plugin SHOULD avoid relying on this field if possible.
// The Plugin MAY choose to ignore this field if it does not
// need any workload related information to fulfill the request.
```

jsafrane:
@jieyu, thanks for your suggestion, I updated the PR with it.

jieyu:
@jsafrane can you fix the CI failure? (re-run make?)

jsafrane:
regenerated proto files

jsafrane:
I'll try to fix this in Kubernetes only, without any changes in CSI. Thanks for all the feedback here!

PR #254, opened 2018-07-18
Fix proto filedescriptor

jdef:
Generated golang bindings invoke proto.Register with a "file name" and a metadata blob. The primary goal of this change is to support protobuf reflection libraries that attempt to examine the generated metadata blobs; abbreviated file names (prior to this change set) make reflection difficult. A secondary goal is to avoid overlap with other non-fully qualified proto packages: a longer, fully qualified package root has
less chance of overlapping with that of another package in the wild.


saad-ali:
LGTM. But I'm not a go expert either.

ddebroy:
Not-an-expert LGTM

jdef:
This isn't so much a "go expert" thing as a "golang/protobuf" thing.

Issue #236, opened 2018-05-15
Volume usage metrics support


gnufied:
This came up during sig-storage f2f, we currently don't have a reliable way of getting volume usage metrics from CSI plugins. One can argue that - one can use "du" or "stat" from CO to get those metrics and this need not be part of CSI specs, but I believe that how "usage" metric of a volume is computed depends on volume type and should be part of CSI spec.

In Kubernetes at least currently - each volume type implements its own metric implementation.

jdef:
FWIW there's a couple of basic measurements that I think are important for quota/usage within a volume:

1. data capacity
2. metadata capacity (read: inodes (fstype-specific) there may be a finite amount of these to go around)

I wonder if it's worth reporting raw values (MB, GB, inode counts) or percentages. I guess the benefit of implementing metrics for these things in a CSI plugin is that it pushes knowledge (and any related tooling dependencies) of the underlying storage further into the plugin layer (as opposed to the CO layer). For BLOCK volumes I can think of a single scenario in which capacity might be accurately measured by a plugin: sparse BLOCK volumes (e.g. dynamically expanding virtual disk); otherwise BLOCKs are generally opaque and it's not clear to me how a plugin would measure usage of one.

Pushing this complexity into the plugin layer simplifies the CO at the expense of plugins. However it's also arguable that storage-domain specific knowledge **should** be captured at the plugin layer. Is higher complexity in the plugin layer acceptable in this case?

It's probably also worth thinking about this in the context of "volume health checks", something that we (the COs, at least) discussed a while ago but was put on the back burner.

I'd like to see some use cases spelled out for why this is important. Some initial thoughts:

(a) The CO should monitor available volume capacity so that it may emit an administrative alert / event.

(b) The CO should monitor available volume capacity for the purposes of auto-expanding a volume when available capacity falls below some threshold.

Are there others?

gnufied:
Currently in Kubernetes we are reporting following metrics:

1. volume_stats_used_bytes
2. volume_stats_available_bytes
3. volume_stats_inodes_free
4. volume_stats_inodes
5. volume_stats_inodes_used
6. volume_stats_capacity_bytes

I think at minimum we will want similar or same metrics from CSI. 

The other use case we have is - quite simply user wants to monitor its PVC usage(volume) in some kind of metrics gathering system. Any action that user may take is upto the user but monitoring the volumes for usage gets them halfway there.

jieyu:
[...]

PR #256, opened 2018-07-23
A design proposal for volume capacity from node

gnufied:
Fixes https://github.com/container-storage-interface/spec/issues/236

cc @saad-ali @jdef


csi.proto
```diff
@@ -1020,6 +1023,54 @@ message NodeUnpublishVolumeRequest {
...
+  // this is the total size of underlying storage and it MAY
+  // not be equal to used_capacity + available_capacity because
+  // filesystem is shared.
+  // The value of this field MUST NOT be negative.
+  int64 total_capacity = 3;
+
+  // The number of inodes used by the volume.
+  // This field is OPTIONAL.
+  // The value of this field MUST NOT be negative.
+  int64 used_inodes = 4;
```
saad-ali: I agree with @ddebroy that we have to avoid leaking OS specific constructs in to the spec. Is there a similar concept in Windows, could we come up with something that would work for all OSes?

garthy:
Cluster size maybe for the NTFS volume but personally I'd not bother
exporting filesystem specific tlinformation. How will it get used? Is it
useful?

Export it via the metrics if needed.

csi.proto
gnufied: Yeah I think our options boil down to:

a. Either come up with a name that may or may not represent inode number in a OS agnostic way. It seems hard to do and could easily cause confusion.
b. Drop the inode information altogether from stat information.
d. Keep the inode capacity information as it is and make it clear that plugin authors need not set this value when running on a platform that does not support it. These numbers will not be available when using raw block devices on *nix too. So, Windows is not alone in this.
e. @msau42 suggested we can move such platform specific information in a opaque dictionary. It could work but I am not sure if it is very different from option#d.

csi.proto
```diff
@@ -1020,6 +1023,54 @@ message NodeUnpublishVolumeRequest {
...
+  // The ID of the volume. This field is REQUIRED.
+  string volume_id = 1;
+
+  // The path at which the volume was published. It MUST be an absolute
+  // path in the root filesystem of the process serving this request.
+  // This is a REQUIRED field.
+  string target_path = 2;
+}
+
+message NodeGetVolumeStatsResponse {
```
jdef: What about consolidating this a bit w/ units that we can extend later?
```protobuf
message Usage {
  enum Unit {
    UNKNOWN = 0;
    INODE = 1;
    CLUSTER = 2;
  }

  int64 free = 1;
  int64 total = 2;
  Unit unit = 3;
} 

repeated Usage usage = 1;
```

gnufied: fixed

spec.md
```diff
@@ -1913,6 +1916,78 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Operation pending for volume | 10 ABORTED | Indicates that there is a already an operation pending for the specified volume. In general the Cluster Orchestrator (CO) is responsible for ensuring that there is no more than one call "in-flight" per volume at a given time. However, in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and MAY issue multiple calls simultaneously for the same volume. The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls. | Caller SHOULD ensure that there are no other calls pending for the specified volume, and then retry with exponential back off. |
 
 
+#### `NodeGetVolumeStats`
+
+A Node plugin MUST implement this RPC call.
```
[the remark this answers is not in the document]

spec.md
```diff
@@ -1913,6 +1916,78 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+```protobuf
+message NodeGetVolumeStatsRequest {
+  // The ID of the volume. This field is REQUIRED.
+  string volume_id = 1;
+
+  // The path at which the volume was published. It MUST be an absolute
+  // path in the root filesystem of the process serving this request.
+  // This is a REQUIRED field.
+  string target_path = 2;
```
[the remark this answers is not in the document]
gnufied: I have renamed this field to `volume_path` and updated the comment to reflect as per our discussion in community call.

spec.md
```diff
@@ -1913,6 +1916,78 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // this is available space on the underlying storage and is
+  // shared with the host processes and other volumes.
+  // The value of this field MUST NOT be negative.
+  int64 available_capacity = 2;
+
+  // The total capacity in bytes. This field is REQUIRED.
+  // For volumes that share a file system with the host (e.g hostpath)
+  // this is the total size of underlying storage and it MAY
+  // not be equal to used_capacity + available_capacity because
+  // filesystem is shared.
```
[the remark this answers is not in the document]
gnufied: I have dropped these elaborate comments, because host paths that are shared with underlying storage may or may not be a strong use case. I expect plugin implementations will use their judgement to return right values from here.

gnufied:
@jdef @jieyu fixed or addressed most review comments. PTAL.

gnufied:
@jdef @jieyu ping again. Can you guys please take another look?

csi.proto
```diff
@@ -1020,6 +1023,41 @@ message NodeUnpublishVolumeRequest {
...
+  string volume_path = 2;
+}
+
+message NodeGetVolumeStatsResponse {
+  repeated VolumeUsage usage = 1;
+}
+
+message VolumeUsage {
+  enum Unit {
+    BYTES = 0;
```
ddebroy: Nit: Most enums typically have a `UNKNOWN = 0;` Can we add one here too?

jieyu: Good catch @ddebroy!

gnufied: Fixed!

saad-ali:
LGTM

csi.proto
ddebroy: thanks .. LGTM

spec.md
```diff
@@ -1913,6 +1916,66 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+##### NodeGetVolumeStats Errors
+
+If the plugin is unable to complete the `NodeGetVolumeStats` call successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return the specified gRPC error code.
+The CO MUST implement the specified error recovery behavior when it encounters the gRPC error code.
+
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
```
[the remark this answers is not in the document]
gnufied: This is the generic error we return when specified volume does not exist. See `NodeUnpublishVolume`  for example.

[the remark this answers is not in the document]
jdef: This seems like a different case than node-unpublish-volume. For example, if the volume_id is already node-unpublished at target_path, then target_path may not exist (this is an expected possibility) - the call is deemed idempotent.

For this RPC, a missing vol staged/published at target_path is an error because of the nature of the operation: the RPC isn't changing the mount state of the volume at target_path, it's trying to query the state of the volume at target_path. And if volume_id exists, but it not present at target_path .. well, that's an error that we should bubble up to the CO.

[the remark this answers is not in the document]
jdef: One might argue that NOT_FOUND isn't appropriate for the case I outlined above and that we should use a different error code for such a case - I'm fine with that too.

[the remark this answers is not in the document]
gnufied: hmm I think the only other error code that makes sense perhaps is `FAILED_PRECONDITION` but in this case, it is not all that different from `NOT_FOUND` error because in both the cases the CO may retry.

I think I am okay with rewording the error to "Indicates that a volume corresponding to `volume_id` does not exist on specified `volume_path`"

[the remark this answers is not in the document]
gnufied: I have updated the comment.

spec.md
```diff
@@ -1913,6 +1916,67 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+  // The total capacity in specified Unit. This field is REQUIRED.
+  // The value of this field MUST NOT be negative.
+  int64 total = 2;
+
+  // The used capacity in specified Unit. This field is OPTIONAL.
+  // The value of this field MUST NOT be negative.
+  int64 used = 3;
+
+  // Units by which values are measured.
```
jdef: This field is REQUIRED.

spec.md
```diff
@@ -1913,6 +1916,67 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // It can be any valid path where volume was previously
+  // staged or published.
+  // It MUST be an absolute path in the root filesystem of
+  // the process serving this request.
+  // This is a REQUIRED field.
+  string volume_path = 2;
+}
+
+message NodeGetVolumeStatsResponse {
+  repeated VolumeUsage usage = 1;
```
jdef: this field is ... OPTIONAL?

spec.md
```diff
@@ -1968,6 +2032,10 @@ message NodeServiceCapability {
     enum Type {
       UNKNOWN = 0;
       STAGE_UNSTAGE_VOLUME = 1;
+      // If Plugin implements GET_VOLUME_STATS capability
+      // then it MUST implement NodeGetVolumeStats RPC
+      // call for fetching volume stats
```
jdef: s/stats/statistics/ (... and end sentences with punctuation please).

gnufied:
@jdef done. thank you. I will squash when ready for merge.

jdef:
Going forward, let's also hold ourselves to a higher standard w/ respect to descriptive PR summaries and/or commit messages. "... a design proposal ..." doesn't cut it here: this was an actual specification change not just a proposal.

PR #265, opened 2018-08-20
doc: Remove REQUIRED for boolean values

Akrog:
Setting REQUIRED on boolean values is misleading, as it may cause CSI
implementers to think that the field must be present in the GRPC payload
and that they should fail if it's not there, which is incorrect.

As mentioned in the GRPC Proto3 docs: "note that if a scalar message
field is set to its default, the value will not be serialized on the wire."

This means that boolean fields will not be sent over the wire when their
value is set to false.  Making it impossible to know whether the GRPC
message had the field set in the first place.

This patch updates the comments to reflect the default value and removes
the misleading REQUIRED notation.


jdef:
Who is making decisions about CSI messages by looking at gRPC message wire format? Yes, it may be difficult for a receiver to know whether a sender actually attempted to send a default value, or simply didn't set a value for some scalar field. If **that** is the problem that we're trying to solve, then instead of removing REQUIRED we could use the Well Known Type cousins of the scalar types (they're basically common message wrappers around scalars).

Is this proving to be a problem in practice?

Akrog:
Decisions are made reading the CSI specs.  If the specs say that a field is **REQUIRED**, then it is common sense that the field **will be present in the message**.   So the right thing to do would be to have a check in the CSI plugin to ensure that the field is present and fail if it's not.

I would say this has been an issue for at least one person (me), as  [I created an issue on csi-sanity](https://github.com/kubernetes-csi/csi-test/issues/86) and [removed the check from my code](https://github.com/Akrog/ember-csi/commit/dfc8600b899fbc7b19b72db5abd65806ff3db58e) before  realizing this was how gRPC works, and that we could not know if the field was present when the call was made.

So I wanted to avoid this issue for anybody else.  After all, given the way gRPC works, saying *REQUIRED* on a `boolean` field means literally *nothing*.  So why would we want to say something in the spec that isn't true?

jdef:
So the argument is "because REQUIRED isn't verifiable for boolean fields in
gRPC messages, the spec should avoid using it for such" ?

Akrog:
I can try to phrase it a little bit better:
> It's incorrect to say that `boolean` fields are *REQUIRED*.  They default to false when not set, and therefore are not *REQUIRED* in any way.

saad-ali:
Being able to differentiate between REQUIRED and OPTIONAL is useful. Maybe we can add a generic comment at the top of the doc that says for REQUIRED scalar fields they will be defaulted if not specified per proto3.

This is more of a problem for OPTIONAL fields that are scalar. But, we took a look at the existing OPTIONAL fields and only found one in `NodeGetInfo`, and that one is the same in both unspecified and unset so not an issue.

Akrog:
> Being able to differentiate between REQUIRED and OPTIONAL is useful. Maybe we can add a generic comment at the top of the doc that says for REQUIRED scalar fields they will be defaulted if not specified per proto3.
> 

This seems like an acceptable compromise.  We can make developers aware of this implementation detail while keeping the spec more agnostic from the underlying transport protocol.

spec.md
```diff
@@ -16,6 +16,8 @@ The key words "unspecified", "undefined", and "implementation-defined" are to be
 An implementation is not compliant if it fails to satisfy one or more of the MUST, REQUIRED, or SHALL requirements for the protocols it implements.
 An implementation is compliant if it satisfies all the MUST, REQUIRED, and SHALL requirements for the protocols it implements.
 
+Scalar fields, even REQUIRED ones, will be defaulted if not specified, and any field set to the defaul value will not be serialized over the wire as per [proto3](https://developers.google.com/protocol-buffers/docs/proto3#default).
```
saad-ali: Wording looks good. But can we move this to a section right after the [Size Limits section](https://github.com/container-storage-interface/spec/blob/master/spec.md#size-limits)?

Akrog: Sure, it does makes more sense to be in the Required vs Optional section.

saad-ali:
LGTM

PR #266, opened 2018-08-21
spec: clean up Concurrency, refactor refs to ABORTED

jdef:
- OPERATION_PENDING_FOR_VOLUME is no longer defined by the spec, remove it.
- ABORTED is the common-case error code for conflicting, overlapping ops.
- consolidate duplicate explanations of ABORTED, it is the same for most RPCs.


jdef:
PTAL @saad-ali @jieyu @julian-hj @ddebroy

jdef:
I left it in because of the snapshot-specific wording: """that there is no
more than one call "in-flight" per snapshot at a given time""". If you
think that's kind of irrelevant then let me know and I will consolidate.

PR #269, opened 2018-09-20
Bump golang and protoc-gen-go

jdef:
* golang v1.10.4
* protoc-gen-go v1.2.0


jdef:
@jieyu @saad-ali @julian-hj @ddebroy PTAL

PR #271, opened 2018-09-27
spec: Align volume validation with creation.

### LINKED ISSUE #242: Consider making ValidateVolumeCapabilitesRequest align with CreateVolumeRequest
Paraphrased from https://github.com/container-storage-interface/spec/pull/188#discussion_r191543503:

The arguments to `ValidateVolumeCapability` should be consistent with those in `CreateVolumeRequest` (i.e., `CapacityRange`, `TopologyRequirement`, etc.) because that's what COs should care about when dealing with pre-existing volumes the same way as if they're dealing with dynamically provisioned volumes.

jdef:
THIS IS A BREAKING CHANGE.

* ValidateVolumeCapabilitiesRequest accepts the same parameters and
  attributes requirements as CreateVolumeRequest.

* ValidateVolumeCapabilitiesResponse is forward-compatible between newer
  COs (clients) and older plugins (servers).

* fixes #242


csi.proto
```diff
@@ -628,28 +628,44 @@ message ValidateVolumeCapabilitiesRequest {
...
+  // plugin has validated. This field SHALL only be set to a non-empty
+  // value for successful validation responses.
+  // For successful validation responses, the CO SHALL compare the
+  // fields of this message to the originally requested capabilities in
+  // order to guard against an older plugin reporting "valid" for newer
+  // capability fields that it does not yet understand.
+  // This field is OPTIONAL.
+  Confirmed confirmed = 1;
 
   // Message to the CO if `supported` above is false. This field is
```
jdef: nit: `supported` is no longer a field

jdef:
TODO: add secrets, storage backend might need them to execute validation for a pre-existing volume

csi.proto
```diff
@@ -628,30 +628,51 @@ message ValidateVolumeCapabilitiesRequest {
...
   // specified below are supported. This field is REQUIRED.
-  repeated VolumeCapability volume_capabilities = 2;
+  repeated VolumeCapability volume_capabilities = 3;
 
-  // Attributes of the volume to check. This field is OPTIONAL and MUST
-  // match the attributes of the Volume identified by `volume_id`.
-  map<string, string> volume_attributes = 3;
+  // See CreateVolumeRequest.parameters.
+  // This field is OPTIONAL.
+  map<string, string> parameters = 4;
```
saad-ali: This should be optional? Explicitly should not error

jdef: already says `This field is OPTIONAL.` Is more needed?

spec.md
```diff
@@ -1172,38 +1172,62 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 A Controller Plugin MUST implement this RPC call.
 This RPC will be called by the CO to check if a pre-provisioned volume has all the capabilities that the CO wants.
-This RPC call SHALL return `supported` only if all the volume capabilities specified in the request are supported.
+This RPC call SHALL return `confirmed` only if all the volume capabilities specified in the request are supported.
```
jdef: This wording still isn't quite right. `confirmed` may contain a subset of what was validated

spec.md
```diff
@@ -1172,38 +1172,62 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
-  bool supported = 1;
+  message Confirmed {
+    // Volume capabilities supported by the plugin.
+    // This field is REQUIRED.
+    repeated VolumeCapability volume_capabilities = 1;
+
+    // Topology requirements supported by the plugin.
+    // This field is OPTIONAL.
+    TopologyRequirement accessibility_requirements = 2;
+  }
```
jdef: add parameters and attributes to `Confirmed`

julian-hj: After looking a bit more closely, I am actually a bit confused why we need topology in this RPC at all.  Since `ListVolume` returns a list of `Volume` messages, and `Volume` contains `accessible_topology` the CO should already have all the information it requires about topology just by listing the volumes.

It makes total sense to me to include `parameters` but maybe we don't need  `accessibility_requirements`?  Or are you thinking that you don't want the CO to have to do the logic to figure out if `accessible_topology` satisfies `accessibility_requirements`

jdef: Discussed this with @jieyu earlier today. I think we're OK with removing topology from the validation call. Would that be satisfactory?

julian-hj: sounds good to me.

jdef:
addressed remaining feedback and rebased to master

csi.proto
```diff
@@ -631,30 +631,51 @@ message ValidateVolumeCapabilitiesRequest {
   // The ID of the volume to check. This field is REQUIRED.
   string volume_id = 1;
 
+  // Attributes of the volume to check. This field is OPTIONAL and MUST
+  // match the attributes of the Volume identified by `volume_id`.
+  map<string, string> volume_attributes = 2;
+
   // The capabilities that the CO wants to check for the volume. This
   // call SHALL return "supported" only if all the volume capabilities
```
jdef: ugh, another reference to "supported"

saad-ali:
LGTM

PR #275, opened 2018-10-09
Fix typos in spec

jicahoo:
As title.


jicahoo:
It seems that the doc was generated. I changed the wrong place. I will try to fix the correct places. Close this PR.

PR #281, opened 2018-10-23
Update Contributing.MD with required CCLA

josepalafox:
This fix will allow large corporations like Intel and their legal departments to easily identify the required CCLA and sign it.


jdef:
PTAL @jieyu @saad-ali @julian-hj @ddebroy

saad-ali:
Late LGTM!

saad-ali:
Actually we should change the `which should be signed and mailed to the [mailing list]( https://groups.google.com/forum/#!topic/container-storage-interface-community/)`

to approver list instead of community list.

PR #298, opened 2018-11-06
Reword Language regarding DeleteVolume/DeleteSnapshot

### LINKED ISSUE #225: reword: storage space associated with ... MUST be released
as noted in https://github.com/container-storage-interface/spec/pull/224#discussion_r184213925 the wording in the spec w/ respect to when the backend storage system must "release" resources should be re-evaluated. there are a variety of ways in which backend storage systems may implement delete operations and resource recovery. the CSI spec should be implementation independent and take care to avoid assuming too much w/ respect to backend system operations.

xing-yang:
This PR reworded some language regarding DeleteVolume and
DeleteSnapshot.

Fixes #225


xing-yang:
@jdef @saad-ali Please take a look.

spec.md
```diff
@@ -1015,7 +1015,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_VOLUME` capability.
 This RPC will be called by the CO to deprovision a volume.
-If successful, the storage space associated with the volume MUST be released and all the data in the volume SHALL NOT be accessible anymore.
+If successful, the storage space associated with the volume MAY or MAY NOT be released depending on the reclaim or retention policy specified on the storage system.
```
j-griffith: The only nit I have here is the use of "reclaim policy" that gives the impression that this may be related to Kubernetes reclaim policy, but in fact it should be soley based upon the backend and it's configuration no?

spec.md
```diff
@@ -1500,7 +1500,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_SNAPSHOT` capability.
 This RPC will be called by the CO to delete a snapshot.
-If successful, the storage space associated with the snapshot MUST be released and all the data in the snapshot SHALL NOT be accessible anymore.
+If successful, the storage space associated with the snapshot MAY or MAY NOT be released depending on the reclaim or retention policy specified on the storage system.
```
j-griffith: Same note regarding Kubernetes centric phrasing.  I understand that's not the intent but was thinking we could be more agnostic here and still get the point across.  Thoughts?

spec.md
j-griffith: @xing-yang  Yeah, maybe omit the "if successful" and then maybe just something like:

> The space associated with teh volume MAY or MAY NOT be released depending upon the configuration of the back end device including any available retention policies.

The only thing for me was "retention policy" being a parameter in Kubernetes it just made it seem too specific.  What do you think?

xing-yang: @j-griffith I like it.  I updated the PR with your suggestion.  I still added "If successful, ..." to start with the sentence.

julian-hj:
To me, this still seems unnecessarily specific.  I am not sure what we gain by having language about retention policies, which again gets into plugin implementation.  Would we not achieve the same intent by just saying `If successful, the storage space associated with the volume MAY be released.`?

spec.md
```diff
@@ -1015,7 +1015,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 A Controller Plugin MUST implement this RPC call if it has `CREATE_DELETE_VOLUME` capability.
 This RPC will be called by the CO to deprovision a volume.
-If successful, the storage space associated with the volume MUST be released and all the data in the volume SHALL NOT be accessible anymore.
+If successful, the space associated with the volume MAY or MAY NOT be released depending upon the configuration of the back end device including any available retention policies.
```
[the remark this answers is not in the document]
xing-yang: Sure.  Removed "If successful..." lines.

xing-yang:
Addressed review comments.  @saad-ali @julian-hj please take a look.

PR #303, opened 2018-11-07
Disallow control characters in name fields

### LINKED ISSUE #267: String encode rules
The current spec doesn't say anything about whether Unicode characters are allowed in various string fields (most notably the volume name), and if so, whether UTF-8 encoding is required or whether other encodings are allowed.

Also, it's not clear whether whitespace, newlines, tabs, or control characters including backspaces, NULs, EOFs, BELLs, etc should be allowed in string fields.

Some fields clearly specify alphanumeric values, and we could probably expand this language to other fields, or put some general language about all strings in the document.

bswartz:
Earlier versions of the spec put no limits on name fields, but
indicate that names should be usable as identifiers on storage
backends. Control characers obviously conflict with this purpose,
so specify that they're prohibitted, while making it clear that
all other valid Unicode strings are allowed.

Fixes #267


jdef:
Since we're considering limits for 1 of the name fields in the spec, should we consider a general limit for "name"-like fields throughout the spec?

bswartz:
@jdef I think I got them all. There are only 2 name fields sent in the CO->Plugin direction. There's another name field in the Plugin->CO direction which already have very stringent restrictions. What other name-like fields did you have in mind?

jdef:
@bswartz I think you're right. I don't see anything else.

bswartz:
Rebased

Issue #293, opened 2018-11-01
ControllerPublishVolume readonly field not supported?


saad-ali:
We should clarify the behavior if a storage system does not support readonly==true on ControllerPublishVolume.

saad-ali:
We should also clarify what readonly means. Maybe: regardless of the mount (NodePublishVolume) readonly flag, this volume is readonly at the storage layer?

saad-ali:
CO may not be able to know when to pass in readonly on ControllerPublish.

Hard answer is new capability and extra CO logic to discover when to do this.

Easy answer is plugin should just ignore the field, if it doesn't support it?

Conclusion after discussing with @jdef and @julian-hj let's go with the easy option and clarify it in the spec.

saad-ali:
Discussed with @julian-hj and @jieyu, we decided that the easy option leaves potential issue if CO uses the same flag for both ControllerPublish and NodePublish readonly fields.

Therefore, we will do the "hard answer".

PR #310, opened 2018-11-08
Add capability for `ControllerPublishVolume.readonly` field

saad-ali:
Closes https://github.com/container-storage-interface/spec/issues/293


saad-ali:
Will change this per https://github.com/container-storage-interface/spec/issues/293#issuecomment-437195734

saad-ali:
@julian-hj @jieyu per our discussion, I changed this from a comment to a new capability.

csi.proto
```diff
@@ -576,8 +576,9 @@ message ControllerPublishVolumeRequest {
   // This is a REQUIRED field.
   VolumeCapability volume_capability = 3;
 
-  // Whether to publish the volume in readonly mode. This field is
-  // REQUIRED.
+  // Whether to publish the volume in readonly mode. This field MUST be
+  // specified only if SP has PUBLISH_READONLY controller capability.
+  // This is a REQUIRED field.
   bool readonly = 4;
```
gnufied: Is making `readonly` flag optional simpler?

saad-ali: Problem with making it optional is that the CO doesn't know when to set it. If CO sets it to true and a plugin doesn't support it, what happens? Consider the case where the CO uses the same flag for both ControllerPublish and NodePublish readonly -- we don't want to get in to a case where the CO thinks a volume is readonly but it's not.

csi.proto
```diff
@@ -576,8 +576,9 @@ message ControllerPublishVolumeRequest {
   // This is a REQUIRED field.
   VolumeCapability volume_capability = 3;
 
-  // Whether to publish the volume in readonly mode. This field is
-  // REQUIRED.
+  // Whether to publish the volume in readonly mode. This field MUST be
+  // specified only if SP has PUBLISH_READONLY controller capability.
```
jieyu: `This field MUST be REQUIRED specified only if` conflicts with `This is a REQUIRED field`.

We should say that if the SP does not have PUBLISH_READONLY capability, the CO will set this field to `false`.

saad-ali:
Feedback addressed PTAL

PR #311, opened 2018-11-09
make: Strip /vXYZ from golang import path.

### LINKED ISSUE #273: Repackage generated lib/go/csi/... stubs for compat w/ go modules
golang modules disallow v0 and v1 from import paths:
* https://github.com/golang/go/wiki/Modules#why-are-major-versions-v0-v1-omitted-from-import-paths

jdef:
Fixes #273


saad-ali:
LGTM

Issue #250, opened 2018-06-14
Why is `VolumeCapability` a list in `CreateVolumeRequest`?


saad-ali:
@msau42 was asking me why `VolumeCapability` is a repeated field in `CreateVolumeRequest` and some other places:

```protobuf
repeated VolumeCapability volume_capabilities = 3;
```

I couldn't remember why we did that? When creating a volume why would we want to specify more then one capability? AFAIK both `access_type` and `access_mode` can be specified in a `VolumeCapability` object.

I vaguely recall a discussion about not knowing exactly how a volume will be consumed at provision time, but I can't recall if this would help that case.

So purpose of this bug is to answer the question: why is `VolumeCapability` a list in `CreateVolumeRequest`?

jieyu:
@saad-ali IIRC, the intention was to allow SP to provision a volume that can be published as both Mount or Block volume. Think about creating an EBS volume that can be used as both a Block volume or a Mount volume. At the creation time, the CO might not know the exact way that the volume will be published.

jdef:
If we don't already, we should probably call out that scenario in the spec,
for clarity

msau42:
If the capability passed in is Mount, is it expected for a block volume plugin to also create the filesystem?  Or is the CO expected to handle filesystem formatting?

jieyu:
@msau42 SP is eventually responsible for formatting the filesystem. When does the SP decide to format the filesystem is up to the SP. I imagine most SP will do that in NodePublishVolume, but that's not a requirement in the spec.

msau42:
I guess I'm kind of confused as a plugin writer about what I should do when given multiple VolumeCapabilities.  Do I pick one of them to support? Or do I need to provision a volume in such a way to support all of them?  There is no option in the VolumeCreateResponse to specify which VolumeCapability I chose.

msau42:
Another aspect I'm a little confused about are for volume plugins that can only support file or block (but not both).  How do I report to the CO that capability?

jieyu:
@msau42 I think the wording in the spec is pretty clear. If the SP cannot support ANY of the capability requested by the CO, an error should be returned.

```// The capabilities that the provisioned volume MUST have: the Plugin
  // MUST provision a volume that could satisfy ALL of the
  // capabilities specified in this list. The Plugin MUST assume that
  // the CO MAY use the  provisioned volume later with ANY of the
  // capabilities specified in this list. This also enables the CO to do
[...]

PR #314, opened 2018-11-09
Clarify VolumeCapability in CreateVolumeRequest

saad-ali:
Clarify why VolumeCapability in CreateVolumeRequest is a repeated list.

Closes https://github.com/container-storage-interface/spec/issues/250


saad-ali:
Fixed stray line in the PR too.

Issue #279, opened 2018-10-16
Consider renaming `Volume.attributes`


saad-ali:
Low priority, but I've had several people mention to me that attributes is a confusing name:

```
message Volume {
  map<string, string> attributes = 3;
...
```

This field should be opaque to CO. But `attributes` suggests that it contains "volume attributes". Which is not the intention.

Maybe call it `server_context` or something else?

saad-ali:
Other options: `cookie`, `plugin_attributes`, `driver_context`?
oem vs plugin: CO doesn't care between these two. 

We should also be consistent with the naming of `ControllerPublishVolumeResponse.publish_info`

`volume_context`, `publish_context`? context should be passed in everywhere
`volume_cookie`, `publish_cookie`? mutable, where as these are not.

Conclusion: leave it as is, and make documentation clearer?

saad-ali:
There are a number of these "opaque key/value bags" in the API, another one is currently called `manifests` in `GetPluginInfoResponse`. We should make sure the naming we decide on is consistent for all of these.

saad-ali:
I'll create a PR saying `...context...`, and we'll go from there.

saad-ali:
> There are a number of these "opaque key/value bags" in the API, another one is currently called manifests in GetPluginInfoResponse. We should make sure the naming we decide on is consistent for all of these.

Will not rename `manifest` unlike `volume_attributes` and `publish_info` it is not passed in on subsequent operations, so `context` doesn't make sense for that field. I don't have a better name for it, so I will leave `manifest` as is.

Issue #300, opened 2018-11-06
Clarify VolumeID name-spacing requirements


saad-ali:
A volume or node ID returned by one CSI driver may collide with a volume or node ID from another driver.

Spec should clarify that it is the responsibility of the CO to ensure that it does appropriate namespacing to handle such collisions.

jdef:
Why is it the responsibility of the SP? Because that keeps things simpler for the CO?

saad-ali:
> Why is it the responsibility of the SP? Because that keeps things simpler for the CO?

Whoops that should say CO not SP, fixed.

saad-ali:
Recommendation from @jdef and @julian-hj : `x is unique per driver, the plugin is NOT responsible for global uniqueness  `

saad-ali:
Closed with #318

PR #318, opened 2018-11-09
Rename volume `attributes`, `publish_info`, and `id`

saad-ali:
Closes https://github.com/container-storage-interface/spec/issues/279


csi.proto
```diff
@@ -351,15 +351,22 @@ message Volume {
   // subsequent calls to refer to the provisioned volume.
   string id = 2;
```
saad-ali: Feedback from @jieyu @julian-hj: change `id` to `volume_id`.

saad-ali:
@jieyu @julian-hj Feedback addressed. PTAL

saad-ali:
Added fix for https://github.com/container-storage-interface/spec/issues/300 in this PR as an additional commit since it was touching the same lines.

PR #320, opened 2018-11-09
owner: Remove mycure from approvers list

jieyu:
No longer with Docker.


jieyu:
@ddebroy can you review and approve this PR?

cc @saad-ali @julian-hj @jdef

Issue #319, opened 2018-11-09
Change all instances of v0.x in spec in prep for 1.0


saad-ali:
For example:

```
package csi.v0;
```
and
```
The Container Storage Interface (CSI) explicitly will not define, provide, or dictate in v0.1:
```

jdef:
We probably want to leave a version in the proto package name because not
all langs have the same kind of flexibility as golang.

jieyu:
I think he meant to change from v0 to v1

PR #321, opened 2018-11-11
Bump CSI package version in anticipation of 1.0 release

saad-ali:
Closes https://github.com/container-storage-interface/spec/issues/319


VERSION
```diff
@@ -1 +1 @@
-0.4.0
```
jieyu: I don't see a reason for this file anymore. Can we just remove it?

saad-ali: Opened https://github.com/container-storage-interface/spec/issues/324 to track that.

Issue #288, opened 2018-11-01
nit: scrub spec for should, must, etc. and capitalize


saad-ali:
We still have instances of the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" that are not capitalized, before 1.0 we should fix that.

Example: `  //    Plugin should ensure that multiple `CreateVolume` calls for the` in `CreateVolumeRequest`.

PR #325, opened 2018-11-12
Scrub spec for "should", "must", etc. and capitalize

saad-ali:
Fixes https://github.com/container-storage-interface/spec/issues/288

This was not a blind find and replace. I tried to look at the intention of the word and only change capitalization if the intention appeared to be inline with the keyword.


csi.proto
```diff
@@ -644,7 +644,7 @@ message ControllerPublishVolumeResponse {
   // The contents of this field SHOULD NOT contain sensitive
   // information.
   // The contents of this field SHOULD NOT be used for uniquely
-  // identifying a volume. The `id` alone should be sufficient to
+  // identifying a volume. The `id` alone SHOULD be sufficient to
```
julian-hj: `id` should be `volume_id` here, right?  I guess we missed that in the `id`->`volume_id` PR

saad-ali: Yes, good catch will fix!

PR #331, opened 2018-11-13
spec: created_at is a Timestamp.

### LINKED ISSUE #330: spec: Snapshot.created_at should use well-known Timestamp protobuf
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto

This would be a breaking change after 1.0, so if this change doesn't land before then maybe we could eventually use a `oneof` to wrap the existing int-based timestamp and a new field that uses `Timestamp`.

Ideally we'd resolve this prior to 1.0

jdef:
Fixes #330


saad-ali:
CC @xing-yang @jingxu97 for FYI

jdef:
I'm happy to make that change later tonight or tomorrow morning... if
people want it

saad-ali:
> I'm happy to make that change later tonight or tomorrow morning... if people want it

I'm fine with it too. Go for it. I'll wait for it.

xing-yang:
Thanks @jdef!  Yes, Jing and I are fine with the change.  Renaming it to creation_time sounds good.

jdef:
@julian-hj I've renamed the field - PTAL

saad-ali:
LGTM

Issue #287, opened 2018-10-31
NodePublishVolumeRequest() should define what to do as a driver when given a targetPath for a block volume


lpabon:
NodePublishVolumeRequest() should define what to do as a driver when given a targetPath for a block volume. Right now in Kubernetes-csi we are assuming it is the responsibility of the CO to create the directory where the Symlink will live, for the CO to pass the targetPath (which is the symlink name) and for the CO to create the symlink to the raw device node.

See #285

jieyu:
We need to make it clear in the spec that who (CO or SP) is responsible for creating the target_path, and the target_path will be a file for block device and a directory for volume.

jieyu:
This is related to #80

saad-ali:
Resolved in #335

PR #335, opened 2018-11-14
Clarify staging and target paths for blocks

bswartz:
The vague wording around the staging and target paths left open too
many possible implementations for raw block volumes, and implementors
ended up doing non-obvious things. This change clarifies the contract
between the CO and the SP when staging and publishing raw block volumes
on the node.


jdef:
related to #287

jdef:
I'm not crazy about the specificity of this PR. Though I understand that it *does* help to clarify things for SP authors. I'm hopeful that we can move to something more generic (in a non-breaking way) once we support something like what's been proposed in #96, where the CO has a bit more control over mount operations.

bswartz:
@jdef I'm trying to understand the problems that the solution proposed in that issue is trying to solve. The discussion seemed to argue for what we have now, which is node plugins that do the mounts themselves. The only specific use case I could find (sshfs) strongly favors the current design. What would be gained by modifying the publish interface to offer another way?

saad-ali:
@bswartz this is related to another discussion where some one proposed potentially allowing CO to control the final mount. See https://github.com/container-storage-interface/spec/issues/96#issuecomment-435147893

That would be an additive change we can consider in the future (post v1). I believe @jdef is arguing that would be a way to introduce the flexibility that is potentially lost with this PR.

spec.md
```diff
@@ -1762,11 +1762,10 @@ message NodeStageVolumeRequest {
 
   // The path to which the volume MAY be staged. It MUST be an
   // absolute path in the root filesystem of the process serving this
-  // request. The CO SHALL ensure that there is only one
-  // `staging_target_path` per volume. The CO SHALL ensure that the
-  // process serving the request has `read` and `write` permission to
-  // the path, and is able to create files or directories at the path
-  // if it does not exist.
+  // request, and MUST be a directory. The CO SHALL ensure that there
+  // is only one `staging_target_path` per volume. The CO SHALL ensure
```
jdef: multiple ways to interpret: "... path is an existing directory". Suggest re-phrasing:

The CO SHALL be responsible for creating the path if it does not exist. The CO SHALL ensure that the path is a directory.

spec.md
```diff
@@ -1904,9 +1903,11 @@ message NodePublishVolumeRequest {
...
   // absolute path in the root filesystem of the process serving this
   // request. The CO SHALL ensure uniqueness of target_path per volume.
-  // The CO SHALL ensure that the process serving the request has
-  // `read` and `write` permissions to the path, and is able to create
-  // files or directories at the path if it does not exist.
+  // The CO SHALL ensure that the path exists and that the process
+  // serving the request has `read` and `write` permissions to the path.
+  // For volumes with an access type of block, the path MUST be a file.
+  // For volumes with an access type of mount, the path MUST be a
+  // directory. The SP SHALL bind mount the volume to the target path.
```
jdef: do we really have to say "bind mount" here? Isn't it enough to say "mount"?

jdef:
> That would be an additive change we can consider in the future (post v1). I believe @jdef is arguing that would be a way to introduce the flexibility that is potentially lost with this PR.

To be clear, I don't think we lose flexibility to do something for #96 with the introduction of this PR. I was trying to say that we still have an escape hatch if we find out (as I think we will) that this PR is too restrictive for all the use cases that we want to support.

saad-ali:
Let's stay focused on this PR. We can discuss the merits of #96 on that PR.

bswartz:
@jdef Okay I think I understand. As long as we agree that this PR doesn't make the problem you want to solve any harder. The staging path stuff shouldn't affect that future enhancement because it's really just a clarification of what SPs are already depending on, and doesn't impose a new burden on COs. The publish target path stuff could all be overridden by a hypothetical future enhancement to the node publish call.

spec.md
bswartz: @jdef The problem I think many people have is that the word "mount" traditionally implies what happens when you add a new filesystem into the VFS at some directory. For block volumes, we've opted to actually bind mount a single device file. There are lots of alternatives we could have chosen, including symlinks or returning a string. Presumably there is some advantage to the bind mount approach (I wasn't around when this decision was made) or perhaps the spec was left so vague that Kubernetes just chose to use bind mounts rather than the other alternatives and that became the de-facto way to do it. This kind of open-endedness is poison for specifications because it makes it very easy for 2 different implementation that both conform to the spec to not interoperate with eachother.

I could change the wording to make it clear that the bind mount is only a requirement for block access types, and that an ordinary mount is find for mount (filesystem) access types.

spec.md
bswartz: Thanks, I will use your language here.

spec.md
jdef: OK, I understand where you're coming from. Again, this is where the proposal gets a bit too specific w/ it's *requirements* of a plugin. I'd like to see just a tiny bit of flexibility here.

Can you replace "SHALL bind mount" with "SHOULD bind mount"?

jdef: FWIW there's multiple ways to interpret "bind mount" so I'm not sure that there's too much actually gained by keeping the word "bind".

Fun read: https://unix.stackexchange.com/questions/198590/what-is-a-bind-mount

jdef: I'm still having a hard time getting behind this change request for `NodePublishVolume`. If a plugin doesn't support node-stage and simply wants to `mknod` a block device at `target_path` then this proposal blocks that behavior. Forcing a plugin to `mknod` elsewhere and then to "bind mount" it to a file that the CO is creating here seems like a mistake.

I think we should keep more neutral wording here. The CO SHALL NOT create a file system object at `target_path`. Plugin implementation recommendations should be made elsewhere in the spec and *should* avoid *requirement* wording like SHALL, and MUST.

bswartz: The word SHOULD implies to me that you can not do something and then there are still some scenarios where that can lead to a good result. SPs must do something at the target path to ensure a good result. We owe it to SP implementors to either enumerate all the valid things an SP can do or describe the end result in such a way that at least a few of the valid approaches are obvious.

bswartz: I'm with you on the fact that there are valid alternatives that can work. Doing a mknod at the target path makes perfect sense to me, but then we have to change the wording to say the CO MUST NOT create anything there, which is is the opposite of what Kubernetes has done and will require changing Kubernetes again (not impossible, just annoying).

bswartz: I could go either way on this, @saad-ali do you feel strongly? The one outcome I don't want is to leave it unspecified which side should create the file/device because then a de-facto standard of "whatever works with kubernetes" will emerge.

bswartz:
@julian-hj For filesystem (mount access type) volumes I absolutely agree, and will change the wording. For block access type volumes, I'm okay loosening the language as long as it's still clear what the CO should or should not create, so it's clear what's left for the SP to do.

bswartz:
@lpabon I'm also curious about your opinion on the CO not creating the "file" that gets bind mounted and instead requiring the SP to create a file before bind mounting the path, if it wants to bind mount rather than doing something like mknod. In this case we could simply require that the CO ensures the parent directory exists. I can't imagine any problems with that approach, but @mkimura will need to modify his kubernetes patch to accomodate that.

mkimura:
You’re using the wrong tag

saad-ali:
@mkimuram not `mkimura`

spec.md
saad-ali: I have a slight preference to leave the spec lose enough for alternate implementations.

> We owe it to SP implementors to either enumerate all the valid things an SP can do or describe the end result in such a way that at least a few of the valid approaches are obvious.

I'd prefer the later.

bswartz:
Apologies for misspelling the name

mkimuram:
@bswartz @lpabon 

> I'm also curious about your opinion on the CO not creating the "file" that gets bind mounted and instead requiring the SP to create a file before bind mounting the path, if it wants to bind mount rather than doing something like mknod.   

If you are talking about ```NodePublishVolume```/```NodeUnpublishVolume```, won't this change affect the assumption that there is no difference in ```NodeUnpublishVolume``` logic between block and mount access type? I understand that we originally assume that ```NodeUnpublishVolume``` requires just unmounting from target path in both case. However, if ```NodeUnpublishVolume``` is required to do something different from unmount, it might need information on which access type is requested, which is not passed to ```NodeUnpublishVolume```, to decide what to do for the path. (The decision to choose other than mount/unmount is up to SP, so it could be at their own risk, though.)

bswartz:
@mkimuram SPs only need to do something different if they create different things at the publish stage. If they just create a mounted directory or a bind mounted file, then blindly calling umount on the target path would be relatively safe.
If on the other hand the stuff that gets created can vary significantly, then the SP would need to detect what got created by calling stat() on the target path at unpublish time or use an external mechanism to remember what kind of volume it is.

spec.md
```diff
@@ -1904,9 +1905,14 @@ message NodePublishVolumeRequest {
...
-  // `read` and `write` permissions to the path, and is able to create
-  // files or directories at the path if it does not exist.
+  // The CO SHALL ensure that the parent directory of this path exists
+  // and that the process serving the request has `read` and `write`
+  // permissions to that parent directory.
+  // For volumes with an access type of block, the SP SHALL place the
+  // block device at target_path.
+  // For volumes with an access type of mount, the SP SHALL place the
+  // mounted directory at target_path.
+  // Creation of target_path is the responsibility of the SP.
```
chhsia0: Nothing needs to be changed here, just leaving a note.

Since the spec never discuss the relationship between volume states (`NODE_READY`, `VOL_READY` and `PUBLISHED`) and node reboots, it could be argued that a `PUBLISHED` volume could be in one of either states after a reboot:
1. `PUBLISHED` since no `NodeUnpublishVolume` has been called. In this case, the CO might call `NodeUnpublishVolume` -> `NodeUnstageVolume` -> `NodeStageVolume` -> `NodePublishVolume` to recover the volume. As a result, the plugin must be able to handle the first two calls when the mount point is missing in the mount table.
2. `NODE_READY` since the mount table has been reset after reboot. In this case, the CO might directly call `NodeStageVolume` and `NodePublishVolume` to recover the volume. As a result, the plugin must be able to deal with the fact that `target_path` might have been created by itself in a previous run.

cc @jdef

jdef: Thanks for leaving this comment. At some point I think we should consider adding a "Notes for plugin writers" section to capture things along these lines. If people aren't comfortable adding to the spec.md file then we could add a separate "notes-for-plugin-writers.md" file or something that gives additional context, examples, and other things to consider when writing plugins.

Issue #346, opened 2019-01-25
Undefined behavior for DeleteVolume with snapshots 


Akrog:
I don't see the spec defining the behavior of a DeleteVolume call when the volume being deleted has snapshots.

In most backends deleting a volume with snapshots is not directly possible, one must delete the snapshots first.  There are also backends that cascade delete the snapshots (after confirmation) when deleting such volume.

I see 3 possible behaviors, and we should be explicit about which one we expect in the spec:

1. Fail the delete operation
2. Cascade delete the snapshots
3. "Not my problem": The CSI plugin will have to mark the volume as deleted and not display it when listing or allow its use as a volume source or the source of a snapshot, and once all its existing snapshots have been deleted the volume itself should be deleted.

From a storage driver perspective the first option is the most reasonable, and from a CO's perspective the third option is the most convenient, as it decouples volumes and snapshots, which makes sense for cases where snapshots are really off-premise backups that are uploaded as a post-processing operation and the cut snapshot is deleted after the post-processing.

mattsmithdatera:
My vote is for consistency between plugins in this behavior.  Lets have both 1 & 2 by specifying a boolean parameter in `DeleteVolumeRequest` called `purgeSnapshots`.

If `purgeSnapshots` is False, then the delete operation should fail if there are snapshots currently associated with the volume.  If `purgeSnapshots` is True, then volume deletion and snapshot deletion should occur.

As part of `DeleteVolumeResponse` if `purgeSnapshots` is True we should have a parameter that allows for returning the snapshotIDs of all snapshots purged by the deletion of the volume for CO book-keeping.

Akrog:
I like that idea, although I'm not sure if it'd fit all cases, as there was much discussion around remotely storing the data during the review of the snapshot feature.
 
What is the expected behavior of v1.0?  Because if csi-sanity's behavior is a good reference, and I hope it is because I've used it for my CSI plugin, then it's applying option #3.

PR #347, opened 2019-01-25
Define DeleteVolume behavior with snapshots

Akrog:
Spec is not clear on what should happen when on a DeleteVolume call when
the volume has snapshots.

This patch clarifies the situation by explicitly mentioning that the
operation should complete and the snapshots should still be operational.

Closes: #346

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1131,6 +1131,12 @@ This RPC will be called by the CO to deprovision a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified `volume_id` does not exist or the artifacts associated with the volume do not exist anymore, the Plugin MUST reply `0 OK`.
 
+CSI plugins SHOULD treat volumes independent from their snapshots.
+
+If the Controller Plugin supports deleting a volume without affecting its existing snapshots, then these snapshots MUST still be fully operational and acceptable as sources for new volumes as well as appear on `ListSnapshot` calls once the volume has been deleted.
+
+When a Controller Plugin does not support deleting a volume without affecting its existing snapshots, then the volume MUST NOT be altered in any way by the request and the operation must return the `FAILED_PRECONDITION` error code and MAY include meaningful human-readable information in the `status.message` field.
+
 ```protobuf
 message DeleteVolumeRequest {
   // The ID of the volume to be deprovisioned.
@@ -1156,7 +1162,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
-| Volume in use | 9 FAILED_PRECONDITION | Indicates that the volume corresponding to the specified `volume_id` could not be deleted because it is in use by another resource. | Caller SHOULD ensure that there are no other resources using the volume, and then retry with exponential back off. |
+| Volume in use | 9 FAILED_PRECONDITION | Indicates that the volume corresponding to the specified `volume_id` could not be deleted because it is in use by another resource or has snapshots and the plugin doesn't treat them as independent entities. | Caller SHOULD ensure that there are no other resources using the volume and that it has no snapshots, and then retry with exponential back off. |
 
 
 #### `ControllerPublishVolume`


spec.md
```diff
@@ -1131,6 +1131,8 @@ This RPC will be called by the CO to deprovision a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified `volume_id` does not exist or the artifacts associated with the volume do not exist anymore, the Plugin MUST reply `0 OK`.
 
+CSI plugins MUST treat volumes independent from their snapshots.  Controller Plugin MUST support deleting a volume without affecting its existing snapshots, which MUST still be fully operational and acceptable as sources for new volumes as well as appear on `ListSnapshot` calls.
```
jdef: I think this could be too onerous for some plugins to support, especially if they're not maintaining their own state store independent of any internal store provided by the "real" backend.

I'd rather let the plugin decide how to do this:

1. Fail the volume-delete RPC because it's not allowed there are snapshots.
2. Honor the volume-delete RPC and maintain the existing snapshots (because they're still usable).
3. Honor the volume-delete RPC and remove all linked snapshots.

If there's an ongoing snapshot-related operation in progress (e.g. creating a snapshot from said volume, or creating a volume from a snapshot that's linked to the volume-to-be-removed) then we have error codes a plugin can use to indicate a conflict with an ongoing operation (`ABORTED`).

If (1) is the case then there's already a documented error code for delete's that fail because a resource is in-use (`FAILED_PRECONDITION`).

If the plugin is able to execute (3) without errors, then it's up to the CO to periodically reconcile its internal state w/ that of the plugin in order to account for snapshots that may have disappeared.

saad-ali: I agree that we should let the SP decide but this is a good suggestion. So how about `/s/MUST/SHOULD/` in this change to make it a recommendation rather then a requirement?

Akrog: I'm updating the patch to change it to `SHOULD` and also make it contemplate @jdef 's first scenario.  This way plugins can either support deleting volumes independently or fail.  We can leave support for cascade deletion of snapshots as a new feature and leave this patch to just clarify the current expected behavior.

spec.md
```diff
@@ -1131,6 +1131,8 @@ This RPC will be called by the CO to deprovision a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified `volume_id` does not exist or the artifacts associated with the volume do not exist anymore, the Plugin MUST reply `0 OK`.
 
+CSI plugins SHOULD treat volumes independent from their snapshots.  If the Controller Plugin supports deleting a volume without affecting its existing snapshots, then these snapshots MUST still be fully operational and acceptable as sources for new volumes as well as appear on `ListSnapshot` calls.
```
jdef: Thanks for the revised language, though I recommend s/MUST/SHOULD/ in the 2nd sentence. We're saying this behavior is optional, but recommended - so the use of MUST only makes it more confusing: is this required or optional behavior?

nit: one sentence per line, please.

Akrog: I forgot about the one line per sentence, sorry.

The reason for using MUST in the second sentence is that it is a conditional sentence, and I believe it to be correct.

**If** a driver does support deleting a volume independently of the snapshots then it **MUST** (not should, this is a hard requirement) allow using those snapshots normally.

If I change the MUST to SHOULD, the specs end up saying that it's OK to have a driver that allows deleting a volume with snapshots and leave its snapshots in a state that they cannot be normally used.  Which is not what we want to say.

I will add another sentence explaining the case of the controller not supporting deleting volumes independently of their snapshots to make it explicit that both options are possible, and that's why we had the `SHOULD` in the first sentence.

Thanks for the review.

jdef:
Thanks for the clarifying language. I wonder if this should be considered a breaking change?

This could be construed as more restrictive behavior vs. what was called out in the original specification.

Akrog:
I think it's not really a breaking change, it's mostly a clarification with a preferred behavior.

Existing v1.0 plugins will have implemented the delete operation in one of those two ways, and it doesn't matter which one they have, because they will conform to the spec regardless.

jdef:
> Existing v1.0 plugins will have implemented the delete operation in one of those two ways, and it doesn't matter which one they have, because they will conform to the spec regardless.

ok, sounds good

zhucan:
maybe, whether snapshots is volumes independent from their snapshots， we can set a flag to volumesnapshotclass？

zhucan:
when deleting volumes , we can judge wether volumes independent from their snapshots from the flag?

Akrog:
> maybe, whether volumes independent from their snapshots， we can set a flag to volumesnapshotclass？
> when deleting volumes , we can judge wether volumes independent from their snapshots from the flag?

From CSI's perspective there's no such thing as the VolumeSnapshotClass.  That is a CO (k8s) specific detail.

From the CSI perspective what we discussed was that a CSI plugins could report whether the plugin supports this capability or not.  But decided to let drivers decide and be the CO's decision.

It should be possible to use a flag in the VolumeSnapshotClass to tell k8s whether a volume with snapshots can be deleted or not, but adding this feature should be brought up with the Kubernetes community.

zhucan:
But our storage support volumes independent from snapshots and volumes not independent      from snapshots发自我的iPhone------------------ Original ------------------From: Gorka Eguileor  <notifications@github.com>Date: Thu,Apr 25,2019 6:48 PMTo: container-storage-interface/spec <spec@noreply.github.com>Cc: zhucan <zhucan.k8s@gmail.com>, Comment <comment@noreply.github.com>Subject: Re: [container-storage-interface/spec] Define DeleteVolume behaviorwith snapshots (#347)
maybe, whether volumes independent from their snapshots， we can set a flag to volumesnapshotclass？
when deleting volumes , we can judge wether volumes independent from their snapshots from the flag?

From CSI's perspective there's no such thing as the VolumeSnapshotClass.  That is a CO (k8s) specific detail.
From the CSI perspective what we discussed was that a CSI plugins could report whether the plugin supports this capability or not.  But decided to let drivers decide and be the CO's decision.
It should be possible to use a flag in the VolumeSnapshotClass to tell k8s whether a volume with snapshots can be deleted or not, but adding this feature should be brought up with the Kubernetes community.

—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/container-storage-interface/spec","title":"container-storage-interface/spec","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/container-storage-interface/spec"}},"updates":{"snippets":[{"icon":"PERSON","message":"@Akrog in #347: \u003e maybe, whether volumes independent from their snapshots， we can set a flag to volumesnapshotclass？\r\n\u003e when deleting volumes , we can judge wether volumes independent from their snapshots from the flag?\r\n\r\nFrom CSI's perspective there's no such thing as the VolumeSnapshotClass.  That is a CO (k8s) specific detail.\r\n\r\nFrom the CSI perspective what we discussed was that a CSI plugins could report whether the plugin supports this capability or not.  But decided to let drivers decide and be the CO's decision.\r\n\r\nIt should be possible to use a flag in the VolumeSnapshotClass to tell k8s whether a volume with snapshots can be deleted or not, but adding this feature should be brought up with the Kubernetes community."}],"action":{"name":"View Pull Request","url":"https://github.com/container-storage-interface/spec/pull/347#issuecomment-486622930"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/container-storage-interface/spec/pull/347#issuecomment-486622930",
"url": "https://github.com/container-storage-interface/spec/pull/347#issuecomment-486622930",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

Akrog:
Then your storage must succeed for independent volumes and return `FAILED_PRECONDITION` error for the non-independent ones.

zhucan:
Ok,  get it发自我的iPhone------------------ Original ------------------From: Gorka Eguileor  <notifications@github.com>Date: Thu,Apr 25,2019 7:08 PMTo: container-storage-interface/spec <spec@noreply.github.com>Cc: zhucan <zhucan.k8s@gmail.com>, Comment <comment@noreply.github.com>Subject: Re: [container-storage-interface/spec] Define DeleteVolume behaviorwith snapshots (#347)Then your storage must succeed for independent volumes and return FAILED_PRECONDITION error for the non-independent ones.

—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/container-storage-interface/spec","title":"container-storage-interface/spec","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/container-storage-interface/spec"}},"updates":{"snippets":[{"icon":"PERSON","message":"@Akrog in #347: Then your storage must succeed for independent volumes and return `FAILED_PRECONDITION` error for the non-independent ones."}],"action":{"name":"View Pull Request","url":"https://github.com/container-storage-interface/spec/pull/347#issuecomment-486628574"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/container-storage-interface/spec/pull/347#issuecomment-486628574",
"url": "https://github.com/container-storage-interface/spec/pull/347#issuecomment-486628574",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

Issue #355, opened 2019-03-05
Alpha Beta GA designation for new capabilities


saad-ali:
**Problem**

Now that CSI is 1.0, we can no longer make backwards compatibility breaking changes. But we want to continue to add new functionality to CSI, and it is almost impossible to propose an API and have it be "perfect" in its initial release (which is why CSI remained in 0.x for a year).

**Goal**

We should define a process for introducing new official APIs as alpha then promote to beta and then GA.

**Proposal**

This can be as simple as a comment on a capability designating it alpha, beta, or GA. The tricky part is how alpha APIs can change without breaking clients. So we may need to say that fields and methods can be renamed/deprecated but never removed. Alpha features can be renamed or deprecated in any release. Beta features can be renamed or deprecated after 6 months or 2 releases (whichever is longer). And GA features can only be renamed or deprecated during major releases (e.g. between CSI v1.x to v2.x).

saad-ali:
Proposal from @jieyu -- maintain multiple proto files: GA, alpha.
* Initially they all start off as identical.
* New functionality is added to alpha, and once graduated moves to the other files.
* Guarantees for each file will differ as mentioned above.
* Features must be added to alpha first. When they graduate, both files must be identical for that feature.
* CI should be added to protect from making changes to mainline without making change in alpha as well.
* Cons:
  * This may conflict with go module semantic versioning requirements -- if v1.1 has an alpha and v1.2 has a backwards in compatible version of alpha.
    * We should just document that very explicitly.

gnufied:
I was thinking that we should have a csi-extension repo that possibly vendors-in the main CSI spec and then adds some extension on top of it.  Some projects may choose to implement csi-extension spec and some other may adopt stable APIs. 

The tricky thing perhaps will be most COs will only want to adopt CSI stable/main and even if a driver has support for csi-extension spec it may not be used immediately if CO does not support it. 

Once changes in csi-extension become stable, they can be moved off to main repo.

saad-ali:
Simpler alternative from @julian-hj during meeting:

* Accept that we will accumulate the cruft.
* Upon deprecation: can mark field numbers as reserved.
* Protobuf extension for alpha to programmatically detect if a field is alpha or not
* Will make bar higher for adding alpha features.

What about RPCs? 
* Everyone has to implement them, and when deprecated they would remain optional to implement.

What about wild and crazy ideas?
* Forks

pohly:
> What about RPCs?
[...]

PR #365, opened 2019-04-04
[[--DO-NOT-MERGE--]] Mock ups for fake, experimental alpha APIs.

jdef:
xref #355


spec.md
```diff
@@ -280,10 +280,45 @@ import "google/protobuf/wrappers.proto";
...
+  // releases.
+  bool alpha_field = 1060;
+}
+extend google.protobuf.MessageOptions {
+  // Indicates that this message is OPTIONAL and part of an experimental
+  // API that may be deprecated and eventually removed between minor
+  // releases.
+  bool alpha_message = 1060;
+}
+extend google.protobuf.MethodOptions {
```
saad-ali: Do we want to allow this?

My understanding after the call was that we allow new fields in the main spec on existing RPCs with a experimental tag. But we force new RPCs in to a separate new experimental service (potentially in a different file).

csi.proto
```diff
@@ -78,17 +118,29 @@ service Node {
...
   rpc NodeGetInfo (NodeGetInfoRequest)
     returns (NodeGetInfoResponse) {}
 }
+
+service ObjectStore {
+  option (alpha_service) = true;
+
+  rpc ObjectStoreRead(ObjectStoreReadRequest)
+    returns (ObjectStoreReadResponse) {
+      option (alpha_method) = true;
```
jdef: it's arguable that marking the method as alpha is redundant because the entire service is already alpha.

Issue #370, opened 2019-06-27
Secrets field in ListSnapshotsRequest


hakanmemisoglu:
Right now, CSI specification defines a `secrets` field in `CreateSnapshotRequest` to pass credentials around. 

However, `ListSnapshotsRequest` does not have this field, meanwhile listing snapshots might require some credentials with read access to work properly.

I think we need an optional `secrets` field in `ListSnapshotRequest`.

j-griffith:
Do you have an example use case of why list snapshots would need/want secrets?  We have the parameter for create volume and snapshot as it's pretty common to need special credentials to create resources on a plugin.  I'm not too familiar with restrictive list operations though.  Would it be possible to use a reference in the plugin from object creation?

If this is something that's needed it has to be considered for ListVolumesRequest as well.

travisghansen:
I'm currently planning/designing a csi driver for FreeNAS (and likely various other zfs based storage systems). In an effort to remain as stateless as possible (hopefully totally stateless) I think secrets should be passed not only to this method but any other controller methods that don't currently receive them.

I need to know *which* appliance I'm dealing with and how to communicate with it (ie: creds).

`GetCapacityRequest` appears to be missing `secrets`.
`ListVolumesRequest` appears to be missing `secrets`.

`ControllerGetCapabilitiesRequest` is probably the only exception.

hakanmemisoglu:
Ceph CSI driver is another one. It doesn't support ListSnapshots because there is no easy way to pass cluster credentials to the method since they switched to a stateless implementation.

xing-yang:
@hakanmemisoglu Do you also need secrets to support ListVolumes for the Ceph CSI driver?  If secret is needed to list snapshots, it should be needed to list volumes as well.

xing-yang:
I thought only write operations need secrets.  List snapshots/volumes are read operations.  I didn't expect that they need secrets as well.

hakanmemisoglu:
@xing-yang, yes you are right. We can include `ListVolumes` too.

I think all of the read operations might need read credentials depending on the SP. It includes `ListVolumes` as you noted and maybe others, like `GetCapacity` as @travisghansen commented.

shay-berman:
Yes, also IBM storage requires secret to list snapshot, like every csi controller operations that communicates with the storage.

j-griffith:
[...]

PR #372, opened 2019-08-05
Adds secrets field to ListSnapshotsRequest

hakanmemisoglu:
There is no way to directly pass credentials to ListSnapshots call.
This change will add a field required to pass credentials by the
caller.


hakanmemisoglu:
#370

hakanmemisoglu:
@saad-ali Do we need to update golang version in Travis? (or enable go module)

spec.md
```diff
@@ -1737,6 +1737,12 @@ message ListSnapshotsRequest {
   // and will not block if the snapshot is being processed after
   // it is cut.
   string snapshot_id = 4;
+
+  // Secrets necessary for listing snapshots.
+  // This field is OPTIONAL. It can be used to pass secrets
+  // required to access some subset of the snapshots in the
```
xing-yang: What do you mean by "some subset of the snapshots"?  How is a subset defined?

hakanmemisoglu: How about `... access the snapshots...`?

jieyu:
Yeah, we should bump the golang in CI. It's still using 1.10.4

hakanmemisoglu:
@jieyu Sure. I will push a separate PR.

spec.md
saad-ali: How about:

```
  // Secrets required by plugin to complete ListSnapshot request.
  // This field is OPTIONAL. Refer to the `Secrets Requirements`
  // section on how to use this field.
```

hakanmemisoglu:
@saad-ali CI passed.

PR #374, opened 2019-08-07
Add PublishedNodes field to ListVolumes Response

davidz627:
Helps the CO reconcile the actual state when the volume may have been Unpublished from the node out of band from the CO.

/assign @saad-ali 
/cc @xing-yang @gnufied @julian-hj 

PTAL Discussion: https://docs.google.com/document/d/1-oiNg5V_GtS_JBAEViVBhZ3BYVFlbSz70hreyaD7c5Y/edit#heading=h.z24362cngqjs


saad-ali:
CC @julian-hj @ddebroy

jieyu:
Can you make sure the protobuf file is also updated accordingly. There's a makefile target for the generation.

davidz627:
How do we feel about adding something like: 
```
List nodes published MAY return volumes not published or created by the driver. The CO MUST be resilient to that
```

jieyu:
@davidz627 that sounds reasonable to me

csi.proto
```diff
@@ -802,6 +802,14 @@ message ListVolumesRequest {
...
   message Entry {
     Volume volume = 1;
+
+    // A list of all `node_id` of nodes that the volume in this entry
+    // is controller published on.
+    // This is a REQUIRED field if the SP has the
+    // LIST_VOLUMES_PUBLISHED_NODES controller capability. If the SP
+    // does not have the LIST_VOLUMES_PUBLISHED_NODES capability this
+    // field MUST be empty.
+    repeated string published_nodes = 2;
```
gnufied: Should this list be here or part of `Volume` type? I think it belongs to `Volume` type rather than here. My reasoning for that is - if another request type possibly needs this field it can use it too.

davidz627: Thanks for bringing this up! We had originally discussed this as well but determined `Volume` contains `Spec`-type information that is an unchanging description of the object, whereas `published_nodes` doesn't really fit in there as a dynamic value.

There are also no other messages that use `Volume` besides the `CreateVolume` responss - and in that case the `published_nodes` should always be nil since the volume was just created

davidz627:
@saad-ali @jieyu resolved comments, checks passing, improved wording, PTAL

davidz627:
> One (tangential) q I have around this: Is there a reason NodeGetInfoResponse does not report the list of volumes that has been published to the node?

I don't think that there's a particular reason it doesn't. But for this proposal IMO it is more efficient to have the one list volumes call give back all volumes and where they're attached. Also for some reason it states specifically: `The result of this call will be used by CO in ControllerPublishVolume.` which is not what we'll be using it for.

csi.proto
gnufied: ack. It seems reasonable.

csi.proto
```diff
@@ -884,6 +894,10 @@ message ControllerServiceCapability {
 
       // See VolumeExpansion for details.
       EXPAND_VOLUME = 9;
+
+      // Indicates the SP supports the
+      // ListVolumesResponse.entry.published_nodes field
+      LIST_VOLUMES_PUBLISHED_NODES = 10;
```
gnufied: Did we want a capability for this one? Seems like a unnecessary knob.

davidz627: must be optional for backwards compatibility of the spec. Otherwise anyone implementing `ListVolumes` will be forced to also implement `PUBLISHED_NODES`.

Callers can then tell the difference between "no published nodes because capability not supported" and "no published nodes because the volume is not published on any nodes"

csi.proto
```diff
@@ -802,6 +802,16 @@ message ListVolumesRequest {
...
+
+    // A list of all `node_id` of nodes that the volume in this entry
+    // is controller published on.
+    // This is a REQUIRED field if the SP has the
+    // LIST_VOLUMES_PUBLISHED_NODES controller capability. If the SP
+    // does not have the LIST_VOLUMES_PUBLISHED_NODES capability this
+    // field MUST be empty.
+    // published_nodes MAY include volumes not published or
+    // created by the driver. The CO MUST be resilient to that
+    repeated string published_nodes = 2;
```
gnufied: The definition of REQUIRED from specs read:

> * A `repeated` or `map` field listed as `REQUIRED` MUST contain at least 1 element.

What happens when volume being listed is not published anywhere?

davidz627: amended to be optional with more verbose wording

gnufied: Okay, so if this is optional - why do we need to introduce a capability for this?

msau42: So that if a driver doesn't implement it, we don't assume that the volume is not published

gnufied: ack.

spec.md
```diff
@@ -1403,6 +1403,18 @@ message ListVolumesRequest {
...
+    // is controller published on.
+    // This field is OPTIONAL. If it is not specified and the SP has
+    // the LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO
+    // MAY assume the volume is not controller published to any nodes.
+    // If the field is not specified and the SP does not have the
+    // LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO MUST
+    // not interpret this field.
+    // published_nodes MAY include volumes not published or
+    // created by the driver. The CO MUST be resilient to that
+    repeated string published_nodes = 2;
```
gnufied: minor nit - missing full stop at the end of last comment.

gnufied:
LGTM to me apart from the minor nit above.

csi.proto
```diff
@@ -802,6 +802,18 @@ message ListVolumesRequest {
 message ListVolumesResponse {
   message Entry {
     Volume volume = 1;
+
+    // A list of all `node_id` of nodes that the volume in this entry
+    // is controller published on.
```
jdef: is a controller-published state more special, in some way, as compared to a node-published state? or node-staged state? reporting additional state here (in the list vols call) is nice, but i'm wondering if we should wrap this in a message type instead so that we have a nicer way to report additional state (if needed) going forward. note the use of wrapper BoolValue types that allow the CO to differentiate between "empty" (SP doesn't know/doesn't report) and `false`.

for example:
```proto
  message PublishContext {
    string node_id = 1; // REQUIRED; reporting *only* a node_id field (and nothing else) doesn't tell the CO anything (and SHOULD NOT happen in practice).
    BoolValue controller_published = 2; // OPTIONAL

    // for future consideration?
    //
    // BoolValue node_staged = 3; // OPTIONAL
    // BoolValue node_published = 4; // OPTIONAL
    // string staging_target_path = 5; // OPTIONAL
    // string target_path = 6; // OPTIONAL
  }

  ...

  // N/A if LIST_VOLUMES_PUBLISH_CTX is not a specified capability.
  repeated PublishContext publish_context = 2;
```

davidz627: I think wrapping in a message type is a great idea in case we want to bundle more information about the Volume in a return in the future. However, I'm not sure that other node related state will be considered in the future in this controller level call. What I would like to propose is actually wrapping the `node_ids` in a `VolumeStatus` message that can contain alternate mutable information about the volume in the future. WDYT?

csi.proto
```diff
@@ -802,6 +802,18 @@ message ListVolumesRequest {
...
+    // This field is OPTIONAL. If it is not specified and the SP has
+    // the LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO
+    // MAY assume the volume is not controller published to any nodes.
+    // If the field is not specified and the SP does not have the
+    // LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO MUST
+    // not interpret this field.
+    // published_nodes MAY include volumes not published or
```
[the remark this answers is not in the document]
davidz627: updated, PTAL and resolve if it makes more sense now

csi.proto
saad-ali: Since `CreateVolume` is a controller operation, it can only know the status of `ControllerPublish` not `NodeStage` or `NodePublish`.

+1 wrapping does seem like a good idea. David's current iteration proposing `VolumeStatus` seems reasonable to me.

@jdef @jieyu @julian-hj @ddebroy thoughts?

jieyu: Yeah, this looks good to me.

PR #375, opened 2019-08-07
Clarify that plugin may return OK for ControllerUnpublish if node or volume not found

### LINKED ISSUE #373: Clarify NOT_FOUND returned from ControllerUnpublishVolumeRequest
The spec says that `NOT_FOUND` means that a volume (or node) corresponding to the specified `volume_id` (`node_id`) does not exist.

But it does not say what does it mean to CO. When CO is sure that it sent the right IDs and something else must have deleted the volume or node, can CO assume that the volume was detached from the node automatically by SP when the volume/node was deleted? In this case, returning `OK` would have the same meaning - the volume has been unpublished and CO does not need to do anything.

CO has no way how to check / fix attachment of deleted volumes (or nodes), so it cannot do any cleanup when `NOT_FOUND` is returned.

davidz627:
This clarifies the plugin is allowed to return `OK` if because of a missing node or volume the plugin is sure that the volume is detached from that node.

Context: https://github.com/kubernetes-csi/external-attacher/pull/165

Fixes #373 

Discussion at CSI Community Meeting Notes here: https://docs.google.com/document/d/1-oiNg5V_GtS_JBAEViVBhZ3BYVFlbSz70hreyaD7c5Y/edit#heading=h.z24362cngqjs

/assign @saad-ali @jdef @julian-hj @jieyu 
/cc @gnufied @jsafrane @msau42

```release-note
Clarifies the plugin is allowed to return `OK` for `ControllerUnpublishResponse` if the specified node is or volume no longer exist and the plugin is sure that the volume is detached from that node.
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1257,6 +1257,7 @@ This RPC is typically called by the CO when the workload using the volume is bei
 
 This operation MUST be idempotent.
 If the volume corresponding to the `volume_id` is not attached to the node corresponding to the `node_id`, the Plugin MUST reply `0 OK`.
+If the volume corresponding to the `volume_id` or the node corresponding to `node_id` cannot be found by the Plugin and the volume can be safely regarded as ControllerUnpublished from the node, the plugin SHOULD return `0 OK`.
 If this operation failed, or the CO does not know if the operation failed or not, it can choose to call `ControllerUnpublishVolume` again.
 
 ```protobuf
@@ -1292,8 +1293,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
-| Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
-| Node does not exist | 5 NOT_FOUND | Indicates that a node corresponding to the specified `node_id` does not exist. | Caller MUST verify that the `node_id` is correct and that the node is available and has not been terminated or deleted before retrying with exponential backoff. |
+| Volume does not exist and volume not assumed ControllerUnpublished from node | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist and is not assumed to be ControllerUnpublished from node corresponding to the specified `node_id`. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
+| Node does not exist and volume not assumed ControllerUnpublished from node  | 5 NOT_FOUND | Indicates that a node corresponding to the specified `node_id` does not exist and the volume corresponding to the specified `volume_id` is not assumed to be ControllerUnpublished from node. | Caller MUST verify that the `node_id` is correct and that the node is available and has not been terminated or deleted before retrying with exponential backoff. |
 
 
 #### `ValidateVolumeCapabilities`


spec.md
```diff
@@ -1257,6 +1257,9 @@ This RPC is typically called by the CO when the workload using the volume is bei
 
 This operation MUST be idempotent.
 If the volume corresponding to the `volume_id` is not attached to the node corresponding to the `node_id`, the Plugin MUST reply `0 OK`.
+If the volume corresponding to `volume_id` or the node corresponding to
+`node_id` cannot be found by the Plugin and the volume can be safely regarded as
+ControllerUnpublished from the node, the plugin MAY return `0 OK`.
```
saad-ali: Does the `ControllerUnpublishVolume Errors` section below need to be updated at all?

saad-ali:
CC @julian-hj @ddebroy

spec.md
davidz627: added a conditional to the two `NotFound` errors

spec.md
```diff
@@ -1257,6 +1257,9 @@ This RPC is typically called by the CO when the workload using the volume is bei
 
 This operation MUST be idempotent.
 If the volume corresponding to the `volume_id` is not attached to the node corresponding to the `node_id`, the Plugin MUST reply `0 OK`.
+If the volume corresponding to `volume_id` or the node corresponding to
```
jieyu: nits: please follow style guide, and use one sentence per line.

spec.md
jieyu: Why `MAY` here? Should that be `MUST`?

davidz627: I wanted to leave room for the plugin to still return `NOT_FOUND` if it is "not sure" whether the volume is still published to the node or not. Both results should be acceptable as returning `NOT_FOUND` should result in the same behavior as currently defined (without this PR)

jieyu: OK, so the reason for not forcing SP to return `0 OK` here is for backward compatibility concern, but we still prefer SP to return `0 OK`?

In that case, i'd say we use `SHOULD`.

spec.md
```diff
@@ -1257,6 +1257,7 @@ This RPC is typically called by the CO when the workload using the volume is bei
 
 This operation MUST be idempotent.
 If the volume corresponding to the `volume_id` is not attached to the node corresponding to the `node_id`, the Plugin MUST reply `0 OK`.
+If the volume corresponding to `volume_id` or the node corresponding to `node_id` cannot be found by the Plugin and the volume can be safely regarded as ControllerUnpublished from the node, the plugin SHOULD return `0 OK`.
```
misterikkit: minor nit: line 1259 uses "the `volume_id`" and line 1260 uses "`volume_id`"

spec.md
```diff
@@ -1292,8 +1293,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
-| Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
-| Node does not exist | 5 NOT_FOUND | Indicates that a node corresponding to the specified `node_id` does not exist. | Caller MUST verify that the `node_id` is correct and that the node is available and has not been terminated or deleted before retrying with exponential backoff. |
+| Volume does not exist and volume not assumed ControllerUnpublished from node | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist and volume not assumed ControllerUnpublished from node. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
```
misterikkit: Could we get a little more verbose in the "Description" column? Specifically, how about this?

> Indicates that a volume corresponding to the specified `volume_id` does not exist and is not assumed to be ControllerUnpublished from the node. 

or with some emphasis,

> Indicates that a volume corresponding to the specified `volume_id` does not exist **and is not assumed to be ControllerUnpublished from the node**.

jingxu97: I think the current external_attacher is also not implementing Recovery Behavior which a "MUST" in the spec.

msau42: external-attacher I think is implementing the recovery behavior to the best of its ability based off of the state that it stores.  Are you suggesting to change the recovery behavior wording? cc @jsafrane

jsafrane: The external-attacher checks that the node exists (i.e. Kubernetes CSINode or Node objects exist).

The external-attacher checks that Kubernetes object for the volume (PersistentVolume) exists. It does **not** verify that the volume exists in SP. The only way is `ListVolumes`, which I find quite clumsy, there is no filtering there.

However, even if the attacher checked that node / volume does did not exist, the spec not say if it can assume the volume has been ControllerUnpublished and what are required cleanup actions, so this check would be quite useless.

davidz627: Improved the wording as per @misterikkit suggestion. Agree with @msau42 and @jsafrane that we already implement the spirit of the behavior through stored data, as well as not doing part of it explicitly for good reason.

davidz627:
@saad-ali @jieyu @julian-hj resolved all comments, PTAL

davidz627:
@jdef added a release note to the PR description

spec.md
```diff
@@ -1292,8 +1293,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
-| Node does not exist | 5 NOT_FOUND | Indicates that a node corresponding to the specified `node_id` does not exist. | Caller MUST verify that the `node_id` is correct and that the node is available and has not been terminated or deleted before retrying with exponential backoff. |
+| Volume does not exist and volume not assumed ControllerUnpublished from node | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist and is not assumed to be ControllerUnpublished from node corresponding to the specified `node_id`. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
+| Node does not exist and volume not assumed ControllerUnpublished from node  | 5 NOT_FOUND | Indicates that a node corresponding to the specified `node_id` does not exist and the volume corresponding to the specified `volume_id` is not assumed to be ControllerUnpublished from node. | Caller MUST verify that the `node_id` is correct and that the node is available and has not been terminated or deleted before retrying with exponential backoff. |
```
saad-ali: Part of the reason for this change is that the current Kubernetes handling of this is wrong to accommodate that. However, even after this change, as specified,`the plugin SHOULD return '0 OK'` means SP could return `NOT_FOUND` and per the error code `Recovery Behavior` the `Caller MUST verify that the '..._id' is correct... before retrying with exponential backoff` -- which caller can not do because it has no way to be able to verify if a given node or volume still exist. Therefore, we should either loosen the `Recovery Behavior` to `Caller SHOULD verify...`, but really this is something that we will need to fix in CSI 2.0 -- NOT_FOUND should not be allowed as an error code, and the SP MUST return OK if volume or node are gone and effectively detached.

saad-ali:
Merging

jdef:
Should we file an issue to track the suggested 2.0 change?

davidz627:
@jdef https://github.com/container-storage-interface/spec/issues/382

PR #376, opened 2019-08-07
Upgrade the golang version and enable gomod

hakanmemisoglu:
This PR aims to fix Travis build problem.

* Upgrade version of golang to 1.11.12.
* Enable gomod.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,9 @@ jobs:
     # Lang stage: Go
     - stage:               lang
       language:            go
-      go:                  1.10.4
+      go:                  1.11.12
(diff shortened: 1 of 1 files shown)


jieyu:
You need to update the generated pb.go file as well

davidz627:
Fixes: https://github.com/container-storage-interface/spec/issues/377?

hakanmemisoglu:
@davidz627 It fixes the problem. However, it created another one. 
I am getting diff error: https://travis-ci.org/container-storage-interface/spec/jobs/569100580
`make` locally doesn't update the generated files.

Issue #380, opened 2019-08-13
Add volume capability for controller and node expansion requests


gnufied:
Currently `ControllerExpandVolumeRequest` and `NodeExpandVolumeRequest` do not contain volume capability information. This could be problematic because - a plugin may use this information to tweak resizing flow. Such as:

1. A plugin may skip doing file system resizing(on node) if volume is of type block.
2. A plugin may return `false` from `ControllerExpandVolume` call if for certain types of volumes (raw block) no `NodeExpandVolume` call is necessary.

cc @bswartz @msau42 @wongma7

ShyamsundarR:
By extension, wouldn't it be required that `NodeGetVolumeStats` also receive `volume_capability` information, to distinguish between when to return block stats and when to return filesystem stats created on the block device, based on volume type?

I am asking to understand if this or a subsequent request should cover this requirement as well, thanks.

gnufied:
May be.. I am interested in learning how a plugin will use that information for `NodeGetVolumeStats` call. In many cases for raw block devices - it is not possible to get used capacity etc. Lets open another issue to track that?

ShyamsundarR:
> May be.. I am interested in learning how a plugin will use that information for `NodeGetVolumeStats` call. In many cases for raw block devices - it is not possible to get used capacity etc. Lets open another issue to track that?

For block volumes we may omit `used` and `available` as per the spec [[1](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetvolumestats)]. 

I do not have this problem yet, so I am not opening an issue on this, unless we want to invite further feedback.

I dropped the comment here as, if we are unable to auto detect a block volume without the `volume_capability` information in `NodeExpandVolumeRequest` RPC, the problem would persist for the `NodeGetVolumeStats` RPC (and other such RPCs) as well. Hence should this issue expand the initial problem statement when considering a solution?

leakingtapan:
If we add `volume_capability` in `ControllerExpandVolume` this helps establish an explicit contract between CSI driver and CO. So that CSI driver can use `node_expansion_required` filed to indicate CO to call `NodeExpandVolume` or not.

PR #381, opened 2019-08-14
Add volume capability to volume expansion calls

gnufied:
Add volume capability to volume expansion calls.

cc @msau42 @wongma7 @saad-ali @davidz627 

Fixes https://github.com/container-storage-interface/spec/issues/380


spec.md
```diff
@@ -1821,6 +1821,15 @@ message ControllerExpandVolumeRequest {
...
   // This field is OPTIONAL.
   map<string, string> secrets = 3 [(csi_secret) = true];
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is
+  // being used as a block device - the SP MAY choose to skip calling
+  // NodeExpandVolume on the node by setting node_expansion_required
+  // to false in ControllerExpandVolumeResponse.
+  // This is a OPTIONAL field.
```
wongma7: an*
or
"This field is OPTIONAL"

jdef:
It's possible that a confused CO might send along capabilities beyond the ability of the volume (and/or plugin) to support. What error code should a CO expect in such circumstances? `InvalidArgument`? `FailedPrecondition`? `Aborted`? How does a CO recover from such?

gnufied:
I think that - if CO sends along capabilities which are not supported by the volume or plugin, then we should return `FailedPreCondition`.  If that happens CO should verify the capabilities by calling `ValidateVolumeCapabilities` and retry with right capabilities.

I will update the spec to reflect that.

csi.proto
```diff
@@ -1303,6 +1312,15 @@ message NodeExpandVolumeRequest {
...
   CapacityRange capacity_range = 3;
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is being
+  // used as a block device the SP MAY choose to skip expanding the
+  // filesystem. If volume_capability is omitted the SP MAY determine
+  // access_type from given volume_path for the volume and perform
+  // node expansion. This is an OPTIONAL field.
+  VolumeCapability volume_capability = 4;
```
leakingtapan: for ``` If volume_capability is omitted the SP MAY determine access_type from given volume_path for the volume and perform node expansion```, may be add some more details about how does SP determine this?

spec.md
```diff
@@ -2330,6 +2339,15 @@ message NodeExpandVolumeRequest {
   // plugin MAY expand the volume to its maximum capacity.
   // This field is OPTIONAL.
   CapacityRange capacity_range = 3;
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is being
+  // used as a block device the SP MAY choose to skip expanding the
+  // filesystem. If volume_capability is omitted the SP MAY determine
+  // access_type from given volume_path for the volume and perform
```
davidz627: >If volume_capability is omitted the SP MAY determine
// access_type from given volume_path for the volume and perform
// node expansion.

I know this is `MAY` but it feels a little bit prescriptive. Do we want to give this suggestion if this is not something a `SP` "should" be doing? I feel like it's a specific workaround that may not always be possible to do from the `volume_path`

gnufied: can you give me some examples where the it may not be possible to determine access_type from given volume_path?

gnufied: >  Do we want to give this suggestion if this is not something a SP "should" be doing? 

I think the only way to convey that is to convert this field from OPTIONAL to REQUIRED and MAY to MUST.

But then we will have to introduce a new node capability for being backward compatible.

davidz627: > can you give me some examples where it may not be possible to determine access_type from given volume_path

I don't have any :) Ack on the backward compatibility, think thats ok

csi.proto
gnufied: I am thinking of adding some examples that SP MAY query "volume_path" and determine if given volume_path is a block device or mounted filesystem. But being too specific here requires us bring in operating system and file system specific calls (`Stat`), which may not be suitable for higher level specs.

spec.md
```diff
@@ -1821,6 +1821,15 @@ message ControllerExpandVolumeRequest {
...
   map<string, string> secrets = 3 [(csi_secret) = true];
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is
+  // being used as a block device - the SP MAY choose to skip calling
+  // NodeExpandVolume on the node by setting node_expansion_required
+  // to false in ControllerExpandVolumeResponse.
+  // This is an OPTIONAL field.
+  VolumeCapability volume_capability = 4;
```
saad-ali: A k8s PV may have multiple access modes. How do we map that to a single VolumeCapability?

gnufied: This was discussed in CSI meeting. Resizing is only dependent on access_type and not on access_mode when it comes to whether volume can be resized(how) and hence just having singular capability should be enough.

gnufied:
@jdef updated the spec to return error code when mis-matching volume capabilities are specified.

spec.md
```diff
@@ -2354,6 +2364,15 @@ message NodeExpandVolumeRequest {
   // plugin MAY expand the volume to its maximum capacity.
   // This field is OPTIONAL.
   CapacityRange capacity_range = 3;
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is being
+  // used as a block device the SP MAY choose to skip expanding the
```
davidz627: It looks like whether the `NodeExpand` can do work is psuedo-determined in two places
1) here, based on the volume capability
2) setting `node_expansion_required` on the response to `ControllerExpand`

Can you explain why you might want to set one or not the other, or when they would differ? Basically: why do we need both, can we simplify the spec to use only one of them?

gnufied: The call to `NodeExpand` is still made if `node_expansion_required` is set to true. It is upto SP to process a block device node expansion and determine how expansion should work. For example:

1. In case of EBS or GCE-PD raw block devices, `NodeExpand` call might be a NO-OP and as such may never be called.
2. For something like iSCSI raw block devices though, even if there is no file system on the device it may still require rescan of the LUN, so as OS can see updated size.

spec.md
saad-ali: Should we say that in the spec here. That only the `access_type` will be set not `access_mode`?

gnufied: The way `VolumeCapability` is defined `access_type` and `access_mode` both will be always set. It is another thing that expansion calls don't care about access_mode but there is just no way to instantiate `VolumeCapability` and not have `access_type` set on it.

spec.md
davidz627: Since it is the `SP` that processes `ControllerExpand` and has `Block/FS` information at that level and sets the `node_expansion_required`. In case 1) couldn't those drivers set `node_expansion_required=false` and in case 2) those drivers set `node_expansion_required=true`. When `node_expansion_required=false` the `NodeExpand` should never get called right?

Then the `NodeExpand` doesn't really need the capability to decide anything anymore. I feel like maybe I'm misunderstanding something here. Could you explain why `node_expansion_required` is not enough of a signal for doing a `NodeExpand`? Is there the possibility of `node_expansion_required=true` and there will be different behavior for either `Block or FS`

spec.md
gnufied: Added additional note.

spec.md
gnufied: > Is there the possibility of node_expansion_required=true and there will be different behavior for either Block or FS

Yes there is! For example - an iSCSI implementation may return `node_expansion_required` as `true` in `ControllerExpandVolume` call for raw block devices but actual implementation of `NodeExpandVolume` call for iSCSI might be different between whether device is raw block or FS. For example - if device is raw block, it could just do a rescan of the LUN. But device has FS on it, it would rescan the LUN and also expand the filesystem.

We must also consider case of volume types that don't necessarily implement a control-plane volume expansion (such as ceph-csi). In that case, actual implementation of `NodeExpandVolume` might be different based on block or FS.

saad-ali:
This will need a rebase

spec.md
davidz627: ah I see, this answers my question. Thanks!

saad-ali:
/lgtm
/approve

@jdef or @julian-hj or @ddebroy can you PTAL

spec.md
```diff
@@ -1805,6 +1805,8 @@ This RPC allows the CO to expand the size of a volume.
 This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
 If plugin has `EXPAND_VOLUME` node capability, then `NodeExpandVolume` MUST be called after successful `ControllerExpandVolume` and `node_expansion_required` in `ControllerExpandVolumeResponse` is `true`.
 
+If specified the `volume_capability` in `ControllerExpandVolumeRequest` should be same as what CO would pass in `ControllerPublishVolumeRequest`.
```
ddebroy: nit: `If specified, the ...`

csi.proto
```diff
@@ -1066,6 +1066,15 @@ message ControllerExpandVolumeRequest {
   // Secrets required by the plugin for expanding the volume.
   // This field is OPTIONAL.
   map<string, string> secrets = 3 [(csi_secret) = true];
+
+  // Volume capability describing how the CO intends to use this volume.
+  // This allows SP to determine if volume is being used as a block
+  // device or mounted file system. For example - if volume is
+  // being used as a block device - the SP MAY choose to skip calling
```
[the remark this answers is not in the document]
gnufied: fixed

spec.md

gnufied:
@ddebroy I have reworded the comment for `NodeExpandVolume` to provide more context and capture https://github.com/container-storage-interface/spec/pull/381#discussion_r335630652 PTAL

ddebroy:
Thanks!
/lgtm

saad-ali:
Merging

Issue #385, opened 2019-08-30
Is it mandatory to 'stage' a volume on 'staging target path' provided by CO ?


humblec:
```
// The path to which the volume MAY be staged. It MUST be an
  // absolute path in the root filesystem of the process serving this
  // request, and MUST be a directory. The CO SHALL ensure that there
  // is only one `staging_target_path` per volume. The CO SHALL ensure
  // that the path is directory and that the process serving the
  // request has `read` and `write` permission to that directory. The
  // CO SHALL be responsible for creating the directory if it does not
  // exist.
  // This is a REQUIRED field.
  string staging_target_path = 3;
```

The Staging target path is carved out by CO and passed to SP. From the spec it is not clear that, whether the SP has to stage a volume on this target path. 

Is it mandatory ?  

If it is not mandatory, I see there are couple of issues. 

1) The SP dont have a mechanism to pass the `new` target path to CO via `NodeStageVolumeResponse`.

2) The cleanup operation of stage volume actually happens on the path provided by CO ( ex: Kubernetes) .  Here I am referring to `removeMountDir` code in kubernetes.

```

func removeMountDir(plug *csiPlugin, mountPath string) error {
	klog.V(4).Info(log("removing mount path [%s]", mountPath))

	....
	if !mnt {
		klog.V(4).Info(log("dir not mounted, deleting it [%s]", mountPath))
		if err := os.Remove(mountPath); err != nil && !os.IsNotExist(err) {
			return errors.New(log("failed to remove dir [%s]: %v", mountPath, err))
		}

        ....


```

humblec:
@msau4 @jsafrane

Madhu-1:
if it becomes mandatory as staging path is a directory how about block volumes?

msau42:
the CO will pass around the staging_target_directory from NodeStageVolume to NodePublishVolume.  As a SP, you stage your volume however you like inside the staging_target_directory.  For example, you can stage it directly at the starging_target_directory, or a subdirectory or file underneath. You can also store other files/metadata inside the staging_target_directory if you like.

gnufied:
There is a interesting consequence of SP storing metadata under staging target_dir, I am noticing that if stage fails then k8s tries to remove the staging directory - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_attacher.go#L330 but if it had files, that would fail. It is somewhat messy.. :(

humblec:
[...]

PR #386, opened 2019-08-31
Mention about the possibilities of using staging_target_path at NodeStage

humblec:
At present, it is unclear to SP that, how to use staging_target_path
and the possibilities of using the same. This patch add more details
on `staging target path` and its usage.

Fix # https://github.com/container-storage-interface/spec/issues/385


shay-berman:
I think its a good clarification.
We actually use the staging target path to store metadata file (with the discovered multipath device), in order to have that metadata during the NodePublishVolume(uses the metadata file to operate the l mkfs+mount it to target path) and for NodeUnStage (uses the metadata to clean up the multipath device).

So I think its good clarification that staging_target_path can be used for CSI driver metadata for example.

humblec:
@saad-ali

spec.md
```diff
@@ -1931,7 +1931,13 @@ message NodeStageVolumeRequest {
   // that the path is directory and that the process serving the
   // request has `read` and `write` permission to that directory. The
   // CO SHALL be responsible for creating the directory if it does not
-  // exist.
+  // exist.The CO will pass around the staging_target_directory from
```
jdef: Thanks for this PR! I like the spirit of this: proper use of the staging directory has come up several times in community discussion because, I think, the spec doesn't go far enough to provide usage examples.

That said, I think the wording here needs a bit of cleanup, as well as punctuation.

humblec: @jdef thanks.. 
>That said, I think the wording here needs a bit of cleanup, as well as punctuation.
>
Can you please point out the change you suggest. I am happy to take it in as it is :+1:

julian-hj: How about something like this:
```
  // The CO MUST pass the same staging_target_directory to 
  // NodeStageVolume and NodePublishVolume. 
  // The SP MAY use staging_target_directory in 
  // any way it chooses, including but not limited to:
  // - as a mount point to stage the volume directly
  // - as a parent directory for a newly created mount point
  // - as a parent directory for an attached device
  // - as a parent directory for metadata files
```

humblec: @julian-hj yeah, better.. reworded as above. @jdef @saad-ali ptal.

humblec:
@saad-ali @msau42 can you please review ?

saad-ali:
I have no objections to this. But please bring it up at the next CSI meeting, would be good to get more eyeballs on this before merge.

PR #390, opened 2019-09-19
Add staging path to NodeExpand/GetStats calls

bswartz:
Allow COs to explicitly pass the staging path to node plugins when
invoking NodeGetVolumeStats and NodeExpandVolume. Previously the
CO could choose to pass either the staging or publish path, but
there was no way for the node plugin to easily know which path it
was.

This change allows COs to pass both paths (if it wishes) or just
the staging path, and in either case, the node plugin can easily
figure out what it's being given. This makes its significantly
easier for node plugins to implement the NodeGetVolumeStats and
NodeExpandVolume RPCs, in particular it relieves the need to
store some state on the node.

For backwards compatibility, the field is not required, but it
should be trivially easy for any CO to populate the field, so we
encourage clients to set it.

Release Note:
```
Added optional staging_target_path field to NodeExpandVolume()
and NodeGetVolumeStats(). CO implementers should populate this
new field if possible when calling these RPCs.
```


csi.proto
```diff
@@ -1213,6 +1213,13 @@ message NodeGetVolumeStatsRequest {
   // the process serving this request.
   // This is a REQUIRED field.
   string volume_path = 2;
+
+  // The path where the volume is staged, if the plugin has the
+  // STAGE_UNSTAGE_VOLUME capability, otherwise empty.
+  // If not empty, it MUST be an absolute path in the root
+  // filesystem of the process serving this request.
+  // This field is OPTIONAL.
+  string staging_path = 3;
```
pure-yesmat: Thank you very much for this change!
If we want to remain consistent with other places this is used, its actually called: `staging_target_path`
eg:
https://github.com/container-storage-interface/spec/blob/9e773d2dd46a9b113d94fb3c245231d231bcfeda/csi.proto#L1152

https://github.com/container-storage-interface/spec/blob/9e773d2dd46a9b113d94fb3c245231d231bcfeda/csi.proto#L1129

bswartz: I like your suggestion, I have implemented it.

gnufied:
lgtm

spec.md
```diff
@@ -2339,6 +2350,11 @@ Otherwise `NodeExpandVolume` MUST be called after successful `NodePublishVolume`
 
 If a plugin only supports expansion via the `VolumeExpansion.OFFLINE` capability, then the volume MUST first be taken offline and expanded via `ControllerExpandVolume` (see `ControllerExpandVolume` for more details), and then node-staged or node-published before it can be expanded on the node via `NodeExpandVolume`.
 
+The `staging_target_path` field is not required, for backwards compatibility, but the CO SHOULD supply it
```
davidz627: if this is an optional field drivers will have to continue to contain logic for `if staging_target_path is unset` (since we cant assume all callers will actually use this new field) and thus not actually fixing the problem specified in the description:
>in particular it relieves the need to store some state on the node.

spec.md
```diff
@@ -2170,6 +2170,10 @@ A Node plugin MUST implement this RPC call if it has GET_VOLUME_STATS node capab
 If the volume is being used in `BlockVolume` mode then `used` and `available` MAY be omitted from `usage` field of `NodeGetVolumeStatsResponse`.
 Similarly, inode information MAY be omitted from `NodeGetVolumeStatsResponse` when unavailable.
 
+The `staging_target_path` field is not required, for backwards compatibility, but the CO SHOULD supply it
+when possible. Plugins can use this field to determine if `volume_path` is where the volume is
```
saad-ali: Drop `when possible`.

spec.md
saad-ali: This is a good point. 

If we assume that good COs will always set `staging_target_path` for plugins that implement `STAGE_UNSTAGE_VOLUME` then SP logic can be:

* If `staging_target_path` is set (which it should be for all good COs going forward), you know it is the staging path, and you can compare it to `target_path` if it is different you can assume it is the publish path.
* If `staging_target_path` is unset, but you absolutely need it you can fail the operation and say you require this field to operate. This means you will only support resizing for newer clients (given volume expansion is still beta in k8s, and SPs can control what external-resizer sidecar they pair with, this seems like a reasonable approach).

But the question is do we have a case where a good CO wants to legitimately not set `staging_target_path` AND SP will want it?

spec.md
```diff
@@ -2339,6 +2350,11 @@ Otherwise `NodeExpandVolume` MUST be called after successful `NodePublishVolume`
 
 If a plugin only supports expansion via the `VolumeExpansion.OFFLINE` capability, then the volume MUST first be taken offline and expanded via `ControllerExpandVolume` (see `ControllerExpandVolume` for more details), and then node-staged or node-published before it can be expanded on the node via `NodeExpandVolume`.
 
+The `staging_target_path` field is not required, for backwards compatibility, but the CO SHOULD supply it
+when possible. Plugins can use this field to determine if `volume_path` is where the volume is
+published or staged, and setting this field to non-empty allows plugins to function with less
+stored state on the node.
```
saad-ali: We should make this more explicit -- if  `staging_target_path` is specified, then `volume_path` MUST contain the publish path NOT stage path.

csi.proto
jdef: I thought this was going to be renamed from staging_path to staging_target_path for consistency .. what happened?

PR #397, opened 2019-11-04
Clarify idempotent for ControllerExpandVolume and NodeExpandVolume

### LINKED ISSUE #396: Add idempotent in ControllerExpandVolume
In our storage system, it needs several seconds to finish disk expansion and only allows the expanded volume size larger than original volume size in volume expansion. 
When I offline expand volume in Kubernetes, the CO may retry ControllerExpandVolume several times and lose the successful response. https://github.com/yunify/qingcloud-csi/issues/152
In my opinion, we can add idempotent requirement in ControllerExpandVolume.  `This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin MUST reply 0 OK.`

wnxn:
Add explicit documentation for clarity idempotent for
ControllerExpandVolume and NoeExpandVolume

Fixes #396 

Signed-off-by: Xin Wang <wileywang@yunify.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1802,6 +1802,9 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Controller plugin MUST implement this RPC call if plugin has `EXPAND_VOLUME` controller capability.
 This RPC allows the CO to expand the size of a volume.
 
+This operation MUST be idempotent.
+If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
+
 This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
 If plugin has `EXPAND_VOLUME` node capability, then `NodeExpandVolume` MUST be called after successful `ControllerExpandVolume` and `node_expansion_required` in `ControllerExpandVolumeResponse` is `true`.
 
@@ -2350,6 +2353,9 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 A Node Plugin MUST implement this RPC call if it has `EXPAND_VOLUME` node capability.
 This RPC call allows CO to expand volume on a node.
+
+This operation MUST be idempotent.
+If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
 
 `NodeExpandVolume` ONLY supports expansion of already node-published or node-staged volumes on the given `volume_path`.
 


wnxn:
/assign @jdef 
PTAL

spec.md
```diff
@@ -1802,6 +1802,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Controller plugin MUST implement this RPC call if plugin has `EXPAND_VOLUME` controller capability.
 This RPC allows the CO to expand the size of a volume.
 
+This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin MUST reply 0 OK.
```
jdef:
```suggestion
This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```

... since we need backwards compatibility for plugins that don't already implement this.

jdef:
Thanks for the PR. If you haven't already signed the CLA for CSI, please do so.

spec.md
wnxn: Thanks for your review. I have updated this PR.

spec.md
```diff
@@ -1802,6 +1802,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Controller plugin MUST implement this RPC call if plugin has `EXPAND_VOLUME` controller capability.
 This RPC allows the CO to expand the size of a volume.
 
+This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```
jdef:
```suggestion
This operation MUST be idempotent.
If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```

wnxn:
> Thanks for the PR. If you haven't already signed the CLA for CSI, please do so.

Tomorrow, I will ask my colleague about how to sign the CLA for CSI. Please wait a moment.

spec.md
```diff
@@ -2351,6 +2353,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Node Plugin MUST implement this RPC call if it has `EXPAND_VOLUME` node capability.
 This RPC call allows CO to expand volume on a node.
 
+This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```
jdef:
```suggestion
This operation MUST be idempotent.
If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```

spec.md
wnxn: Done.

spec.md
wnxn: Done.

spec.md
```diff
@@ -1802,6 +1802,9 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Controller plugin MUST implement this RPC call if plugin has `EXPAND_VOLUME` controller capability.
 This RPC allows the CO to expand the size of a volume.
 
+This operation MUST be idempotent.
+If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
+
```
gnufied: Should this say "after expansion" or "before expansion"?

wnxn: "The capacity requirement after expansion" means target size. The sentence means to compare current volume size with target size. If current size is larger than or equal to target size, the interface SHOULD reply 0 OK.

wnxn:
> > Thanks for the PR. If you haven't already signed the CLA for CSI, please do so.
> 
> Tomorrow, I will ask my colleague about how to sign the CLA for CSI. Please wait a moment.

Hi @jdef, I signed the CLA for CSI and sent email to container-storage-interface-approvers@googlegroups.com.
Please review this PR again. Thanks.

spec.md
```diff
@@ -1802,6 +1802,9 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Controller plugin MUST implement this RPC call if plugin has `EXPAND_VOLUME` controller capability.
 This RPC allows the CO to expand the size of a volume.
 
+This operation MUST be idempotent.
+If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```
jdef:
```suggestion
If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
```

spec.md
```diff
@@ -2351,6 +2354,9 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 A Node Plugin MUST implement this RPC call if it has `EXPAND_VOLUME` node capability.
 This RPC call allows CO to expand volume on a node.
 
+This operation MUST be idempotent.
+If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.
```
jdef:
```suggestion
If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
```

wnxn:
> Thanks for the feedback @gnufied. I made some additional suggestions to clarify. PTAL

Thanks for @jdef and @gnufied , this PR has been updated.

gnufied:
lgtm

jdef:
@saad-ali @jieyu @julian-hj @ddebroy PTAL

jdef:
> If the service returns an error when the volume is already the correct size, wouldn't that make it not idempotent?

No ... idempotent has to do w/ the resulting state of the system vs. the error codes returned. Ideally, all plugins would return 0 if the target size has already been achieved.
If a plugin doesn't implement this behavior and returns a non-zero error but otherwise leaves the volume untouched then it's still idempotent, it's just returning a non-recommended RPC status (and so may be sacrificing CO compatibility).

wnxn:
> > If the service returns an error when the volume is already the correct size, wouldn't that make it not idempotent?
> 
> No ... idempotent has to do w/ the resulting state of the system vs. the error codes returned. Ideally, all plugins would return 0 if the target size has already been achieved.
> If a plugin doesn't implement this behavior and returns a non-zero error but otherwise leaves the volume untouched then it's still idempotent, it's just returning a non-recommended RPC status (and so may be sacrificing CO compatibility).

What is the definition of idempotent in our CSI spec?
(1) For the same input, the operation can get the same output.
(2) Operations that have no side-effects if executed multiple times. 

In my opinion,
For (1), the first keyword, MUST or SHOULD, should be the same as second keyword.
> This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin MUST reply 0 OK. 

> This operation SHOULD be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.

For (2), for the purpose of backwards compatibility, we can clarify the idempotent same as now.
> This operation MUST be idempotent. If a volume corresponding to the specified volume ID is larger than or equal to the capacity requirements after expansion, the plugin SHOULD reply 0 OK.

julian-hj:
> What is the definition of idempotent in our CSI spec?
> (1) For the same input, the operation can get the same output.
> (2) Operations that have no side-effects if executed multiple times.

Elsewhere in the spec, we've normally observed the first definition.  For example: https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume

Having said that, it seems like the rest of the world's definition of idempotent is converging on what @jdef said.  So I'd be fine with that, except that it's a bit inconsistent with other RPCs.

Is there any reason that we need to use SHOULD here?  If not, maybe we could just switch it to MUST/MUST in order to keep things consistent any tidy.

jdef:
SHOULD is for backwards compatibility for plugins that might not return 0 OK

julian-hj:
> SHOULD is for backwards compatibility for plugins that might not return 0 OK
Fair enough.

jdef:
Is it worth filing a follow up issue for csi v2 that fixes the wording to
be more consistent?

gnufied:
I think so yeah. Do we have a label for csiv2? I can write a github issue.

PR #407, opened 2020-01-06
WIP: Add NodeFreeze/UnfreezeVolumeFilesystem Support

xing-yang:


jdef:
Freeze/unfreeze is *only* related to snapshots, yes? Or are there other use cases that we're thinking about? If it's just for snapshotting, why can't freeze/unfreeze implementation be built in to the CreateSnapshot RPC itself?

If there are other, high level use cases related to freeze then please enumerate them here for the benefit of everyone reviewing this PR. Asking people to review a series of gdocs to dig out relevant threads of conversation is .. probably asking too much.

xing-yang:
CreateSnapshot is a controller RPC while filesystem freeze/unfreeze proposed here will be in the node PRCs.  It is possible to do filesystem freeze/unfreeze using ExecutionHook without these new node PRCs but the concern is the elevated privileges required on the pod for that to happen.

We still need to figure out lots of details before formally proposing this.

jdef:
once #417 lands, this PR should mark up API additions w/ "alpha" annotations

xing-yang:
Sure.

xujianhai666:
what is the progress status for this proposal @xing-yang

xujianhai666:
I found I need nodeserver freeze as same as u

PR #412, opened 2020-02-04
Fix error code on wrong snapshot ID or topology

jsafrane:
Add one more reason to return `ALREADY_EXISTS` from `CreateVolumeRequest`: the volume to be created already exists, but it was restored from a *different* snapshot than requested

This follows all other `ALREADY_EXISTS` cases - the volume already exists, but it's something else than CO requested.


jieyu:
This change LGTM. I think we also need to add one more here, which is `accessibility_requirements`. I.e., the SP should return `ALREADY_EXISTS` if the CO requested different `accessibility_requirements` but using the same name. Thoughts?

jsafrane:
Added `accessibility_requirements` to the sentence.

spec.md
```diff
@@ -1119,7 +1119,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
 | Source does not exist | 5 NOT_FOUND | Indicates that the specified source does not exist. | Caller MUST verify that the `volume_content_source` is correct, the source is accessible, and has not been deleted before retrying with exponential back off. |
-| Volume already exists but is incompatible | 6 ALREADY_EXISTS | Indicates that a volume corresponding to the specified volume `name` already exists but is incompatible with the specified `capacity_range`, `volume_capabilities` or `parameters`. | Caller MUST fix the arguments or use a different `name` before retrying. |
+| Volume already exists but is incompatible | 6 ALREADY_EXISTS | Indicates that a volume corresponding to the specified volume `name` already exists but is incompatible with the specified `capacity_range`, `volume_capabilities`, `parameters`, `accessibility_requirements` or `content_source`. | Caller MUST fix the arguments or use a different `name` before retrying. |
```
xing-yang: s/content_source/volume_content_source

jsafrane: Fixed

xing-yang:
lgtm

PR #421, opened 2020-03-25
Added HashiCorp Nomad to Container Orchestrators

mpron:
Added a bullet link for HashiCorp Nomad to the readme file linking to the information about CSI support. New feature so it may change as documentation pages are created.


jieyu:
@mpron Can you remove `HashiCorp` and just leave the project name? We tend to only list project names here for COs.

PR #422, opened 2020-03-30
 Added Nomad to Container Orchestrators #421

mpron:
Added a bullet link for Nomad to the readme file linking to the information about CSI support. New feature so it may change as documentation pages are created.

PR #424, opened 2020-04-09
add reference for HashiCorp Nomad

shantanugadgil:
HashiCorp Nomad (>= v0.11.0) supports CSI plugins.
Add a reference of the same.


shantanugadgil:
@tgross besides the short name vs. full name, this seems to be a duplicate of #421 and #422

Issue #402, opened 2019-11-26
Allow passing filesystem formatting options in NodeStageVolume phase


dsupure:
**What would you like to be added:**

Filesystem `formatting options` for the host formatting tools, such as linux mkfs. 

**Why is this needed:**

At present, `VolumeCapability_MountVolume` only has `FsType` and `MountFlags` fields, 
there is no way to pass the options for formatting filesystem such as ext4, xfs in linux.  Each vendor needs to implement its own way to deliver these options.  It would be nice to have a field like `FsFmtFlags` to passing the options to plugins/drivers.

PR #427, opened 2020-04-29
Add mkfs_options to VolumeCapability.MountVolume

nixpanic:
Make it possible to add options to `mkfs` for formatting a filesystem.
With additional options to create a filesystem on a volume, provisioners
can be instructed to enable/disable certain features that may not be
available on all nodes.

See-also: ceph/ceph-csi#374
Fixes: #402


NicolasT:
While mount flags are something one may want to change over time for a particular volume, the options used to create this filesystem (`mkfs` flags) can't be changed over time (after initial formatting) across a volume's lifetime (some options may be reconfigurable, but then those are not strictly `mkfs` options).

As such, shouldn't this information be only made available to the `CreateVolume` call, instead of passing it to a bunch of RPC calls that have little to no use for it?

ameade:
> While mount flags are something one may want to change over time for a particular volume, the options used to create this filesystem (`mkfs` flags) can't be changed over time (after initial formatting) across a volume's lifetime (some options may be reconfigurable, but then those are not strictly `mkfs` options).
> 
> As such, shouldn't this information be only made available to the `CreateVolume` call, instead of passing it to a bunch of RPC calls that have little to no use for it?

I think a lot of CSI drivers do the format on NodeStageVolume since that's when they have the volume mounted.

pohly:
I agree that this is common, but can't such CSI drivers take such parameters through the CreateVolume call, store them in the volume context, and then use that during NodeStageVolume?

nixpanic:
> I agree that this is common, but can't such CSI drivers take such parameters through the CreateVolume call, store them in the volume context, and then use that during NodeStageVolume?

Yes, this is a very good alternative.

PR #429, opened 2020-05-07
Support online and offline expansion without requiring them to be declared

gnufied:
Ability for a plugin to delcare online and offline controller expansion
capabilities via GetPluginCapabilities is deprecated. A plugin may
support either mode of operation without having to declare them
in GetPluginCapabilities call.

If a plugin can not support controller expansion of published volume
it may return Volume-in-use error and CO should ensure that volume is
not published before retrying with exponential backoff.


gnufied:
cc @jdef @saad-ali

gnufied:
cc @xing-yang

spec.md
```diff
@@ -1930,24 +1931,15 @@ This RPC allows the CO to expand the size of a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
 
-This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
+This call MAY be made by the CO during any time in the lifecycle of the volume after creation but an SP may not permit controller expansion of volumes which are controller published or available on a node. In which case - the plugin may return gRPC error code `FAILED_PRECONDITION` (Volume in use) and CO SHOULD ensure that volume is not published and retry with exponential backoff.
+
```
xing-yang, line 1935: s/and retry/before retrying

spec.md
```diff
@@ -644,6 +644,7 @@ message PluginCapability {
   oneof type {
     // Service that the plugin supports.
     Service service = 1;
+    // deprecated
```
xing-yang, line 647: "deprecated" should also be added in front of the "message VolumeExpansion" on line 598 as well?

spec.md
gnufied: fixed

spec.md

spec.md
```diff
@@ -1930,24 +1945,15 @@ This RPC allows the CO to expand the size of a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
 
-This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
+This call MAY be made by the CO during any time in the lifecycle of the volume after creation but an SP may not permit controller expansion of volumes which are controller published or available on a node. In which case - the plugin may return gRPC error code `FAILED_PRECONDITION` (Volume in use) and CO SHOULD ensure that volume is not published before retrying with exponential backoff.
+
```
xing-yang, line 1949: Line 659 says "published and available volumes on a node".  Here it says "published or available on a node".  Is it "and" or "or"?  Should be consistent in both places.

spec.md
```diff
@@ -1930,24 +1945,15 @@ This RPC allows the CO to expand the size of a volume.
...
-
-- The plugin does NOT have controller `PUBLISH_UNPUBLISH_VOLUME` capability, the plugin has node `STAGE_UNSTAGE_VOLUME` capability, and `NodeUnstageVolume` has been completed successfully.
-
-OR ELSE
-
-- The plugin does NOT have controller `PUBLISH_UNPUBLISH_VOLUME` capability, nor node `STAGE_UNSTAGE_VOLUME` capability, and `NodeUnpublishVolume` has completed successfully.
 
 Examples:
-* Offline Volume Expansion:
+* Offline Volume Expansion(SP does not permit controller expansion of controller published volume):
```
xing-yang, line 1956: nit: Add a space before (

spec.md
```diff
@@ -595,6 +595,10 @@ message PluginCapability {
     Type type = 1;
   }
 
+  // Deprecated - Ability for a plugin to delcare online and offline
+  // controller expansion capabilities via PluginCapability
+  // is deprecated. A plugin may support either mode of operation
+  // without having to declare them in PluginCapability.
   message VolumeExpansion {
     enum Type {
       UNKNOWN = 0;
```
saad-ali, line 604: Should we also remove the comments on `ONLINE = 1;` and `OFFLINE = 2;` below and replace them with `// Deprecated` + ` [deprecated=true]` annotation?

spec.md
```diff
@@ -1930,24 +1945,15 @@ This RPC allows the CO to expand the size of a volume.
 This operation MUST be idempotent.
 If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.
 
-This call MAY be made by the CO during any time in the lifecycle of the volume after creation if plugin has `VolumeExpansion.ONLINE` capability.
+This call MAY be made by the CO during any time in the lifecycle of the volume after creation but an SP may not permit controller expansion of volumes which are controller published or available on a node. In which case - the plugin may return gRPC error code `FAILED_PRECONDITION` (Volume in use) and CO SHOULD ensure that volume is not published before retrying with exponential backoff.
```
saad-ali, line 1948:
```suggestion
This call MAY be made by the CO during any time in the lifecycle of the volume after creation. If an SP does not permit controller expansion of volumes which are controller published or available on a node, the SP may return a gRPC error code `FAILED_PRECONDITION` (Volume in use) and the CO SHOULD ensure that volume is not published on controller or node before retrying with exponential backoff.
```

spec.md
saad-ali: or makes more sense I think

spec.md
saad-ali, line 1956: Should we remove the offline example altogether?

spec.md
```diff
@@ -2003,7 +2009,7 @@ message ControllerExpandVolumeResponse {
 |-----------|-----------|-------------|-------------------|
 | Exceeds capabilities | 3 INVALID_ARGUMENT | Indicates that CO has specified capabilities not supported by the volume. | Caller MAY verify volume capabilities by calling ValidateVolumeCapabilities and retry with matching capabilities. |
 | Volume does not exist | 5 NOT FOUND | Indicates that a volume corresponding to the specified volume_id does not exist. | Caller MUST verify that the volume_id is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
-| Volume in use | 9 FAILED_PRECONDITION | Indicates that the volume corresponding to the specified `volume_id` could not be expanded because it is currently published on a node but the plugin does not have ONLINE expansion capability. | Caller SHOULD ensure that volume is not published and retry with exponential back off. |
+| Volume in use | 9 FAILED_PRECONDITION | Indicates that the volume corresponding to the specified `volume_id` could not be expanded because it is currently published on a node. | Caller SHOULD ensure that volume is not published and retry with exponential back off. |
```
saad-ali, line 2012:
```suggestion
| Volume in use | 9 FAILED_PRECONDITION | Indicates that the volume corresponding to the specified `volume_id` could not be expanded because it is currently published on a controller or node and the SP does not support expansion while published. | Caller SHOULD ensure that volume is not published to controller or node before retrying with exponential back off. |
```

saad-ali:
Also, please add a 
```
 ```release-note
```
block with a draft of the intended release note so we can all review and agree on that in this PR as well.

spec.md
xing-yang: I think this example is very important for CSI drivers that support offline volume expansion.  It clearly explains how this case would work.

Issue #434, opened 2020-05-22
Floating grpc dependencies and build failures


gnufied:
The build here is failing - https://travis-ci.org/github/container-storage-interface/spec/jobs/689878382 with following error:

```
get "google.golang.org/protobuf/runtime/protoimpl": found meta tag get.metaImport{Prefix:"google.golang.org/protobuf", VCS:"git", RepoRoot:"https://go.googlesource.com/protobuf"} at //google.golang.org/protobuf/runtime/protoimpl?go-get=1
get "google.golang.org/protobuf/runtime/protoimpl": verifying non-authoritative meta tag                                                      
get "golang.org/x/sys/unix": found meta tag get.metaImport{Prefix:"golang.org/x/sys", VCS:"git", RepoRoot:"https://go.googlesource.com/sys"} at //golang.org/x/sys/unix?go-get=1
get "golang.org/x/sys/unix": verifying non-authoritative meta tag                                                                             
golang.org/x/sys (download)                                                                                                                   
get "golang.org/x/net/http2": found meta tag get.metaImport{Prefix:"golang.org/x/net", VCS:"git", RepoRoot:"https://go.googlesource.com/net"} at //golang.org/x/net/http2?go-get=1
get "golang.org/x/net/http2": verifying non-authoritative meta tag                                                                            
get "golang.org/x/net/http2/hpack": found meta tag get.metaImport{Prefix:"golang.org/x/net", VCS:"git", RepoRoot:"https://go.googlesource.com/net"} at //golang.org/x/net/http2/hpack?go-get=1
get "golang.org/x/net/http2/hpack": verifying non-authoritative meta tag                                                                      
go install ./csi                                                                                                                              
# google.golang.org/genproto/googleapis/rpc/status                                                                                            
../../../../../google.golang.org/genproto/googleapis/rpc/status/status.pb.go:42:11: undefined: "github.com/golang/protobuf/proto".ProtoPackageIsVersion4
Makefile:123: recipe for target 'csi.a' failed                                                                                                
make[1]: *** [csi.a] Error 2                                                                                                                  
make[1]: Leaving directory '/home/hekumar/goal/src/github.com/container-storage-interface/spec/lib/go'                                        
Makefile:36: recipe for target 'build' failed                                                                                                 
make: *** [build] Error 2
```

[...]

PR #438, opened 2020-05-30
Fix grpc build by cloning dependencies via Makefile

gnufied:
Fixes https://github.com/container-storage-interface/spec/issues/434

cc @saad-ali @jdef @xing-yang


xing-yang:
LGTM

PR #457, opened 2020-10-07
Add Secrets in NodeExpandVolumeRequest

### LINKED ISSUE #445: Secrets in NodeExpandVolume
Hello!

It would be nice to have secrets available in the NodeExpandVolume.

Usecase : encrypted device with LUKS, which needs the passphrase in order to resize the device on the node.

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

Closes #445 

cc @humblec


humblec:
@Sh4d1 Thanks.. LGTM. 

@gnufied @jsafrane @msau42 @xing-yang @saad-ali can we get some attention to this ? this become a blocker for different use cases of nodeExpansion.

xing-yang:
How are you going to pass the secrets from the Kubernetes side?  Does this require a change in Kubernetes API and also in external-provisioner like the following?
```
https://github.com/kubernetes/kubernetes/blob/v1.20.0-alpha.1/pkg/apis/core/types.go#L1656
https://github.com/kubernetes-csi/external-provisioner/blob/v2.0.2/pkg/controller/controller.go#L713
```

Sh4d1:
@xing-yang yeah these changes would also be needed

gnufied:
Okay lets open a issue in k/k repo for that. Do we expect node expansion secrets to be different from controller one? If yes - we will indeed need API change.

humblec:
> https://github.com/kubernetes-csi/external-provisioner/blob/v2.0.2/pkg/controller/controller.go#L713

Ack.. We need similar changes.. @xing-yang 
>
Okay lets open a issue in k/k repo for that. Do we expect node expansion secrets to be different from controller one? If yes - we will indeed need API change.
>

@gnufied sure will do that.. NodeExpansion Secret can be same as "ControllerSecret" ,  I dont see a reason to have different one atleast for our use case.

Sh4d1:
I agree with @humblec the secret should be the same. 
@humblec do you open the issue on k/k?

humblec:
> I agree with @humblec the secret should be the same.
> @humblec do you open the issue on k/k?

@Sh4d1 Sure.. On it!

gnufied:
cc @saad-ali @jdef

spec.md
```diff
@@ -2562,6 +2562,11 @@ message NodeExpandVolumeRequest {
   // access_type from given volume_path for the volume and perform
   // node expansion. This is an OPTIONAL field.
   VolumeCapability volume_capability = 5;
+
+  // Secrets required by plugin to complete node expand volume request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 6 [(csi_secret) = true];
```
saad-ali, line 2569: We want to make sure all new fields, methods, etc. go through alpha/beta/ga.

So please add `(alpha_field) = true` to this field to indicate that it is alpha in this release, in the next release we can move to beta, and GA after that.

saad-ali (changes requested):

spec.md
saad-ali:
```suggestion
  map<string, string> secrets = 6 [(csi_secret) = true, (alpha_field) = true];
```

csi.proto
```diff
@@ -1494,6 +1494,11 @@ message NodeExpandVolumeRequest {
   // access_type from given volume_path for the volume and perform
   // node expansion. This is an OPTIONAL field.
   VolumeCapability volume_capability = 5;
+
+  // Secrets required by plugin to complete node expand volume request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 6 [(csi_secret) = true];
```
saad-ali, line 1501:
```suggestion
  map<string, string> secrets = 6 [(csi_secret) = true, (alpha_field) = true];
```

saad-ali:
Please make sure you sign the CLA for CSI
See https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#licence-and-cla

csi.proto
```diff
@@ -1494,6 +1494,11 @@ message NodeExpandVolumeRequest {
   // access_type from given volume_path for the volume and perform
   // node expansion. This is an OPTIONAL field.
   VolumeCapability volume_capability = 5;
+
+  // Secrets required by plugin to complete node expand volume request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 6 [(csi_secret) = true, (alpha_field) = true];
```
gnufied, line 1501: you have to split this line, for now. our tooling does not like long lines. I think tooling should be fixed to make an exception for code lines but that is a discussion for another issue perhaps.

saad-ali: Could try

```
  map<string, string> secrets = 6
      [(csi_secret) = true, (alpha_field) = true];
```

Sh4d1:
Just sent out the CLA!

saad-ali:
CLA confirmed! Thanks.
Please rebase this PR.

Sh4d1:
Rebased!

saad-ali (approved):
Will merge as soon as 1.4.0 is cut (1.4.0-rc was cut earlier today and we're in a 72 hour waiting period). This will end up being part of 1.5.0

saad-ali:
@jdef Any objections? If not, please feel free to merge.

PR #475, opened 2021-04-19
Relax node_id length limit to 256

Jiawei0227:
node_id has been relaxed to 192 in https://github.com/container-storage-interface/spec/commit/396c3332ca1216dea620f64f5f2d60686ae9a0a5. But it is still not long enough for some plugins node_id. So relax the constraint to 256 to make sure this satisfy the requirements.


saad-ali:
Thanks. No objections to this in the CSI Community meeting either. 
Going to go ahead and merge this.

Issue #465, opened 2020-12-04
2.0 Tracking Issue


bswartz:
This issue is to collect ideas for an eventual 2.0 release of the CSI spec. This issue in no way suggests we plan to address any of these ideas, or that a 2.0 release will ever happen. Rather, the intention is to collect ideas that would be very difficult to implement in a 1.x release because they can't be made backward compatible, such that if we ever did do a 2.0 release, we would want to implement as many of these ideas as possible in that release to avoid needed a 3.0 later on.

bswartz:
The SINGLE_NODE / MULTI_NODE distinction in the access modes is and how they are handled in NodePublishVolume()
completely contrary to what Kubernetes actually wants. The documentation for SINGLE_NODE / MULTI_NODE matches what would be useful semantics for Kubernetes, but the documentation for NodePublishVolume() then says the opposite thing (basically saying that supporting more than one pod requires MULTI_NODE). To fix this in 1.0 would require adding new access modes and deprecating the existing ones, and updating NodePublishVolume() to interpret the new access modes in the way we actually want. A capability bit would need to be exposed to COs to detect if the SP supported the new access modes so Kubernetes could use them. All of that is far more gross than just rolling a 2.0 version with different semantics for NodePublishVolume() with regard to access modes. We could add new access modes in a 2.0 that correspond to the existing behavior in the unusual case that someone would want that.

bswartz:
There is some reworking we would want to do around resize if we had a 2.0 version. In particular, cloning volumes from snapshots or other volumes and making them larger at the same time should have been tied to the resize functionality so that COs could leverage existing filesystem-resize workflows rather than forcing the SPs to have 2 different ways to expand filesystems (one for explicit expand operations, and another for clone-to-larger-size).

PR #476, opened 2021-04-26
Access modes for single node single/multi writer

chrishenzie:
## Access Modes

In the CSI spec there are conflicting definitions of the [`SINGLE_NODE_WRITER`] access mode. By definition, `SINGLE_NODE_WRITER` means "Can only be published once as read/write on a single node, at any given time." The problem is how this access mode is used during `NodePublishVolume`, which is typically where volume mounting is performed.

The CSI spec defines that when [`NodePublishVolume`] is called a second time for a volume with a non-`MULTI_NODE` access mode and with a different target path, the plugin should return `FAILED_PRECONDITION`. For CSI plugins that strictly adhere to the spec, this guarantees that a volume can only be mounted to a single target path, which means `SINGLE_NODE_WRITER` restricts access to a single pod on a single node. This behavior conflicts with the original definition. Due to this conflict, we do not have an access mode that represents multiple writers on the same node. See https://github.com/container-storage-interface/spec/issues/465#issuecomment-739046063 for more details.

The proposed solution for this issue is to introduce two new access modes: `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`. These access modes exist to replace `SINGLE_NODE_WRITER` and disambiguate the number of writers on a single node.

[`SINGLE_NODE_WRITER`]: https://github.com/container-storage-interface/spec/blob/v1.4.0/csi.proto#L405-L407
[`NodePublishVolume`]: https://github.com/container-storage-interface/spec/blob/v1.4.0/spec.md#nodepublishvolume

## Capabilities

Included in this change are capability bits in the controller and node services to indicate that a driver understands these new access modes. This approach was chosen in order to preserve backwards compatibility. When a client calls a CSI driver that supports the `SINGLE_NODE_MULTI_WRITER` capability, the client should replace any uses of the `SINGLE_NODE_WRITER` access mode with the appropriate `SINGLE_NODE_{SINGLE,MULTI}_WRITER` access mode depending on their use case.


spec.md
```diff
@@ -2232,6 +2244,15 @@ The following table shows what the Plugin SHOULD return when receiving a second
 | MULTI_NODE     | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                 |
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
+|                                | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|--------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER      | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| Non SINGLE_NODE_SINGLE_WRITER  | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
```
gnufied, line 2252: Did `Non SINGLE_NODE_SINGLE_WRITER` meant -> `SINGLE_NODE_MULTI_WRITER` ? We are saying above that - this table applies for `SINGLE_NODE_MULTI_WRITER` but we don't use that word in the table - so it is bit confusing.

chrishenzie: Yeah, I went back and forth on the wording here. I had a row for `SINGLE_NODE_MULTI_WRITER` but realized I would probably also need a row for "Everything else" but it would be treated in the same way as `SINGLE_NODE_MULTI_WRITER` so I combined them into the "Non" row. I think having a row for this to make it explicit is a good idea, I'll add this.

chrishenzie: Done

chrishenzie:
@bswartz Would you be able to take a look at this? Thank you!

spec.md
```diff
@@ -2232,6 +2244,16 @@ The following table shows what the Plugin SHOULD return when receiving a second
 | MULTI_NODE     | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                 |
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non SINGLE_NODE_{SINGLE,MULTI}_WRITER | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
```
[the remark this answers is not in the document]
chrishenzie: Done, expanded the table to include the `MULTI_NODE` and Non `MULTI_NODE` in the above table, and left a note requiring continued support for `SINGLE_NODE_WRITER`

spec.md
```diff
@@ -887,6 +887,14 @@ message VolumeCapability {
       // Can be published as read/write at multiple nodes
       // simultaneously.
       MULTI_NODE_MULTI_WRITER = 5;
+
+      // Can only be published once as read/write at a single worklad on
```
gnufied, line 891: typo in word worklad.

spec.md
```diff
@@ -1704,6 +1712,10 @@ message ControllerServiceCapability {
       // This enables COs to, for example, fetch per volume
       // condition after a volume is provisioned.
       GET_VOLUME = 12 [(alpha_enum_value) = true];
+
+      // Indicates the SP supports the SINGLE_NODE_MULTI_WRITER access
+      // mode.
+      SINGLE_NODE_MULTI_WRITER = 13 [(alpha_enum_value) = true];
     }
```
gnufied, line 1719: Why does this exists as a control-plane capability?

spec.md
```diff
@@ -2232,6 +2244,17 @@ The following table shows what the Plugin SHOULD return when receiving a second
 | MULTI_NODE     | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                 |
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
```
gnufied, line 2248: So I think we might have a problem here - can you query control-plane capabilities from node service?

spec.md
chrishenzie: COs require this capability in order to determine which access mode to supply for control plane operations that take `VolumeCapability`, like `CreateVolume()`. This capability will be used in the same way as the capability on the node side.

spec.md
chrishenzie: The node service has an equivalent capability bit that is used for determining which access mode to supply for node operations that take `VolumeCapability`.

spec.md
chrishenzie: Fixed

spec.md
```diff
@@ -2232,6 +2244,17 @@ The following table shows what the Plugin SHOULD return when receiving a second
...
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode in order to clarify the number of writers for a volume on a single node. Plugins MUST continue to support `SINGLE_NODE_WRITER` to allow older COs to continue working.
+
```
gnufied, line 2257: Do we need to explicitly deprecate `SINGLE_NODE_WRITER` at some point? I guess we need to wait for new modes to go GA before we can deprecate older one.

gnufied:
lgtm

spec.md
chrishenzie: I think it would be a good idea, but there are no deprecation plans as of yet.

spec.md
```diff
@@ -2232,6 +2244,17 @@ The following table shows what the Plugin SHOULD return when receiving a second
...
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode in order to clarify the number of writers for a volume on a single node. Plugins MUST continue to support `SINGLE_NODE_WRITER` to allow older COs to continue working.
```
saad-ali, line 2256: > Plugins MUST continue to support `SINGLE_NODE_WRITER` to allow older COs to continue working.

Can you detail what "support" means here.

spec.md
```diff
@@ -2232,6 +2244,17 @@ The following table shows what the Plugin SHOULD return when receiving a second
 | MULTI_NODE     | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                 |
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
```
saad-ali, line 2249: nit: Can you add an explanation before or after these tables explaining what T1, T2, P1, P2 are in this context (I realize existing table doesn't do this, but let's fix it while we're here).

spec.md
chrishenzie: Done, here is the new wording
> Plugins MUST continue to accept and allow use of the `SINGLE_NODE_WRITER` access mode to support older COs to continue working.

spec.md
chrishenzie: Done. There was a little section on that below the table. I moved it to the top of the first table so it's the first thing the user sees.

csi.proto
```diff
@@ -420,6 +420,14 @@ message VolumeCapability {
...
       // simultaneously.
       MULTI_NODE_MULTI_WRITER = 5;
+
+      // Can only be published once as read/write at a single workload
+      // on a single node, at any given time.
+      SINGLE_NODE_SINGLE_WRITER = 6 [(alpha_enum_value) = true];
+
+      // Can be published as read/write at multiple workloads on a
+      // single node simultaneously.
+      SINGLE_NODE_MULTI_WRITER = 7 [(alpha_enum_value) = true];
```
jingxu97, line 430: with this change, I am trying to reason on MULTI_NODE_SINGLE_WRITER. There are two possible interoperation 

1. only a single pod from a node can access it, 
2. multiple pods on the same single node can access it

From the comments of this mode, it seems refer to 2. Is that the intention? Do we want to support both? @saad-ali
```
// Can be published at multiple nodes simultaneously. Only one of
// the node can be used as read/write. The rest will be readonly.
```

chrishenzie: I think because the name mentions multiple nodes, it's saying this volume is attachable to multiple machines and readable by multiple machines, but only one pod on one of the machines has write access.

jingxu97: i don't see it mentioned "one workload on one of the machines has write access" in the doc?

bswartz: I'm 99% sure that Kubernetes never uses the MULTI_NODE_SINGLE_WRITER access mode for anything, so while it suffers from the same vagueness issues as other single writer access modes, nobody actually experiences problems related to this. Ideally we could sharpen up whether the single writer for that access mode is a single workload or all the workloads on a single node, but I don't see that as a priority.

spec.md
```diff
@@ -2227,12 +2239,23 @@ If this RPC failed, or the CO does not know if it failed or not, it MAY choose t
...
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node. Plugins MUST continue to accept and allow use of the `SINGLE_NODE_WRITER` access mode to support older COs to continue working.
```
saad-ali, line 2258:
```suggestion
The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node. Plugins MUST continue to accept and allow use of the `SINGLE_NODE_WRITER` access mode to permit older COs to continue working.
```

saad-ali (approved):
/lgtm
/approve

mod nit

@ddebroy can you PTAL?

spec.md
```diff
@@ -2227,12 +2239,23 @@ If this RPC failed, or the CO does not know if it failed or not, it MAY choose t
 This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE capability (i.e., `access_mode` is either `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`).
```
ddebroy, line 2239: Should the `the volume has MULTI_NODE capability` bit in the sentence above be amended to also include `SINGLE_NODE_MULTI_WRITER`?

ddebroy (commented):
Just one suggestion for a potential clarification. Otherwise LGTM.

spec.md
chrishenzie: Yes this is a good idea. Added, here is the amended sentence. Is this understandable?

> This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE or SINGLE_NODDE_MULTI_WRITER (see second table) access modes. Example MULTI_NODE access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER.

spec.md
```diff
@@ -2224,15 +2236,26 @@ If the volume corresponding to the `volume_id` has already been published at the
 
 If this RPC failed, or the CO does not know if it failed or not, it MAY choose to call `NodePublishVolume` again, or choose to call `NodeUnpublishVolume`.
 
-This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE capability (i.e., `access_mode` is either `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`).
+This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE or SINGLE_NODDE_MULTI_WRITER (see second table) access modes. Example MULTI_NODE access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```
jdef, line 2239:
```suggestion
This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE or SINGLE_NODE_MULTI_WRITER (see second table) access modes. Example MULTI_NODE access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```

jdef:
thanks for putting this together, looking now..

spec.md
```diff
@@ -887,6 +887,14 @@ message VolumeCapability {
       // Can be published as read/write at multiple nodes
       // simultaneously.
       MULTI_NODE_MULTI_WRITER = 5;
+
+      // Can only be published once as read/write at a single workload
+      // on a single node, at any given time.
+      SINGLE_NODE_SINGLE_WRITER = 6 [(alpha_enum_value) = true];
```
jdef, line 893: CO's using the experimental capability `SINGLE_NODE_MULTI_WRITER` should probably avoid specifying `SINGLE_NODE_WRITER`, yes? if for no other reason than for clarity. if so, please mention that here.

spec.md
```diff
@@ -887,6 +887,14 @@ message VolumeCapability {
...
+
+      // Can be published as read/write at multiple workloads on a
+      // single node simultaneously.
+      SINGLE_NODE_MULTI_WRITER = 7 [(alpha_enum_value) = true];
```
jdef, line 897: @chhsia0 any initial thoughts re: Mesos compatibility?

spec.md
```diff
@@ -2224,15 +2236,26 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node. Plugins MUST continue to accept and allow use of the `SINGLE_NODE_WRITER` access mode to permit older COs to continue working.
```
jdef, line 2258:
```suggestion
The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node. Plugins MUST accept and allow use of the `SINGLE_NODE_WRITER` access mode (subject to the processing rules above), when either `SINGLE_NODE_SINGLE_WRITER` and/or `SINGLE_NODE_MULTI_WRITER` are supported, in order to permit older COs to continue working.
```
^ is this more precise? or do I not understand what the goal is?

jdef: I feel like this should probably be called out somewhere in the generated .proto too

spec.md
chrishenzie: Thanks for catching this

spec.md
chrishenzie: Yes, COs should not use `SINGLE_NODE_WRITER` in this case. Added this to the description.

spec.md
chrishenzie: Yes this is useful, thank you.
- I called this out on the `SINGLE_NODE_MULTI_WRITER` access mode capabilities in the node and controller services
- I updated the `SINGLE_NODE_MULTI_WRITER` capability to explain it implies support for either or both access modes
- The node service capability description contains "(subject to the processing rules for NodePublishVolume)", I left it out for the controller service capability description since these access modes are also used in other RPCs like CreateVolume where these rules don't apply

spec.md
chrishenzie: Fixed this.

spec.md
```diff
@@ -2224,15 +2246,26 @@ If the volume corresponding to the `volume_id` has already been published at the
 
 If this RPC failed, or the CO does not know if it failed or not, it MAY choose to call `NodePublishVolume` again, or choose to call `NodeUnpublishVolume`.
 
-This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE capability (i.e., `access_mode` is either `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`).
+This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE or SINGLE_NODE_MULTI_WRITER (see second table) access modes. Example MULTI_NODE access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```
jdef, line 2249:
```suggestion
This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE or SINGLE_NODE_MULTI_WRITER (see second table) access modes.
Example MULTI_NODE access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```
^ nit: one sentence per line, please

spec.md
```diff
@@ -2224,15 +2246,26 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node. Plugins MUST accept and allow use of the `SINGLE_NODE_WRITER` access mode (subject to the processing rules above), when either `SINGLE_NODE_SINGLE_WRITER` and/or `SINGLE_NODE_MULTI_WRITER` are supported, in order to permit older COs to continue working.
```
jdef, line 2268:
```suggestion
The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node.
Plugins MUST accept and allow use of the `SINGLE_NODE_WRITER` access mode (subject to the processing rules above), when either `SINGLE_NODE_SINGLE_WRITER` and/or `SINGLE_NODE_MULTI_WRITER` are supported, in order to permit older COs to continue working.
```
one sentence per line

spec.md
chrishenzie: Done

spec.md
chrishenzie: Done

spec.md
```diff
@@ -2224,15 +2246,28 @@ If the volume corresponding to the `volume_id` has already been published at the
...
 | MULTI_NODE     | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                 |
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
-(`Tn`: target path of the n-th `NodePublishVolume`, `Pn`: other arguments of the n-th `NodePublishVolume` except `secrets`)
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
```
gnufied, line 2265: For many volume types - I suspect they may not support consuming same volume with different capability on same node. For  example - would they allow mounting volume with different mount flags or would they allow consuming volume both as block and fs volume at same time?

So I am bit doubtful that if "P1!=P2" and "T1!=T2" - node-publish will be okay (assuming driver really enforces supplied capabilities).  I suspect this will lead to drivers ignoring certain capabilities and allowing node-publish anyways. It may be okay, but strictly speaking they are not spec compliant then.

csi.proto
```diff
@@ -420,6 +420,18 @@ message VolumeCapability {
       // Can be published as read/write at multiple nodes
       // simultaneously.
       MULTI_NODE_MULTI_WRITER = 5;
+
+      // Can only be published once as read/write at a single workload
```
humblec, line 424: @chrishenzie can we mount this same volume for READONLY in the same node or another NODE ?

bswartz: Access modes are meant to be a floor/minimum level of access requested. If a user wants a volume to have read-only ceiling/maximum level of access, that is enforced at NodePublish time using the readonly flag, which is completely orthagonal to access modes.

humblec: Sure, let me rephrase the clarification which I am looking for, suppose a SP driver declared only SINGLE_NODE_SINGLE_WRITER, can we still publish a volume in READONLY ? because minimum level of access is "allowed writing on single node" in this case.

bswartz: Yes the choice whether to ControllerPublish or NodePublish any given volume readonly ALWAYS lies with the CO. What the access mode means is that the SP MUST support write access if the CO chooses to request it. Access mode is a promise that the SP makes to the CO, but the CO doesn't have to take advantage of it.

spec.md
chrishenzie: We currently make the clarification that `Pn` refers to other arguments passed to `NodePublishVolume` except `secrets`. We could extend this to also include `volume_capability`. If we want to play it safe we could make this change just for the new table. Does this sound reasonable?

gnufied: I think the *except `secrets`* wording does not make it clear, if calling `NodePublish` multiple times on same node has to be done with same secret or secrets can be different. I always found that wording to be ambiguous, so when we add `volume_capability` to it - does it mean the capabilities has to be same?  If yes - may be we should add more fields to the table. Something like:

```
T1!=T2, P1!=P2, Cap1!=Cap2 -> FAILED_PRECONDITION
T1!=T2, P1!=P2, Cap1==Cap2 -> OK
```

spec.md
```diff
@@ -2224,15 +2246,28 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| Non MULTI_NODE                        | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+
+The `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes are intended to replace the `SINGLE_NODE_WRITER` access mode to clarify the number of writers for a volume on a single node.
+Plugins MUST accept and allow use of the `SINGLE_NODE_WRITER` access mode (subject to the processing rules above), when either `SINGLE_NODE_SINGLE_WRITER` and/or `SINGLE_NODE_MULTI_WRITER` are supported, in order to permit older COs to continue working.
```
gnufied, line 2271: Do we need update `NodePublish` errors? Currently it says:

> | Exceeds capabilities | 9 FAILED_PRECONDITION | Indicates that the CO has exceeded the volume's capabilities because the volume does not have MULTI_NODE capability. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |

chrishenzie: I would prefer not to introduce another error if possible. Since the condition is capabilities being exceeded, I think this description could be generalized to any time a CO requests capabilities not supported by the volume.

chrishenzie: Updated to use the error description for "Exceeds capabilities" used for ControllerExpandVolume and NodeExpandVolume which is more general:
> Indicates that CO has specified capabilities not supported by the volume.

Also updated this for NodeStageVolume since we used the old wording there as well.

spec.md
chrishenzie: Discussed this in the CSI standup, we're adding a sentence to make this explicit to avoid adding another column and more rows to the table. Added this sentence:
> COs SHOULD NOT call `NodePublishVolume` a second time with different `volume_capability`. If this happens, the Plugin SHOULD return `FAILED_PRECONDITION`.

chrishenzie: @bswartz for review on this

bswartz: +1 I like it

gnufied: Thanks. I think having explicit wording about capability helps. +1

saad-ali:
@jdef @humblec @gnufied PTAL

I'd like to merge and cut the next release on Monday.

spec.md
chhsia0: This seems okay at a first glance. At least this is the same as how Mesos shared persistent volumes work: http://mesos.apache.org/documentation/latest/shared-resources/

spec.md
```diff
@@ -2224,15 +2246,30 @@ If the volume corresponding to the `volume_id` has already been published at the
 
 If this RPC failed, or the CO does not know if it failed or not, it MAY choose to call `NodePublishVolume` again, or choose to call `NodeUnpublishVolume`.
 
-This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has MULTI_NODE capability (i.e., `access_mode` is either `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`).
+This RPC MAY be called by the CO multiple times on the same node for the same volume with possibly different `target_path` and/or other arguments if the volume has `MULTI_NODE` or `SINGLE_NODE_MULTI_WRITER` (see second table) access modes.
```
jdef, line 2249:
```suggestion
This RPC MAY be called by the CO multiple times on the same node for the same volume with a possibly different `target_path` and/or other arguments if the volume supports either `MULTI_NODE_...` or `SINGLE_NODE_MULTI_WRITER` access modes (see second table).
```

spec.md
```diff
@@ -2224,15 +2246,30 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+Example `MULTI_NODE` access modes include `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```
jdef, line 2250:
```suggestion
The possible `MULTI_NODE_...` access modes are `MULTI_NODE_READER_ONLY`, `MULTI_NODE_SINGLE_WRITER` or `MULTI_NODE_MULTI_WRITER`.
```

spec.md
```diff
@@ -2224,15 +2246,30 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+COs SHOULD NOT call `NodePublishVolume` a second time with different `volume_capability`.
```
jdef, line 2251:
```suggestion
COs SHOULD NOT call `NodePublishVolume` a second time with a different `volume_capability`.
```

spec.md
```diff
@@ -2224,15 +2246,30 @@ If the volume corresponding to the `volume_id` has already been published at the
...
+If this happens, the Plugin SHOULD return `FAILED_PRECONDITION`.
 The following table shows what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
 
+(`Tn`: target path of the n-th `NodePublishVolume`, `Pn`: other arguments of the n-th `NodePublishVolume` except `secrets`)
```
jdef, line 2255:
```suggestion
(**T<sub>n</sub>**: target path of the n<sup>th</sup> `NodePublishVolume`; **P<sub>n</sub>**: other arguments of the n<sup>th</sup> `NodePublishVolume` except `secrets`)
```

spec.md
```diff
@@ -2224,15 +2246,30 @@ If the volume corresponding to the `volume_id` has already been published at the
...
 | Non MULTI_NODE | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION|
 
-(`Tn`: target path of the n-th `NodePublishVolume`, `Pn`: other arguments of the n-th `NodePublishVolume` except `secrets`)
+NOTE: If the Plugin supports the `SINGLE_NODE_MULTI_WRITER` capability, use the following table instead for what the Plugin SHOULD return when receiving a second `NodePublishVolume` on the same volume on the same node:
+
+|                                       | T1=T2, P1=P2    | T1=T2, P1!=P2  | T1!=T2, P1=P2       | T1!=T2, P1!=P2      |
+|---------------------------------------|-----------------|----------------|---------------------|---------------------|
+| SINGLE_NODE_SINGLE_WRITER             | OK (idempotent) | ALREADY_EXISTS | FAILED_PRECONDITION | FAILED_PRECONDITION |
+| SINGLE_NODE_MULTI_WRITER              | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
+| MULTI_NODE                            | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
```
jdef, line 2268:
```suggestion
| MULTI_NODE_...                        | OK (idempotent) | ALREADY_EXISTS | OK                  | OK                  |
```

spec.md
```diff
@@ -2313,7 +2350,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
 | Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
 | Volume published but is incompatible | 6 ALREADY_EXISTS | Indicates that a volume corresponding to the specified `volume_id` has already been published at the specified `target_path` but is incompatible with the specified `volume_capability` or `readonly` flag. | Caller MUST fix the arguments before retrying. |
-| Exceeds capabilities | 9 FAILED_PRECONDITION | Indicates that the CO has exceeded the volume's capabilities because the volume does not have MULTI_NODE capability. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |
+| Exceeds capabilities | 9 FAILED_PRECONDITION | Indicates that CO has specified capabilities not supported by the volume. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |
```
jdef, line 2353:
```suggestion
| Exceeds capabilities | 9 FAILED_PRECONDITION | Indicates that the CO has specified capabilities not supported by the volume. | Caller MAY choose to call `ValidateVolumeCapabilities` to validate the volume capabilities, or wait for the volume to be unpublished on the node. |
```

jdef (commented):
LGTM mod comments

spec.md
```diff
@@ -887,6 +887,18 @@ message VolumeCapability {
       // Can be published as read/write at multiple nodes
       // simultaneously.
       MULTI_NODE_MULTI_WRITER = 5;
+
+      // Can only be published once as read/write at a single workload
+      // on a single node, at any given time. Should be used instead of
```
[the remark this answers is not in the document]
chrishenzie: Yes, thanks for the heads up

chrishenzie:
@jdef Done, addressed comments. Thank you for the review!

csi.proto
```diff
@@ -420,6 +420,18 @@ message VolumeCapability {
...
+      // on a single node, at any given time. SHOULD be used instead of
+      // SINGLE_NODE_WRITER for COs using the experimental
+      // SINGLE_NODE_MULTI_WRITER capability.
```
humblec, line 427: @chrishenzie this should be `SINGLE_NODE_SINGLE_WRITER` ?

spec.md
humblec: @chrishenzie just to understand in bit more depth, but for COs with controller service operations, its just about knowing `SINGLE_NODE` or `MULTI_NODE` , isnt it ?  in that sense, `writer` thing ( single or multiple) is not relevant for CO's  control plane operations ?

csi.proto
chrishenzie: Both `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` are replacements for `SINGLE_NODE_WRITER` depending on the number of intended writers. The capability `SINGLE_NODE_MULTI_WRITER` implies support for `SINGLE_NODE_MULTI_WRITER` and/or `SINGLE_NODE_SINGLE_WRITER`. The reason for naming the capability like so was because it was the access mode that was missing from the CSI spec.

spec.md
chrishenzie: Yes, the number of writers here does not make a difference.

csi.proto
humblec: `The capability SINGLE_NODE_MULTI_WRITER implies support for SINGLE_NODE_MULTI_WRITER and/or SINGLE_NODE_SINGLE_WRITER` -> yeah, this aligns with my thought too or what I meant with `replacement as a default` in above comment, so we are good to mention this over over here ! Thanks @chrishenzie. :+1:

humblec:
lgtm from end. Thanks :+1:

saad-ali:
> LGTM mod comments

> lgtm from end. Thanks 👍

Thanks folks. Mergining this PR.

PR #479, opened 2021-05-03
spec: add optional secrets to NodeGetVolumeStatsRequest

nixpanic:
For BlockVolume type of volumes, there are very little stats that can be
gathered. The only reasonable metric on the Node that is available
without knowing the contents of the BlockVolume, is the Total capacity.

Different backend storage systems can thin-provision BlockVolumes.
Initially the Used storage will be low, and growing over time. This is
not detectable through standard Operating System APIs, but the backend
storage system can provide more acurate Available/Used stats through
expert APIs.

In order to allow the Node to retrieve advanced stats from the backend
storage system, the secrets that are used for staging the volume can be
provided.


nixpanic:
Rebased on top of the master branch to resolve some conflicts. This is still something we're interested in so that Ceph-CSI can detect the actual consumed space of (thin-provisioned) BlockVolumes instead of the size of the block-device.

Issue #486, opened 2021-07-21
published_node_ids has different requirements in ListVolumes vs ControlerGetVolume


bswartz:
In ListVolumesResponse the docs read:

    // A list of all `node_id` of nodes that the volume in this entry
    // is controller published on.
    // This field is OPTIONAL. If it is not specified and the SP has
    // the LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO
    // MAY assume the volume is not controller published to any nodes.
    // If the field is not specified and the SP does not have the
    // LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO MUST
    // not interpret this field.
    // published_node_ids MAY include nodes not published to or
    // reported by the SP. The CO MUST be resilient to that.
    repeated string published_node_ids = 1;

In ControllerGetVolume, the docs read:

    // A list of all the `node_id` of nodes that this volume is
    // controller published on.
    // This field is OPTIONAL.
    // This field MUST be specified if the PUBLISH_UNPUBLISH_VOLUME
    // controller capability is supported.
    // published_node_ids MAY include nodes not published to or
    // reported by the SP. The CO MUST be resilient to that.
    repeated string published_node_ids = 1;

For a driver wants to implement LIST_VOLUMES, GET_VOLUME, and PUBLISH_UNPUBLISH_VOLUME, but not LIST_VOLUMES_PUBLISHED_NODES, this creates an impossible situation because you have to populate the field but you don't have anything to populate it with.

I think we we simply changed the wording in the ControllerGetVolume case to refer to the LIST_VOLUMES_PUBLISHED_NODES capability instead of the PUBLISH_UNPUBLISH_VOLUME capability, then the conflict would be avoided. In that case LIST_VOLUMES_PUBLISHED_NODES would be the capability that specified COs can rely on published_node_ids in both ControllerGetVolume and ListVolumes. It's unfortunate that the capability has the word "LIST" in the name, but it predates the new Get call so unless we want to add a whole other capability, it's best to just reuse this capability and apologize for the slightly confusing name.

PR #487, opened 2021-07-21
Change required capability for published_node_ids in GetVolume

bswartz:
Instead of PUBLISH_UNPUBLISH_VOLUME, use the LIST_VOLUMES_PUBLISHED_NODES
capability to control the published_node_ids field in ControllerGetVolume.
This change is to be more consistent with the existing behavior. It is
a backwards-incompatible change, but the API is alpha.

Also clarify the LIST_VOLUMES_PUBLISHED_NODES docs.

Fixes: #486

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -922,7 +922,7 @@ message ControllerGetVolumeResponse {
     // A list of all the `node_id` of nodes that this volume is
     // controller published on.
     // This field is OPTIONAL.
-    // This field MUST be specified if the PUBLISH_UNPUBLISH_VOLUME
+    // This field MUST be specified if the LIST_VOLUMES_PUBLISHED_NODES
     // controller capability is supported.
     // published_node_ids MAY include nodes not published to or
     // reported by the SP. The CO MUST be resilient to that.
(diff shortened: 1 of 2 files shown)


bswartz:
@saad-ali @xing-yang please take a look at this

xing-yang:
LGTM

jdef:
I wonder if it's also worth clarifying (in the docs for the capability) that LIST_VOLUMES_PUBLISHED_NODES requires PUBLISH_UNPUBLISH_VOLUME. While the careful reader may already observe this relationship, it might be worth calling out explicitly for the sake of clarity.

Also, given the confusing name, maybe also mention ControllerGetVolumeReponse in the docs for LIST_VOLUMES_PUBLISHED_NODES ?

bswartz:
> I wonder if it's also worth clarifying (in the docs for the capability) that LIST_VOLUMES_PUBLISHED_NODES requires PUBLISH_UNPUBLISH_VOLUME. While the careful reader may already observe this relationship, it might be worth calling out explicitly for the sake of clarity.
> 
> Also, given the confusing name, maybe also mention ControllerGetVolumeReponse in the docs for LIST_VOLUMES_PUBLISHED_NODES ?

I agree with both these.

bswartz:
@jdef Updated, PTAL.

jdef (approved):

saad-ali:
/lgtm
/approve

>> I wonder if it's also worth clarifying (in the docs for the capability) that LIST_VOLUMES_PUBLISHED_NODES requires PUBLISH_UNPUBLISH_VOLUME. While the careful reader may already observe this relationship, it might be worth calling out explicitly for the sake of clarity.
>>
>> Also, given the confusing name, maybe also mention ControllerGetVolumeReponse in the docs for LIST_VOLUMES_PUBLISHED_NODES ?
>
> I agree with both these.



Opened https://github.com/container-storage-interface/spec/issues/489 to track that.

PR #500, opened 2022-03-16
fix refresh of generated Go files

pohly:
"make build" did not rebuild csi.pb.go because the file existed and make wasn't
told that the file depends on spec.proto.


saad-ali:
@jdef can you PTAL?

jdef (approved):

PR #502, opened 2022-03-22
test: add github action build test

andyzhangx:
It seems that build test result won't show in the first github action PR, and it should show in the next PR after this PR merged.


saad-ali:
LGTM Thanks @andyzhangx 

@jdef can you PTAL?

CC @msau42

.github/workflows/build.yaml
```diff
@@ -0,0 +1,27 @@
...
+      with:
+        go-version: ^1.16
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build Test
+      run: |
+        make -C lib/go protoc
```
jdef, line 25:
```suggestion
        make
        make -C lib/go protoc
```
^ `make` as a separate step, before installing deps, to run tests. which works as expected if you also ...
1. update Makefile and lib/go/Makefile: s/TRAVIS/GITHUB_ACTIONS/
2. drop the .travis.yml file

andyzhangx: fixed the comments, PTAL.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        go-version: ^1.16
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build Test
+      env:
+        GITHUB_ACTIONS: true
```
jdef, line 25: `GITHUB_ACTIONS` is provided automatically, right? if not we can keep this here - otherwise it doesn't seem to add value

jdef (changes requested):

.github/workflows/build.yaml
andyzhangx: fixed

jdef (approved):
I think this looks OK. Sad that we cannot run it before merging

saad-ali:
Thank you both.
Let's merge and see.

jdef:
Looks like we missed the build badge on the readme

PR #499, opened 2022-03-15
GetCapacityResponse: promote maximum_volume_size

pohly:
[Kubernetes](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1472-storage-capacity-tracking/README.md) and several CSI drivers (for example, [PMEM-CSI](https://github.com/intel/pmem-csi/blob/8ed05e3a38eba355c1d57e00d38b3fd9c3559d70/pkg/pmem-csi-driver/controllerserver-node.go#L475-L486)) have been using that field for a while now
and no changes to the semantic are expected, therefore the alpha status can and
should be removed.

The minimum_volume_size field is not used by Kubernetes. The semantic is
probably stable, but further practical experience with it might be desirable
before promoting it.

PR #509, opened 2022-06-08
Move volume mount group feature GA

gnufied:
Drop alpha from volume_mount_group feature.


jdef:
is this the first time we've promoted something from alpha to GA? i'm wondering if there's a compatibility impact w/ respect to complete dropping the alpha annotations vs. setting them to `false`. maybe i'm overthinking this. has anyone tested interactions between older bindings (w/ alpha set) against newer bindings (w/o alpha)?

gnufied:
This is another instance https://github.com/container-storage-interface/spec/pull/499 when we moved things over from alpha to GA. I will look into compatibility aspects of when we move a feature from alpha to GA.

gnufied:
@jdef So I ran e2e of k8s by updating kubernetes with my version of CSI driver and let driver run a released version of CSI with alpha APIs. And all e2es are still passing:

```
  should pass FSGroup to CSI driver if it is set in pod and driver supports VOLUME_MOUNT_GROUP                                                                                                                                                               
    should not pass FSGroup to CSI driver if it is set in pod and driver does not supports VOLUME_MOUNT_GROUP                                                                                                                                                         
```

Is that enough as signal that dropping alpha annotation does not break anything?

jdef (approved):
LGTM

Issue #371, opened 2019-07-19
Secrets filed addition to NodeGetVolumeStats


humblec:
At present, there is no `secret` field available on NodeGetVolumeStats RPC, however this is required at times for specific storage vendor to map volumeID to the `targetPath`. 
 
As per the spec:
```
Condition | gRPC Code | Description | Recovery Behavior
-- | -- | -- | --
Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified volume_id does not exist on specified volume_path. | Caller MUST verify that the volume_id is correct and that the volume is accessible on specified volume_path and has not been deleted before retrying with exponential back off.
```
 
`targetpath` contains the pvc name and volumeID is generated by plugin. The mapping between the `targetpath` to `volumeID` (may) require the secret/credentials. For example, in Ceph, we need secret to connect to the ceph cluster to get the volumeID from volume Name.

gnufied:
@humblec if we have a valid `targetPath` - do we still need to verify if `targetPath` belongs to given volume ID? Does this apply to something like `NodeUnpublishVolume` too - which also does not have secrets in requst?

IIRC - @j-griffith was looking into a similar issue.

j-griffith:
@humblec could you please provide more detailed and specific information regarding this use case?  It's not entirely clear what the issue is here or that it can't be solved without secrets (or why).

saad-ali:
To summarize, we'd like to understand:
* Does Ceph for NodeGetVolumeStats actually require backend call (which would require a secret)?

j-griffith:
> @humblec could you please provide more detailed and specific information regarding this use case? It's not entirely clear what the issue is here or that it can't be solved without secrets (or why).

I was under the assumption that this is the same problem that has been discussed here:  https://github.com/container-storage-interface/spec/issues/370#issuecomment-519248111

ShyamsundarR:
> > @humblec could you please provide more detailed and specific information regarding this use case? It's not entirely clear what the issue is here or that it can't be solved without secrets (or why).
> 
> I was under the assumption that this is the same problem that has been discussed here: [#370 (comment)](https://github.com/container-storage-interface/spec/issues/370#issuecomment-519248111)

The problem here is different, the issue was how to validate the given `volume_id` is the one that is mapped to `volume_path` and hence return stats or return a `NOT_FOUND` error.

[...]

PR #515, opened 2022-06-16
add `secrets` field support for ControllerGetVolume Request

humblec:
At present `secrets` are not part of this request and some
consumers need this field  for its operation.: That said, 
storage backends like Ceph Cluster require the access to be
achieved with the help of authentication. For such systems this 
becomes a requirement to have secrets/credentials be part of 
the request CO make for its further operations or fetching the state
from the backend cluster.

Additional note for reviewer: 
Similar request was raised for NodeGetVolumeStats() here # 
https://github.com/container-storage-interface/spec/issues/371 
which was not reached into consensus. One concern raised in similar request (#371)
is the explosion of secrets field in the spec for various RPC calls. 
In turn this secret field requirement becoming a request for pretty
much all the RPC  calls.

  

csi.proto
```diff
@@ -913,6 +913,11 @@ message ControllerGetVolumeRequest {
   // The ID of the volume to fetch current volume information for.
   // This field is REQUIRED.
   string volume_id = 1;
+
+  // Secrets required by plugin to complete node expand volume request.
```
Madhu-1, line 917: node expand volume request?

xing-yang:
Do we also need a corresponding field in the PV?

humblec:
> Do we also need a corresponding field in the PV?

@xing-yang I was planning to have the PV/CO side ( implementation) details listed in Kube issue as the  field and implementation require some thoughts as discussed in https://github.com/container-storage-interface/spec/issues/371. Considering  CSI spec support can be independent to the implementation of CO,  thought of pulling this first.

humblec:
Cc @jdef

bswartz:
I agree that we need to pass secrets down to ControllerGetVolume, but I'd prefer to do it in such a way that doesn't require COs to store ever more secrets (as I mentioned in #371). Is anyone working on a proposal to constrain secret handling?

PR #517, opened 2022-07-12
Promote GetVolume and VolumeCondition to GA

xing-yang:
This PR promotes volume health monitoring related capabilities, RPCs, fields to GA.

Volume health monitoring is implemented in Kubernetes CSI External Health Monitor Controller and Kubelet.
The feature on the controller side is Alpha since Kubernetes v1.19.
The feature on the Kubelet side is Alpha since  Kubernetes v1.21.
We are planning to move both to Beta in Kubernetes v1.26.
https://github.com/kubernetes/enhancements/pull/3321
https://github.com/kubernetes-csi/external-health-monitor/releases

We want to move the corresponding feature in CSI spec to GA before that.


jdef:
same question as [on Jun 8](https://github.com/container-storage-interface/spec/pull/509#issuecomment-1150338555) re: backwards compat: should alpha really be deleted, or just set to "false" ?

humblec (approved):

xing-yang:
> same question as https://github.com/container-storage-interface/spec/pull/509#issuecomment-1150338555 re: backwards compat: should alpha really be deleted, or just set to "false" ?

@jdef regarding this question, I think it is safe to delete "alpha". After upgrading to a newer CSI spec with GetVolume and VolumeCondition without the alpha label, there won't be anything else in the system that still refers to the "alpha" API.

PR #518, opened 2022-07-21
Maybe add c# namespace into the spec

Zetanova:
c# users of the spec need to edit it to generate the source-code to include the csharp namespace.
If the option is not allowed to be included then please remove the comment 'DO NOT EDIT'

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -7,6 +7,7 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
 
 option go_package = "csi";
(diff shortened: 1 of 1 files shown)


nixpanic:
I think that should be fine, there is already a `go_package` option too.

Note that you should include this in `spec.md` instead, and regenerate the code (recreates `csi.proto`) from there.

Zetanova:
@nixpanic thx, the namespace got added to spec.md too

jdef:
Thanks for the PR. Please sign the CCLA if you haven't already

PR #522, opened 2022-09-19
Add VolumeSnapshotDelta CSI RPCs

ihcsim:
KEP: https://github.com/kubernetes/enhancements/pull/4082

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -5,7 +5,6 @@ package csi.v1;
 import "google/protobuf/descriptor.proto";
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
-import "google/protobuf/duration.proto";
 
 option go_package = "csi";
 
@@ -1246,11 +1245,11 @@ message ListSnapshotDeltasRequest {
   // The ID of the base snapshot handle to use for comparison. If
   // not specified, return all changed blocks up to the target
   // specified by snapshot_target. This field is OPTIONAL.
-  string snapshot_base_id = 1;
+  string from_snapshot_id = 1;
 
   // The ID of the target snapshot handle to use for comparison. If
   // not specified, an error is returned. This field is REQUIRED.
-  string snapshot_target_id = 2;
+  string to_snapshot_id = 2;
 
   // Defines the type of storage. Default to "BLOCK". This field is
   // REQUIRED.
@@ -1331,7 +1330,7 @@ message BlockSnapshotChangedBlockToken {
   // The TTL of the token in seconds. The expiry time is calculated by
   // adding the time of issuance with this value. This field is
   // REQUIRED.
-  .google.protobuf.Duration ttl_seconds = 3;
+  int32 ttl_seconds = 3;
 }
 
 message ControllerExpandVolumeRequest {
--- a/csi.proto
+++ b/csi.proto
@@ -1298,8 +1298,9 @@ message ListSnapshotDeltasResponse {
 message BlockSnapshotChangedBlock {
   option (alpha_message) = true;
 
-  // The block logical offset on the volume. This field is REQUIRED.
-  uint64 offset = 1;
+  // The block logical offset (in bytes) on the volume. This field is
+  // REQUIRED.
+  uint64 offset_bytes = 1;
 
   // The size of the block in bytes. This field is REQUIRED.
   uint64 block_size_bytes = 2;
--- a/csi.proto
+++ b/csi.proto
@@ -5,6 +5,7 @@ package csi.v1;
 import "google/protobuf/descriptor.proto";
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
+import "google/protobuf/duration.proto";
 
 option go_package = "csi";
 
@@ -99,6 +100,11 @@ service Controller {
   rpc ControllerGetVolume (ControllerGetVolumeRequest)
     returns (ControllerGetVolumeResponse) {
         option (alpha_method) = true;
+    }
+
+  rpc ListSnapshotDeltas(ListSnapshotDeltasRequest)
+    returns (ListSnapshotDeltasResponse) {
+      option (alpha_method) = true;
     }
 }
 
@@ -1080,6 +1086,10 @@ message ControllerServiceCapability {
       // SINGLE_NODE_SINGLE_WRITER and/or SINGLE_NODE_MULTI_WRITER are
       // supported, in order to permit older COs to continue working.
       SINGLE_NODE_MULTI_WRITER = 13 [(alpha_enum_value) = true];
+
+      // Indicates the SP can compute and retrieve the deltas between
+      // two snapshots.
+      LIST_SNAPSHOT_DELTAS = 14 [(alpha_enum_value) = true];
     }
 
     Type type = 1;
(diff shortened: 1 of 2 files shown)


jdef (commented):
Please provide more context in the description of this RPC. There's no need to regurgitate the KEP in its entirety, but it was difficult for me to understand the usefulness of this w/o reading through parts of the related KEP.

For example (and I'm struggling w/ how to phrase this next part, so please bear w/ me)..

Dataplane is out of scope of CSI - and at the same time, CSI has "publish" calls that feel very close to dataplane territory (maybe since mount operations can involve a bit of (meta)data xfer). This proposal is purely controlplane and folks w/o CBT experience may be left wondering how in the world a CO is supposed to make use of this metadata since there are no other CSI calls that point to the dataplane at all here. Its only half of the overall useful API surface area. Without going into k8s details, it may be useful for such readers if the docs here mentioned something about the dataplane API aspect (block transfer w/ respect to backups) and how it's the CO's job to facilitate access to the raw data.

spec.md
```diff
@@ -1997,6 +2006,132 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // unspecified if the volume isn't of block type. This field is
+  // OPTIONAL.
+  repeated BlockSnapshotChangedBlock changed_blocks = 3;
+}
+
+message BlockSnapshotChangedBlock {
+  option (alpha_message) = true;
+
+  // The block logical offset on the volume. This field is REQUIRED.
+  uint64 offset = 1;
```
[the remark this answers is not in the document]
ihcsim: Not necessarily. E.g., the AWS EBS Direct API uses its own logical offset scheme.

spec.md
```diff
@@ -1997,6 +2006,132 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // offset. This field is REQUIRED.
+  string token = 1;
+
+  // Timestamp when the token is issued. This field is REQUIRED.
+  .google.protobuf.Timestamp issuance_time = 2;
+
+  // The TTL of the token in seconds. The expiry time is calculated by
+  // adding the time of issuance with this value. This field is
+  // REQUIRED.
+  .google.protobuf.Duration ttl_seconds = 3;
```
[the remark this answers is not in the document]
ihcsim: Good point - will fix.

spec.md
```diff
@@ -1997,6 +2006,132 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+##### ListSnapshotDeltas Errors
+
+If the plugin is unable to complete the `ListSnapshotDeltas` call successfully,
+it MUST return a non-ok gRPC code in the gRPC status. If the conditions defined
+below are encountered, the plugin MUST return the specified gRPC error code. The
+CO MUST implement the specified error recovery behavior when it encounters the
+gRPC error code.
+
+| Condition | gRPC Code | Description | Recovery Behavior |
```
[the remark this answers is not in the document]
ihcsim: The intent is to let the SP do the error handling, and CSI just relays the response back to the end users. Different SP might handle this differently.

spec.md
```diff
@@ -1997,6 +2006,132 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // The ID of the base snapshot handle to use for comparison. If
+  // not specified, return all changed blocks up to the target
+  // specified by snapshot_target. This field is OPTIONAL.
+  string snapshot_base_id = 1;
+
+  // The ID of the target snapshot handle to use for comparison. If
+  // not specified, an error is returned. This field is REQUIRED.
+  string snapshot_target_id = 2;
+
+  // Defines the type of storage. Default to "BLOCK". This field is
```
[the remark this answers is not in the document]
ihcsim: I see what you mean. The original intent for UX is to let end users re-use the same CRD for both block and file storage. (The KEP work on file storage hasn't started yet.) I don't know enough SPs to say whether they can figure this out from the snapshot IDs, but I don't see why they can't. I will bring this up with the WG.

csi.proto
```diff
@@ -1227,6 +1236,103 @@ message ListSnapshotsResponse {
...
+  // restriction on the number of entries that can be returned.
+  // The value of this field MUST NOT be negative.
+  int32 max_entries = 4;
+
+  // A token to specify where to start paginating. Set this field to
+  // `next_token` returned by a previous `ListSnapshotDeltas` call to
+  // get the next page of entries. This field is OPTIONAL.
+  // An empty string is equal to an unspecified field value.
+  string starting_token = 5;
+}
```
humblec, line 1281: @ihcsim i think its good to have secretRefs in this call.

spec.md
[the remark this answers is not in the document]
ihcsim: The WG has agreed to remove this field.

spec.md
```diff
@@ -1997,6 +2006,133 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+If the plugin is unable to complete the `ListSnapshotDeltas` call successfully,
+it MUST return a non-ok gRPC code in the gRPC status. If the conditions defined
+below are encountered, the plugin MUST return the specified gRPC error code. The
+CO MUST implement the specified error recovery behavior when it encounters the
+gRPC error code.
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Target volume snapshot does not exist | 5 NOT_FOUND | Indicates that the target volume snapshot corresponding to the specified `to_snapshot_id` does not exist. | Caller MUST verify that the `to_snapshot_id` is correct and that the volume snapshot is accessible and has not been deleted before retrying with exponential back off. |
```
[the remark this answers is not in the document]
ihcsim: In that case, the delta between the first snapshot since volume creation and the target snapshot will be retrieved.

spec.md
```diff
@@ -1997,6 +2006,133 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  map<string, string> secrets = 3 [(csi_secret) = true];
+
+  // If specified (non-zero value), the Plugin MUST NOT return more
+  // entries than this number in the response. If the actual number of
+  // entries is more than this number, the Plugin MUST set `next_token`
+  // in the response which can be used to get the next page of entries
+  // in the subsequent `ListSnapshotDeltas` call. This field is
+  // OPTIONAL. If not specified (zero value), it means there is no
+  // restriction on the number of entries that can be returned.
+  // The value of this field MUST NOT be negative.
```
nixpanic, line 2057: if this value MUST NOT be negative, why not use an unsigned int?

nixpanic, line 2057: if this value MUST NOT be negative, why not use an unsigned int?

spec.md
```diff
@@ -1997,6 +2006,133 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+to retrieve the raw data blocks. The end users are responsbile for devising
+their own transport and retrieval mechanism to access the raw data blocks.
+
+```protobuf
+message ListSnapshotDeltasRequest {
+  option (alpha_message) = true;
+
+  // The ID of the base snapshot handle to use for comparison. If
+  // not specified, return all changed blocks up to the target
+  // specified by snapshot_target. This field is OPTIONAL.
```
nixpanic, line 2037: `snapshot_target` is not part of this message, I think you mean `to_snapshot_id`.

spec.md
```diff
@@ -1997,6 +2006,133 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+results, then it is possible that the CO MAY either witness duplicate results in
+the list, not witness existing deltas, or both. The CO SHALL NOT expect a
+consistent "view" of all snapshot deltas when paging through the snapshot list
+via multiple calls to `ListSnapshotDeltas`.
+
+A Controller Plugin MUST NOT attempt to include any raw data blocks in the RPC
+response to ensure data plane operations remain outside the scope of CSI. The
+response payload MUST contain enough block location metadata for the end users
+to retrieve the raw data blocks. The end users are responsbile for devising
+their own transport and retrieval mechanism to access the raw data blocks.
```
nixpanic, line 2029: Without an option to transport changed blocks, I am not sure how useful this is. Users would still need to talk to the specific SP directly, and that is exactly what CSI tries to prevent.

ihcsim: At this point, data path remains outside the scope of CSI. The data protection white paper describes a SP-agnostic data movement mechanism [here][1]. Our observation is that different providers have different data movement implementation, which makes coming up with generic APIs challenging. 

[1]: https://github.com/kubernetes/community/blob/master/wg-data-protection/data-protection-workflows-white-paper.md#change-block-tracking

spec.md
ihcsim: This the convention used for other definitions of `max_entries` in the spec.

ihcsim:
Superseded by #551.

PR #523, opened 2022-09-22
control groups has to be represented as cgroups and not cgroup

humblec:

- [x] s/cgorup/cgroup in spec
- [x] corrected table alignment
- [x] corrected few typos


humblec:
Cc @jdef

jdef:
Lgtm, thanks

PR #524, opened 2022-09-22
use go v1.19 and update dependencies

humblec:


humblec:
cc @saad-ali

go.mod
```diff
@@ -1,8 +1,16 @@
 module github.com/container-storage-interface/spec
 
-go 1.16
+go 1.19
```
jdef, line 3: is this version actually required to build CSI (w/ reasonable deps)? if not, why not bump to 1.17 or 1.18 instead? CSI module doesn't need to be bleeding edge (and thereby force downstream users to use the latest Go)

humblec: sure @jdef , I have declared it on 1.18 in the last refresh.

jdef (approved):

jdef:
@saad-ali any opinions here? if not, i'll merge as is

saad-ali:
No objections. LGTM

PR #529, opened 2022-10-30
Add msg to clarify usage

xing-yang:
**What type of PR is this?**
This PR adds a message to clarify how VOLUME_CONDITION is used.

Fixes #

```release-note
none
```

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -1056,6 +1056,8 @@ message ControllerServiceCapability {
       // If for some reason Controller and Node Plugins report
       // misaligned volume conditions, CO SHALL assume the worst case
       // is the truth.
+      // Note that `VolumeCondition` can be used to generate events and
+      // metrics.
       VOLUME_CONDITION = 11;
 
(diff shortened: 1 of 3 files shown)


xing-yang:
@jdef Please take a look.  Thanks.

csi.proto
```diff
@@ -1056,6 +1056,8 @@ message ControllerServiceCapability {
       // If for some reason Controller and Node Plugins report
       // misaligned volume conditions, CO SHALL assume the worst case
       // is the truth.
+      // Note that `VolumeCondition` can be used to generate events and
```
jdef, line 1059: "events" .. smells like k8s lingo.

also, I specifically called out use of the message field of VolumeCondition. and how it's not clear what the intended use is for it. can you please attempt to clarify? we cannot assume k8s semantics/terminology in this spec - it should be very clear to all users of all COs what `message` means. and whether it is for humans or machines to interpret

xing-yang:
Closing this one.  See this: https://github.com/container-storage-interface/spec/pull/530

PR #545, opened 2023-06-14
[WIP] Update spec doc for GroupController service

carlory:
**What type of PR is this?**

/kind document

```release-note
none
```


spec.md
```diff
@@ -268,7 +268,13 @@ Each SP MUST provide:
 
 * **Node Plugin**: A gRPC endpoint serving CSI RPCs that MUST be run on the Node whereupon an SP-provisioned volume will be published.
 * **Controller Plugin**: A gRPC endpoint serving CSI RPCs that MAY be run anywhere.
-* In some circumstances a single gRPC endpoint MAY serve all CSI RPCs (see Figure 3 in [Architecture](#architecture)).
+
+Each SP MAY provide:
+
+- **GroupController Plugin**: A gRPC endpoint serving CSI RPCs that MAY be run anywhere.
```
carlory, line 274: Hi @xing-yang,  Is the `GroupController plugin` optional or required for a SP?

spec.md
```diff
@@ -268,7 +268,13 @@ Each SP MUST provide:
...
+
+
+In some circumstances a single gRPC endpoint MAY serve all CSI RPCs (see Figure 3 in [Architecture](#architecture)).
```
carlory, line 277: I'll update related figures in Architecture.

spec.md
xing-yang: It is optional.

carlory: thanks

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
+|                  +-------->     Plugin       |  |
+|                           +------------------+  |
+|                                                 |
++-------------------------------------------------+
 
 Figure 2: Headless Plugin deployment, only the CO Node hosts run
-Plugins. Separate, split-component Plugins supply the Controller
-Service and the Node Service respectively.
+Plugins. Separate, split-component Plugins supply the Node Service,
+the Controller Service and the Group Controller Service (Optional)
```
jdef, line 137:
```suggestion
the Controller Service and the Group Controller Service (optional)
```

spec.md
```diff
@@ -322,10 +339,11 @@ extend google.protobuf.ServiceOptions {
 }
 ```
 
-There are three sets of RPCs:
+There are four sets of RPCs:
 
-* **Identity Service**: Both the Node Plugin and the Controller Plugin MUST implement this sets of RPCs.
+* **Identity Service**: Both the Controller Plugin, the Group Controller Plugin and the Node Plugin MUST implement this sets of RPCs.
```
jdef, line 344:
```suggestion
* **Identity Service**: Every Controller Plugin, Group Controller Plugin and Node Plugin MUST implement this sets of RPCs.
```

carlory:
A new change is that the `RPC Interactions` section is added.

cc @jdef @xing-yang

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
+|  |     CO     |   gRPC    |       Node       |  |
+|  |            +----------->      Plugin      |  |
+|  +------------+           +------------------+  |
+|                                                 |
++-------------------------------------------------+
 
 Figure 1: The Plugin runs on all nodes in the cluster: a centralized
-Controller Plugin is available on the CO master host and the Node
-Plugin is available on all of the CO Nodes.
+Controller Plugin and Group Controller Plugin (optional) are available
```
xing-yang, line 109: "Control Plugin" is also optional.
I'd suggest change "(optional)" to "(alpha)".

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
+|                  +-------->     Plugin       |  |
+|                           +------------------+  |
+|                                                 |
++-------------------------------------------------+
 
 Figure 2: Headless Plugin deployment, only the CO Node hosts run
-Plugins. Separate, split-component Plugins supply the Controller
-Service and the Node Service respectively.
+Plugins. Separate, split-component Plugins supply the Node Service,
+the Controller Service and the Group Controller Service (optional)
```
xing-yang, line 137: optional -> alpha

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
+|  +------------+           |     Plugin       |  |
+|                           +------------------+  |
+|                                                 |
++-------------------------------------------------+
 
 Figure 3: Headless Plugin deployment, only the CO Node hosts run
-Plugins. A unified Plugin component supplies both the Controller
-Service and Node Service.
+Plugins. A unified Plugin component supplies both the Node Service,
+the Controller Service and the Group Controller Service (optional).
```
xing-yang, line 156: optional -> alpha

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
-+-------------------------------------------+
+                                  CO "Node" Host(s)
++-------------------------------------------------+
+|                                                 |
+|  +------------+           +------------------+  |
+|  |     CO     |   gRPC    |       Node       |  |
+|  |            +----------->      Plugin      |  |
+|  +------------+           +------------------+  |
+|                                                 |
++-------------------------------------------------+
```
xing-yang, line 168: Is it necessary to update this drawing?

spec.md
```diff
@@ -268,7 +280,12 @@ Each SP MUST provide:
 
 * **Node Plugin**: A gRPC endpoint serving CSI RPCs that MUST be run on the Node whereupon an SP-provisioned volume will be published.
 * **Controller Plugin**: A gRPC endpoint serving CSI RPCs that MAY be run anywhere.
-* In some circumstances a single gRPC endpoint MAY serve all CSI RPCs (see Figure 3 in [Architecture](#architecture)).
+
+Each SP MAY provide:
```
xing-yang, line 284: Line 284 is duplicate of line 279.

spec.md
```diff
@@ -3066,6 +3084,20 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Snapshot list mismatch | 3 INVALID_ARGUMENT | Besides the general cases, this code SHOULD also be used to indicate when plugin supporting CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT detects a mismatch in the `snapshot_ids`. | If a mismatch is detected in the `snapshot_ids`, caller SHOULD use different `snapshot_ids`. |
 | Volume group snapshot does not exist | 5 NOT_FOUND | Indicates that a volume group snapshot corresponding to the specified `group_snapshot_id` does not exist. | Caller MUST verify that the `group_snapshot_id` is correct and that the volume group snapshot is accessible and has not been deleted before retrying with exponential back off. |
 
+#### RPC Interactions
+
+##### `CreateVolumeGroupSnapshot`, `DeleteVolumeGroupSnapshot`, `GetVolumeGroupSnapshot`
+
+The plugin-generated `group_snapshot_id` is a REQUIRED field for both the `DeleteVolumeGroupSnapshot` RPC and the `GetVolumeGroupSnapshot` PRC, as opposed to the CO-generated snapshot `name` that is REQUIRED for the `CreateVolumeGroupSnapshot` RPC.
```
xing-yang, line 3091: snapshot_ids is also REQUIRED.

spec.md
carlory: No, this change is only for the unified width as same as the other in this chapter.

spec.md
carlory: do you mean that source_volume_ids is also REQUIRED?

spec.md
```diff
@@ -264,11 +276,13 @@ This section describes the interface between COs and Plugins.
 ### RPC Interface
 
 A CO interacts with an Plugin through RPCs.
-Each SP MUST provide:
+Each SP MAY provide:
```
xing-yang, line 279: Node plugin is a MUST have.

spec.md
```diff
@@ -3066,6 +3081,20 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Snapshot list mismatch | 3 INVALID_ARGUMENT | Besides the general cases, this code SHOULD also be used to indicate when plugin supporting CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT detects a mismatch in the `snapshot_ids`. | If a mismatch is detected in the `snapshot_ids`, caller SHOULD use different `snapshot_ids`. |
 | Volume group snapshot does not exist | 5 NOT_FOUND | Indicates that a volume group snapshot corresponding to the specified `group_snapshot_id` does not exist. | Caller MUST verify that the `group_snapshot_id` is correct and that the volume group snapshot is accessible and has not been deleted before retrying with exponential back off. |
 
+#### RPC Interactions
+
+##### `CreateVolumeGroupSnapshot`, `DeleteVolumeGroupSnapshot`, `GetVolumeGroupSnapshot`
+
+The plugin-generated `group_snapshot_id` and `source_volume_ids` are a REQUIRED field for both the `DeleteVolumeGroupSnapshot` RPC and the `GetVolumeGroupSnapshot` PRC, as opposed to the CO-generated group snapshot `name` that is REQUIRED for the `CreateVolumeGroupSnapshot` RPC and the CO-generated volume  `name` that is REQUIRED for the `CreateVolume` RPC.
```
xing-yang, line 3088: `group_snapshot_id` and `snapshot_ids` are REQUIRED.
Also there's no need to mention `CreateVolume` here.

carlory:
@xing-yang updated. please review again.

spec.md
```diff
@@ -3066,6 +3085,20 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+The plugin-generated `group_snapshot_id` and `snapshot_ids` are a REQUIRED field for both the `DeleteVolumeGroupSnapshot` RPC and the `GetVolumeGroupSnapshot` PRC.
+
+A `CreateVolumeGroupSnapshot` operation SHOULD return with a `group_snapshot_id` when the group snapshot is cut successfully. If a `CreateVolumeGroupSnapshot` operation times out before the group snapshot is cut, leaving the CO without an ID with which to reference a group snapshot, and the CO also decides that it no longer needs/wants the group snapshot in question then the CO MAY choose one of the following paths:
+
+1. Retry the `CreateVolumeGroupSnapshot` RPC to possibly obtain a group snapshot ID that may be used to execute a `DeleteVolumeGroupSnapshot` RPC; upon success execute `DeleteVolumeGroupSnapshot`. If the `CreateVolumeGroupSnapshot` RPC returns a server-side gRPC error, it means that SP do clean up and make sure no snapshots are leaked.
+
+2. The CO takes no further action regarding the timed out RPC, a group snapshot is possibly leaked and the operator/user is expected to clean up. But this way isn't considered as a good practice.
```
bswartz, line 3098: This is bad guidance, and we can't recommend this in the spec. We need to be clear about what the CO must do and what is optional, and for optional things, how the SP can ensure correct behavior under either option.

bswartz (changes requested):

spec.md
```diff
@@ -80,85 +80,97 @@ A CO SHOULD be equipped to handle both centralized and headless plugins, as well
...
+|  |     CO     |   gRPC    |       Node       |  |
+|  |            +----------->      Plugin      |  |
+|  +------------+           +------------------+  |
+|                                                 |
++-------------------------------------------------+
 
 Figure 1: The Plugin runs on all nodes in the cluster: a centralized
-Controller Plugin is available on the CO master host and the Node
-Plugin is available on all of the CO Nodes.
+Controller Plugin (alpha) and Group Controller Plugin (alpha) are available
```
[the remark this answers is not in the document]
carlory: Hi @xing-yang do you mean like this?

https://github.com/container-storage-interface/spec/pull/545#discussion_r1235745577

```suggestion
Controller Plugin (optional) and Group Controller Plugin (alpha) are available
```

spec.md
carlory: There is a similar expression in the spec.

The following content is excerpted from PRC interactions of the controller service. 

```
2. The CO takes no further action regarding the timed out RPC, a snapshot is possibly leaked 
and the operator/user is expected to clean up.
```

spec.md
```diff
@@ -3066,6 +3085,20 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
 
+#### RPC Interactions
+
+##### `CreateVolumeGroupSnapshot`, `DeleteVolumeGroupSnapshot`, `GetVolumeGroupSnapshot`
+
+The plugin-generated `group_snapshot_id` and `snapshot_ids` are a REQUIRED field for both the `DeleteVolumeGroupSnapshot` RPC and the `GetVolumeGroupSnapshot` PRC.
+
+A `CreateVolumeGroupSnapshot` operation SHOULD return with a `group_snapshot_id` when the group snapshot is cut successfully. If a `CreateVolumeGroupSnapshot` operation times out before the group snapshot is cut, leaving the CO without an ID with which to reference a group snapshot, and the CO also decides that it no longer needs/wants the group snapshot in question then the CO MAY choose one of the following paths:
+
+1. Retry the `CreateVolumeGroupSnapshot` RPC to possibly obtain a group snapshot ID that may be used to execute a `DeleteVolumeGroupSnapshot` RPC; upon success execute `DeleteVolumeGroupSnapshot`. If the `CreateVolumeGroupSnapshot` RPC returns a server-side gRPC error, it means that SP do clean up and make sure no snapshots are leaked.
```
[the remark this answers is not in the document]
carlory: If the implementation of the SP's group snapshot function is flawed, for example, if two data volumes are snapshotted at the same time, one will succeed and the other will always fail, then no matter how many times CO retries the call, it will not get the expected results. It maybe lead to a deadlock in CO. But in the spec, I can't find a suitable status code to tell CO not to retry.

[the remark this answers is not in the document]
carlory: If we can assume that the SP implementation can handle the above situation, then the following description is not needed.

```
If the `CreateVolumeGroupSnapshot` RPC returns a server-side gRPC error, it means that SP do clean up and make sure no snapshots are leaked.
```

spec.md
bswartz: Thanks for pointing this out. We should probably clean up the language for volume snapshots too. It should be fine to duplicate the existing language from volume snapshots to group snapshots. I just want to make sure the language we're copying it up to date and accurate.

spec.md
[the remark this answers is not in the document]
xing-yang: How do we mark a snapshot as "broken" or "partial"?  Currently we don't have a way to do that and communicate back to CO.

spec.md
xing-yang: Agreed that this sentence regarding volume snapshot should be updated. SP should clean up the snapshot in this case.

spec.md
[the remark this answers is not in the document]
carlory: @bswartz @xing-yang 

I drop the sentence `If the CreateVolumeGroupSnapshot RPC returns a server-side gRPC error, it means that SP do clean up and make sure no snapshots are leaked.` and open an issue https://github.com/container-storage-interface/spec/issues/554 to track it.

saad-ali:
Confirmed @carlory signed CSI CLA

saad-ali:
>> How do we mark a snapshot as "broken" or "partial"? Currently we don't have a way to do that and communicate back to CO.
>
> I don't necessarily recommend this, but we could add an additional boolean field in the response message to indicate it.
>
> It's a unique problem for group snapshots because if you have multiple volumes and most of the snapshots were okay, but one cannot succeed, there are good arguments both to preserve the partial group snapshot and arguments to delete it and try over again. My personal inclination would to be to delete it -- in which case the SP could do that work with no RPC changes. But if we return success with a "partial" boolean set then the CO could also delete it, while also having the option to not delete it.


Per @xing-yang should fail and clean up if even one snapshot in group fails because they have to be consistent.  
Per @bswartz if SP gets 9/10 it is responsible for error handling.

@xing-yang will take another look at this PR.

spec.md
```diff
@@ -3066,6 +3085,20 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+##### `CreateVolumeGroupSnapshot`, `DeleteVolumeGroupSnapshot`, `GetVolumeGroupSnapshot`
+
+The plugin-generated `group_snapshot_id` and `snapshot_ids` are a REQUIRED field for both the `DeleteVolumeGroupSnapshot` RPC and the `GetVolumeGroupSnapshot` PRC.
+
+A `CreateVolumeGroupSnapshot` operation SHOULD return with a `group_snapshot_id` when the group snapshot is cut successfully. If a `CreateVolumeGroupSnapshot` operation times out before the group snapshot is cut, leaving the CO without an ID with which to reference a group snapshot, and the CO also decides that it no longer needs/wants the group snapshot in question then the CO MAY choose one of the following paths:
+
+1. Retry the `CreateVolumeGroupSnapshot` RPC to possibly obtain a group snapshot ID that may be used to execute a `DeleteVolumeGroupSnapshot` RPC; upon success execute `DeleteVolumeGroupSnapshot`.
+
+2. The CO takes no further action regarding the timed out RPC, a group snapshot is possibly leaked and the operator/user is expected to clean up.
```
saad-ali, line 3098: Per @bswartz and @xing-yang - It is not possible to tell k8s partial completion. And if partial completion, it is SP responsibility to clean up. It's all or nothing, if it's not all, SP must clean up.

PR #551, opened 2023-08-18
Added the SnapshotMetadata service.

carlbraganza:
**What type of PR is this?**
This is a proposed addition to the CSI specification.

It adds a SnapshotMetadata service through which one can obtain allocated or changed block metadata for snapshots.

n/a

Derived from https://github.com/kubernetes/enhancements/pull/4082

```release-note
NONE
```


xing-yang:
What about this PR https://github.com/container-storage-interface/spec/pull/522?

carlbraganza:
> What about this PR #522?

It has been closed.

csi.proto
```diff
@@ -219,6 +229,14 @@ message PluginCapability {
...
       // GroupControllerGetCapabilities.
       GROUP_CONTROLLER_SERVICE = 3 [(alpha_enum_value) = true];
+
+      // SNAPSHOT_METADATA_SERVICE indicates that the Plugin provides
+      // RPCs to retrieve metadata on the allocated blocks of a single
+      // snapshot, or the changed blocks between a pair of snapshots of
+      // the same block volume.
+      // The presence of this capability determines whether the CO will
+      // attempt to invoke the OPTIONAL SnapshotMetadata service RPCs.
+      SNAPSHOT_METADATA_SERVICE = 4 [(alpha_enum_value) = true];
```
carlbraganza, line 239: I blindly copied this pattern for the SnapshotMetadata service.

Does this imply that if an SP wants to use a separate container for the SnapshotMetadata service that they'd have to also implement an Identity service in this container?

spec.md
```diff
@@ -629,6 +639,14 @@ message PluginCapability {
...
       // GroupControllerGetCapabilities.
       GROUP_CONTROLLER_SERVICE = 3 [(alpha_enum_value) = true];
+
+      // SNAPSHOT_METADATA_SERVICE indicates that the Plugin provides
+      // RPCs to retrieve metadata on the allocated blocks of a single
+      // snapshot, or the changed blocks between a pair of snapshots of
+      // the same block volume.
+      // The presence of this capability determines whether the CO will
+      // attempt to invoke the OPTIONAL SnapshotMetadata service RPCs.
+      SNAPSHOT_METADATA_SERVICE = 4 [(alpha_enum_value) = true];
```
carlbraganza, line 649: I blindly copied this pattern for the SnapshotMetadata service as it is an optional service.

Does this imply that if an SP wants to use a separate container for the SnapshotMetadata service that they'd have to also implement an Identity service in this container?

If this was indeed the case then the Kubernetes CO implementation (sidecar) could check for this.

xing-yang:
cc @bswartz

xing-yang:
cc @jdef

spec.md
```diff
@@ -3149,6 +3167,184 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+##### GetDelta Errors
+If the plugin is unable to complete the `GetDelta` call successfully it
+must return a non-OK gRPC code in the gRPC status.
+
+The following conditions are well defined:
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Missing or otherwise invalid argument | 3 INVALID_ARGUMENT | Indicates that a required argument field was not specified or an argument value is invalid | The caller should correct the error and resubmit the call. |
```
nixpanic, line 3344: Should there be an error indicating that the original volume was deleted and the snapshots can not be tracked back? As CSI Snapshots are independent objects, a SP may not be able to link the snapshot back to the original volume and compute the diff between two snapshots. This can be seen as a requirement on the CSI-driver implementations, but an error for it might be useful.

spec.md
```diff
@@ -3149,6 +3167,184 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+#### GetAllocated RPC
+
+The plugin must implement this RPC if it provides the SnapshotMetadata service.
+
+```protobuf
+// The GetAllocatedRequest message is used to solicit metadata on the
+// allocated blocks of a snapshot: i.e. this identifies the data ranges
+// that have valid data as they were the target of some previous write
+// operation on the volume.
+message GetAllocatedRequest {
```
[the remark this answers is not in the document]
PrasadG193: I think that makes sense. Many providers might be using this for volume operations @carlbraganza wdyt?
We also might want to add a way to pass the provider-specific params just like the way it's done with `CreateVolumeRequest`?

```
  // Plugin specific creation-time parameters passed in as opaque 
  // key-value pairs. This field is OPTIONAL. The Plugin is responsible 
  // for parsing and validating these parameters. COs will treat 
  // these as opaque.
  map<string, string> parameters = 4;
```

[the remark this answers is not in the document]
nixpanic: Parameters are probably not required, as the CSI-driver should be able to fetch those based on the `snapshot_id`.

spec.md
```diff
@@ -3149,6 +3167,184 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+// ranges in the snapshot.
+message GetAllocatedResponse {
+  // This specifies the style used in the BlockMetadata sequence.
+  // This value must be the same in all such messages returned by
+  // the stream.
+  BlockMetadataType block_metadata_type = 1;
+
+  // This returns the size of the underlying volume.
+  // This value must be the same in all such messages returned by
+  // the stream.
```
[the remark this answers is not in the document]
carlbraganza: The intent here is to return the "logical" volume size. I guess it would be the value returned in the `Volume` property of the `CreateVolumeResponse`:

> message Volume {
  // The capacity of the volume in bytes. This field is OPTIONAL. If not
  // set (value of 0), it indicates that the capacity of the volume is
  // unknown (e.g., NFS share).
  // The value of this field MUST NOT be negative.
  int64 capacity_bytes = 1;

spec.md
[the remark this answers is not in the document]
carlbraganza: @nixpanic In K8s, does the secret data come from the `csi.storage.k8s.io/provisioner-secret-[name|namespace]` fields of the `VolumeSnapshotClass.Spec.Parameters`?

spec.md
```diff
@@ -3149,6 +3167,184 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+// The GetDeltaRequest message is used to solicit metadata on the data
+// ranges that have changed between two snapshots.
+message GetDeltaRequest {
+  // This is the identifier of the snapshot against which changes
+  // are to be computed.
+  string base_snapshot_id = 1;
+
+  // This is the identifier of a second snapshot in the same volume,
+  // created after the base snapshot.
+  string target_snapshot_id = 2;
```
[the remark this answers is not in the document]
ihcsim: While I can see `base_snapshot_id` being optional, I'd think `target_snapshot_id` is required. So far, the CBT implementation we have seen use snapshots to compute the diff. How does Ceph do it? Does it perform CBT computation on snapshots, on snapshot against live volumes (to get the volume current state) etc.?

spec.md
ihcsim: Good point. There can also be other reasons why the metadata service can't compute the diff between the two snapshots. Can we introduce a more generic error here, instead of just one pertaining to invalid snapshot-to-volume mapping (due to missing volume)?

spec.md
carlbraganza: Feedback I received verbally on this is that every CSI server process must provide an Identity service.

spec.md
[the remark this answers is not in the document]
carlbraganza: It will be a REQUIRED field.  The origins of a volume do not matter to these new APIs.

spec.md
carlbraganza: I think `NOT_FOUND` would suffice.  Any error message can be embellished with additional details.

csi.proto
```diff
@@ -1912,3 +1930,139 @@ message GetVolumeGroupSnapshotResponse {
...
+  // This value must be the same in all such messages returned by
+  // the stream.
+  // This field is REQUIRED.
+  uint64 volume_size_bytes = 2;
+
+  // This is list of data range tuples. The number of entries
+  // in this list is controlled by the max_results value in the
+  // GetAllocatedRequest message.
+  // This field is OPTIONAL.
+  repeated BlockMetadata block_metadata = 3;
```
bswartz, line 2007:
```suggestion
  repeated BlockMetadata block_metadata = 3;
  
  // Indicates there are no more allocated blocks in the list.
  // This field is REQUIRED.
  bool end_of_list = 4;
```

csi.proto
```diff
@@ -1912,3 +1930,139 @@ message GetVolumeGroupSnapshotResponse {
...
+  map<string, string> secrets = 4 [(csi_secret) = true];
+}
+
+// GetAllocatedResponse messages are returned in gRPC stream.
+// Cumulatively, they provide information on the allocated data
+// ranges in the snapshot.
+message GetAllocatedResponse {
+  // This specifies the style used in the BlockMetadata sequence.
+  // This value must be the same in all such messages returned by
+  // the stream.
```
bswartz, line 1993:
```suggestion
  // If block_metadata_type is FIXED_LENGTH, then the size_bytes field
  // of each message in the block_metadata list MUST be constant.
```

csi.proto
```diff
@@ -1912,3 +1930,139 @@ message GetVolumeGroupSnapshotResponse {
...
+}
+
+// GetDeltaResponse messages are returned in gRPC stream.
+// Cumulatively, they provide information on the data ranges that
+// have changed between the base and target snapshots specified
+// in the GetDeltaRequest message.
+message GetDeltaResponse {
+  // This specifies the style used in the BlockMetadata sequence.
+  // This value must be the same in all such messages returned by
+  // the stream.
```
bswartz, line 2052: Same as above

csi.proto
```diff
@@ -1912,3 +1930,139 @@ message GetVolumeGroupSnapshotResponse {
...
+  // This field is REQUIRED.
+  uint64 volume_size_bytes = 2;
+
+  // This is list of data range tuples. The number of entries
+  // in this list is controlled by the max_results value in the
+  // GetDeltaRequest message.
+  // This field is OPTIONAL.
+  repeated BlockMetadata block_metadata = 3;
```
bswartz, line 2066: Same as above

bswartz (changes requested):

csi.proto
carlbraganza: As I understand it, a gRPC stream returns an EOF on proper termination and an error otherwise. Would that not suffice?

csi.proto
```diff
@@ -1912,3 +1930,139 @@ message GetVolumeGroupSnapshotResponse {
...
+
+  // This returns the logical size of the underlying volume.
+  // This value must be the same in all such messages returned by
+  // the stream.
+  // This field is REQUIRED.
+  uint64 volume_size_bytes = 2;
+
+  // This is list of data range tuples. The number of entries
+  // in this list is controlled by the max_results value in the
+  // GetAllocatedRequest message.
```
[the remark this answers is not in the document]
carlbraganza: Applying changes to `spec.md` - `csi.proto` is auto-generate.

csi.proto
carlbraganza: Picking up your suggestion as an addition - the first sentence is still required.

csi.proto
bswartz: My concern here is when max_results is specified, and you get an EOF, how do you know if you need to call it one more time? Or even if max_results is set to zero, can the SP decide to send less than all the results in one stream?

During my review I was going to add like a whole paragraph explaining the relationship between the max_results field and the number of elements in the block_metadata field, and I decided it was too confusing and it would be better to make it explicit.

The core of the problem is that we've introduced BOTH pagination and streaming and this means the client can't rely on the stream ending to indicate that there are no more pages.

carlbraganza: > The core of the problem is that we've introduced BOTH pagination and streaming and this means the client can't rely on the stream ending to indicate that there are no more pages.

Hmm ... I'm not sure I understand your confusion! Perhaps we can use this [snippet of Go client prototype code](https://github.com/PrasadG193/external-snapshot-metadata/blob/e9d364a93b77d5a2d00193f164c58c9a676491e3/cmd/client/main.go#L217) (written by @PrasadG193) to dissect the problem?
```
	c.initGRPCClient(snapMetaSvc.Spec.CACert, snapMetaSvc.Spec.Address, saToken, snapNamespace)
	stream, err := c.client.GetDelta(ctx, &pgrpc.GetDeltaRequest{
		BaseSnapshotId:   snapNamespace + "/" + baseVolumeSnapshot,
		TargetSnapshotId: snapNamespace + "/" + targetVolumeSnapshot,
		StartingOffset:   0,
		MaxResults:       uint32(256),
	})
	if err != nil {
		return err
	}
	done := make(chan bool)
	fmt.Println("Resp received:")
	go func() {
		for {
			resp, err := stream.Recv()
			if err == io.EOF {
				done <- true //means stream is finished
				return
			}
			if err != nil {
				log.Fatalf("cannot receive %v", err)
			}
			respJson, _ := json.Marshal(resp)
			fmt.Println(string(respJson))
		}
	}()

	<-done //we will wait until all response is received
```
The example above doesn't really illustrate processing of the multiple entries in the returned `block_metadata` slice, as the response is simply dumped as a string in JSON format, but it is clear that instead of the `fmt.Println` a real client would process the `block_metadata` slice ("page") in the `resp`. The stream processing loop itself is exited only on a non-nil `err` value from `stream.Recv()`, where `io.EOF` is not really an error but the formal proper end-of-stream indicator.

The corresponding [server (sidecar) side logic for this](https://github.com/PrasadG193/external-snapshot-metadata/blob/e9d364a93b77d5a2d00193f164c58c9a676491e3/cmd/server/main.go#L202) primarily consists of a loop doing this:
```
func (s *Server) GetDelta(req *pgrpc.GetDeltaRequest, cbtClientStream pgrpc.SnapshotMetadata_GetDeltaServer) error {
	... // call the SP service after assembling the parameters
	done := make(chan bool)
	go func() {
		for {
			... // receive resp from the SP service
			log.Print("Received response from csi driver, proxying to client")
			if err := cbtClientStream.Send(resp); err != nil {
				log.Printf(fmt.Sprintf("cannot send %v", err))
				return
			}
		}
	}()
	<-done //we will wait until all response is received
	return nil
}
```
(The prototype sidecar has very little error handling)

carlbraganza: Yes, it appears to be the idiomatic way in Go.

bswartz: I don't hate this way of doing it, but even if it's idiomatic in Go, I'm not sure we use errors in this way anywhere else in CSI. I just want to highlight that a boolean flag in the return message would achieve the same effect as using an error to signal the end of iteration.

spec.md
```diff
@@ -3149,6 +3167,232 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+- The **extent-based** style does not constrain the value of the `size_bytes` field in a sequence.
+
+The SnapshotMetadata service permits either style at the discretion of the
+plugin as long as the style does not change mid-stream in any given RPC.
+
+```protobuf
+enum BlockMetadataType {
+  // The FIXED_LENGTH value indicates that data ranges are
+  // returned in fixed size blocks.
+  FIXED_LENGTH=0;
```
jdef, line 3213: we've strived to have the zero-val for enum's called UNKNOWN. for compat reasons. why does this vary from that approach?

spec.md
```diff
@@ -412,6 +412,16 @@ service GroupController {
     }
 }
 
+service SnapshotMetadata {
+  option (alpha_service) = true;
+
+  rpc GetAllocated(GetAllocatedRequest)
```
jdef, line 418: recommend including a noun in the RPC (and related req, resp messages), like we have elsewhere. e.g. `GetBlocksAllocated`, `GetBlocksDelta`, or something like that

spec.md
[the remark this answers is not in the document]
jdef: > Ceph-CSI would like to implement this functionality. In order to communicate with the Ceph cluster, the requests (this one and `GetDeltaRequest`) need to provide credentials like the `secrets` as in `CreateVolumeRequest`:
> 
> ```
> map<string, string> secrets = 5 [(csi_secret) = true];
> ```

why aren't backend secrets configured at the plugin level? why do they need to be part of the RPC here? is this an exercise in future-proofing?

spec.md
```diff
@@ -412,6 +412,16 @@ service GroupController {
     }
 }
 
+service SnapshotMetadata {
```
jdef, line 415: i'm wondering about "stream" for the responses. we don't stream any other type of response. i notice that each of the response types already has `repeated` block metadata. in practice, how large do we expect these metadata chains to be? it also looks like there's a mechanism to resume from some offset (perhaps in case the stream is interrupted?) - which means that clients already have to keep track of what they've consumed.

why not just send a very large batch of metadata for each request? the golang default gRPC max message size is 4MiB. is that not good enough?

is the use of streaming an example of premature optimization?

spec.md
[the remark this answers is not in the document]
nixpanic: > why aren't backend secrets configured at the plugin level? why do they need to be part of the RPC here? is this an exercise in future-proofing?

Ceph-CSI for example has a single instance for the CSI-drivers, but (in Kubernetes) the StorageClass points to a particular cluster. Different StorageClasses can point to different clusters, each with their own secrets (also linked in the StorageClass). Because of the stateless nature, there is no guarantee that secrets can be (temporarily) stored for plugin before APIs of this new SnapshotMetadata service are called.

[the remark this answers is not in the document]
nixpanic: > @nixpanic In K8s, do the secret data come from either of:
> 
>     * `csi.storage.k8s.io/snapshotter-secret-[name|namespace]` entries of the `VolumeSnapshotClass.Parameters`
> 
>     * `csi.storage.k8s.io/provisioner-secret-[name|namespace]` entries of a `StorageClass.Parameters`

Yes, that is right. The kubernetes-csi/external-snapshotter resolves those secrets and passes the contents in the CSI procedures.

spec.md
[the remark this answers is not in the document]
nixpanic: Ok, consider calling it `capacity_bytes` as well maybe?

spec.md
carlbraganza: This is definitely a new gRPC pattern for the CSI API, but I don't think this is a case of premature optimization.

The amount of changed block data would be proportional to the size of the volume, the amount of change, and the mechanism chosen to represent changed areas (FIXED_LENGH or VARIABLE_LENGTH).  For example, assuming a 1TiB volume (2^40) and 1MiB blocks, there would be 1MiB such blocks; with 8KiB blocks there would 128MiB such blocks. In the worst case for FIXED_LENGTH all the blocks could have changed. In the worst case for VARIABLE_LENGTH every alternative block could have changed.  16 bytes (+ some overhead?) are used for each changed block datum, which should also be factored in.

Beyond the length of a single stream response, assuming a fixed slice length forces the client to support this size, which could break some memory conservative clients. (This is why there is an option for the client to request a `max_results` limit).

Recall too, that an earlier iteration of this KEP was criticized for passing all this metadata through the Kubernetes API server. The direct connection from the backup app to this new CSI service (sidecar) bypasses the K8s API server but cannot reduce the quantity of metadata!

spec.md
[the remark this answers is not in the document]
carlbraganza: Okay, I'll change it to `volume_capacity_bytes` to match the `Volume` message.

Interestingly,  in `Snapshot` you'll find:
>message Snapshot {
  // This is the complete size of the snapshot in bytes. The purpose of
  // this field is to give CO guidance on how much space is needed to
  // create a volume from this snapshot. The size of the volume MUST NOT
  // be less than the size of the source snapshot. This field is
  // OPTIONAL. If this field is not set, it indicates that this size is
  // unknown. The value of this field MUST NOT be negative and a size of
  // zero means it is unspecified.
  int64 size_bytes = 1;

spec.md
```diff
@@ -3149,6 +3167,232 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Snapshot list mismatch | 3 INVALID_ARGUMENT | Besides the general cases, this code SHOULD also be used to indicate when plugin supporting CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT detects a mismatch in the `snapshot_ids`. | If a mismatch is detected in the `snapshot_ids`, caller SHOULD use different `snapshot_ids`. |
 | Volume group snapshot does not exist | 5 NOT_FOUND | Indicates that a volume group snapshot corresponding to the specified `group_snapshot_id` does not exist. | Caller MUST verify that the `group_snapshot_id` is correct and that the volume group snapshot is accessible and has not been deleted before retrying with exponential back off. |
 
+### Snapshot Metadata Service RPCs
+
+The Snapshot Metadata service is an optional service that is used to retrieve metadata on the allocated blocks of a single snapshot,
```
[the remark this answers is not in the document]
carlbraganza: Thanks, I'll fix all such violations.

spec.md
carlbraganza: I was not aware of this. I will add an `UNKNOWN` zero value.

spec.md
carlbraganza: @PrasadG193 @ihcsim ^

spec.md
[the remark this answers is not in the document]
bswartz: @jdef, I would say yes, this is an exercise in future proofing. Every other CSI RPC that we've added without a secrets field has later needed a secrets field to be added. My stance is that the default should be to require the CO to provide pre-RPC secrets and to figure out what those look like, unless we can prove it doesn't make sense to do so.

The underlying problem are the few CSI drivers that use per-volume secrets which the CO is responsible for managing. Those drivers find it impossible to implement new features until a secrets field is added and the CO is updated to provide the appropriate per-volume secrets.

spec.md
ihcsim: `GetBlocksAllocated` and `GetBlocksDelta` sound good to me. In the context of `SnapshotMetadata` service, it's clear that we are talking about blocks metadata, not blocks.

csi.proto
ihcsim: Why not both? Conventionally, client side should keep the connection alive until `io.EOF`, to [avoid leaks](https://pkg.go.dev/google.golang.org/grpc#ClientConn.NewStream). The `end_of_list` is a confirmation from the server to the client that everything has been sent. If there is an `io.EOF` but `end_of_list` is `false`, then the client knows it got an incomplete list, and will need to handle it.

spec.md
PrasadG193: `GetBlocksAllocated` and `GetBlocksDelta` sounds a bit misleading to me as we are not returning actual blocks. Maybe `GetBlocksMetadataAllocated`/`GetBlocksMetadataDelta`?

spec.md
carlbraganza: @jdef We discussed this in the SIG Data Protection WG meeting. Some relevant observations:

- We're dealing with metadata access here, and not a control path. There is no existing precedent for such operations in the current CSI spec.
- The performance of a gRPC stream would better than the comparative use of a classic pagination API.
   - A classic pagination API would involve multiple round-trip calls, versus a single call with the stream API.
   - A classic pagination API requires the SP to fill the next page of metadata to return. Depending on the underlying APIs, this may require the SP to reissue the original allocated/diff call and skip the metadata records already returned in previous calls.  In contrast, an SP server processing a stream API can send multiple responses without reissuing the allocated/diff call.

spec.md
carlbraganza: I'll go with `GetBlockMetadataAllocated` and `GetBlockMetadataDelta`.

csi.proto
carlbraganza: IMO, its a bit paranoic!

ihcsim: I was thinking of a server-side graceful termination scenario, where the sidecar (or gRPC) had a chance to send an `io.EOF` before an unexpected termination, and not all the block metadata has been retrieved from the driver plugin. There is no way for the backup software to know the list it got is incomplete. Is there not a valid scenario?

saad-ali:
@carlbraganza can you make sure you've signed the CLA. See https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#licence-and-cla

carlbraganza:
> @carlbraganza can you make sure you've signed the CLA. See https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#licence-and-cla

Thanks. Looking into getting this signed.

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Snapshot list mismatch | 3 INVALID_ARGUMENT | Besides the general cases, this code SHOULD also be used to indicate when plugin supporting CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT detects a mismatch in the `snapshot_ids`. | If a mismatch is detected in the `snapshot_ids`, caller SHOULD use different `snapshot_ids`. |
 | Volume group snapshot does not exist | 5 NOT_FOUND | Indicates that a volume group snapshot corresponding to the specified `group_snapshot_id` does not exist. | Caller MUST verify that the `group_snapshot_id` is correct and that the volume group snapshot is accessible and has not been deleted before retrying with exponential back off. |
 
+### Snapshot Metadata Service RPCs
+
```
jdef, line 3171: `**ALPHA FEATURE**` notice (as appears elsewhere) would probably be nice to have here

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+The Snapshot Metadata service permits either style at the discretion of the plugin as long as the style does not change mid-stream in any given RPC.
+
+```protobuf
+enum BlockMetadataType {
+  UNKNOWN = 0;
+
+  // The FIXED_LENGTH value indicates that data ranges are
+  // returned in fixed size blocks.
+  FIXED_LENGTH=1;
```
jdef, line 3210: nit: consistent spacing around `=` sign, here and below

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+  // The VARIABLE_LENGTH value indicates that data ranges
+  // are returned in potentially variable sized extents.
+  VARIABLE_LENGTH=2;
+}
+
+```
+
```
jdef, line 3218: nit: only 1 newline please

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+#### GetMetadataAllocated RPC
```
jdef, line 3220: nit: other RPC section headers don't end w/ ` RPC`. i haven't scanned the entire doc, but whatever we do here should be consistent w/ the rest (or, at least, the majority of the rest)

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+}
+
+```
+
+The `BlockMetadataType` enumerated type describes how the `size_bytes` fields of `BlockMetadata` tuples may vary in a sequence. There are two prevalent styles in which data ranges are described:
+
+- The **FIXED_LENGTH** style requires a fixed value for the `size_bytes` field in all tuples in a given sequence.
+- The **VARIABLE_LENGTH** style does not constrain the value of the `size_bytes` field in a sequence.
+
+The Snapshot Metadata service permits either style at the discretion of the plugin as long as the style does not change mid-stream in any given RPC.
```
jdef, line 3202: this seems to imply that subsequent invocations of the same RPC may specify different length styles - even for the same snapshot. is this intended? i hope so, because otherwise the plugin would need to maintain state to enforce consistent styles across multiple RPCs - and we probably don't want to force plugins to do that

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Snapshot list mismatch | 3 INVALID_ARGUMENT | Besides the general cases, this code SHOULD also be used to indicate when plugin supporting CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT detects a mismatch in the `snapshot_ids`. | If a mismatch is detected in the `snapshot_ids`, caller SHOULD use different `snapshot_ids`. |
 | Volume group snapshot does not exist | 5 NOT_FOUND | Indicates that a volume group snapshot corresponding to the specified `group_snapshot_id` does not exist. | Caller MUST verify that the `group_snapshot_id` is correct and that the volume group snapshot is accessible and has not been deleted before retrying with exponential back off. |
 
+### Snapshot Metadata Service RPCs
+
+The Snapshot Metadata service is an optional service that is used to retrieve metadata on the allocated blocks of a single snapshot, or the changed blocks between a pair of snapshots of the same block volume.
+Retrieval of the data blocks of a snapshot is not addressed by this service and it is assumed that existing mechanisms to read snapshot data will suffice.
+
```
jdef, line 3174: back when the spec was first written, I believe there was language around concurrency of RPCs issued by CO to a plugin - some recommendation meant to ease the burden on plugins. has thought been given to the concurrency of these new RPCs? i.e. is the intent to align w/ the current spirit of the spec w/ respect to concurrency, or is a different concurrency model needed/wanted?

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // This field is REQUIRED.
+  uint64 volume_capacity_bytes = 2;
+
+  // This is a list of data range tuples.
+  // If the value of max_results in the GetMetadataAllocatedRequest
+  // message is non-zero, then the number of entries in this list
+  // MUST be less than or equal to that value.
+  // The byte_offset field of each message MUST be strictly increasing.
+  // The SP MUST NOT return a BlockMetadata message with a byte_offset
+  // less than than the value of starting_offset in the
```
jdef, line 3280: this is inconsistent w/ the downward rounding option granted to plugins as previously describe - please reconcile

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // This field is OPTIONAL.
+  repeated BlockMetadata block_metadata = 3;
+}
+
+```
+
+##### GetMetadataAllocated Errors
+If the plugin is unable to complete the `GetMetadataAllocated` call successfully it must return a non-OK gRPC code in the gRPC status.
+
+The following conditions are well defined:
```
jdef, line 3295: if the snapshot is in the middle of a deletion (which could be a long running op, maybe), what error might this return? if NOT_FOUND then please clarify here.

are there any other snapshot states to consider?

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Missing or otherwise invalid argument | 3 INVALID_ARGUMENT | Indicates that a required argument field was not specified or an argument value is invalid | The caller should correct the error and resubmit the call. |
+| Invalid `snapshot_id` | 5 NOT_FOUND | Indicates that the snapshot specified was not found. | The caller should re-check that this object exists. |
+| Invalid `starting_offset` | 11 OUT_OF_RANGE | The starting offset exceeds the volume size. | The caller should specify a `starting_offset` less than the volume's size. |
+
+#### GetMetadataDelta RPC
```
jdef, line 3303: ditto re: ` RPC` section suffix

spec.md
jdef: the last bullet point is probably the most convincing here

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+Tuples in a sequence are in ascending order of `byte_offset`, with no overlap between adjacent tuples.
+
+```protobuf
+
+// BlockMetadata specifies a data range.
+message BlockMetadata {
+  // This is the 0 based byte position in the volume or snapshot,
+  // measured from the start of the object.
+  // This field is REQUIRED.
+  uint64 byte_offset = 1;
```
jdef, line 3187: volume capacity is represented as int64 elsewhere, offsets/size/capacity fields throughout this PR should align

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // This field is REQUIRED.
+  string target_snapshot_id = 2;
+
+  // This indicates the zero based starting byte position in the volume
+  // snapshot from which the result should be computed. It is intended
+  // to be used to continue a previously interrupted call.
+  // The plugins may round down this offset to the nearest alignment
+  // boundary based on the BlockMetadataType used.
+  // This field is REQUIRED.
+  uint64 starting_offset = 3;
```
jdef, line 3327: int64

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+  // This is an optional parameter, and if non-zero it specifies the
+  // maximum number of tuples to be returned in each
+  // GetMetadataDeltaResponse message returned by the RPC stream.
+  // The plugin will determine an appropriate value if 0, and is
+  // always free to send less than the requested value.
+  // This field is OPTIONAL.
+  uint32 max_results = 4;
```
jdef, line 3335: int32

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+  // Secrets required by plugin to complete the request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 5 [(csi_secret) = true];
+}
+
+// GetMetadataDeltaResponse messages are returned in gRPC stream.
```
jdef, line 3343: same grammar nit

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // If block_metadata_type is FIXED_LENGTH, then the size_bytes field
+  // of each message in the block_metadata list MUST be constant.
+  // This field is REQUIRED.
+  BlockMetadataType block_metadata_type = 1;
+
+  // This returns the capacity of the underlying volume in bytes.
+  // This value must be the same in all such messages returned by
+  // the stream.
+  // This field is REQUIRED.
+  uint64 volume_capacity_bytes = 2;
```
jdef, line 3360: int64

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+  // This is a list of data range tuples.
+  // If the value of max_results in the GetMetadataDeltaRequest message
+  // is non-zero, then the number of entries in this list MUST be less
```
jdef, line 3364: greater than zero

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // than or equal to that value.
+  // The byte_offset field of each message MUST be strictly increasing.
+  // The SP MUST NOT return a BlockMetadata message with a byte_offset
+  // less than than the value of starting_offset in the
```
jdef, line 3368: same conflict w/ earlier, optional rounding-down by plugin

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // This field is OPTIONAL.
+  repeated BlockMetadata block_metadata = 3;
+}
+
+```
+
+##### GetMetadataDelta Errors
+If the plugin is unable to complete the `GetMetadataDelta` call successfully it must return a non-OK gRPC code in the gRPC status.
+
+The following conditions are well defined:
```
jdef, line 3383: similar concerns w/ snapshots in not-ready states: does NOT_FOUND work or some other code needed?

jdef (changes requested):
thanks for being patient. completed another round of review

spec.md
jdef: for example, if the CO requests a starting offset of 600 and the plugin realigns/rounds to boundary starting at offset 512, then the first block metadata response could be reported w/ an initial offset of 512 - which is not strictly increasing from the CO-specified starting offset.

or am I missing something?

spec.md
carlbraganza: Yes. The spec leaves it to the SP to chose the style on a per RPC basis; consistency is required only within each RPC.

spec.md
carlbraganza: Looks like other RPC sections have the name in the form: `RPC Name` (within backquotes). The errors subsection has no special formatting. Changing...

spec.md
carlbraganza: I had not looked at that previously.

Looking at the `Concurrency` and `Error Scheme` sections it appears that returning an `ABORTED` error code would suffice if the SP decides to constrain operations. The existing recommendation that the caller retry with an exponential back off would also apply.

spec.md
```diff
@@ -3149,6 +3167,227 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // This field is REQUIRED.
+  uint64 starting_offset = 2;
+
+  // This is an optional parameter, and if non-zero it specifies the
+  // maximum number of tuples to be returned in each
+  // GetMetadataAllocatedResponse message returned by the RPC stream.
+  // The plugin will determine an appropriate value if 0, and is
+  // always free to send less than the requested value.
+  // This field is OPTIONAL.
+  uint32 max_results = 3;
```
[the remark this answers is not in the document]
carlbraganza: Okay. I'll use the `int32` data type used in `ListVolumesRequest` and `ListSnapshotsRequest`.  The odd one out is `NodeGetInfoResponse` which uses `int64`.

carlbraganza:
@jdef This comment was in an outdated part of the document so I'm not sure if it was addressed:

> why aren't backend secrets configured at the plugin level? why do they need to be part of the RPC here? is this an exercise in future-proofing?

This was brought to our attention by @nixpanic, because the Ceph CSI plugin supports multiple sets of secrets (per Pool, I believe).

spec.md
carlbraganza: Got it - no unsigned integers!

spec.md
carlbraganza: No, I think you're correct here - this is not quite right.  How does this sound:

> The data range of the first BlockMetadata message in 
> the sequence MUST include the starting_offset requested
> in the GetMetadataAllocatedRequest message.

spec.md
carlbraganza: Changed to:

>  // The data range of the first BlockMetadata message in the list MUST
>  // include the starting_offset requested in the GetMetadataDeltaRequest
>  // message.

spec.md
carlbraganza: The NOT_FOUND error seems the most appropriate from the set of [gRPC status codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html).

I think it would be fine even if this status code were returned mid-stream when the underlying implementation of this API catches up with the asynchronous deletion happening elsewhere in the storage subsystem.  If it doesn't catch up then all is good for this RPC call, though the client would probably encounter errors later if it were to attempt to access the snapshot concerned.  Either way, there is not much that the CO can do if the snapshot object is deleted from the storage subsystem.

Regarding other snapshot states not involving deletion, the gRPC status codes offer subtle distinctions between FAILED_PRECONDITION, UNAVAILABLE and ABORTED.  I'd expect the SP to be able to map into these states, but I don't see the need to call them out explicitly here.

spec.md
carlbraganza: Please see my previous response.

carlbraganza:
> thanks for being patient. completed another round of review

Thanks @jdef. I tried to address all the items you surfaced. PTAL.

csi.proto
```diff
@@ -1912,3 +1930,164 @@ message GetVolumeGroupSnapshotResponse {
...
+  // This is the identifier of a second snapshot in the same volume,
+  // created after the base snapshot.
+  // This field is REQUIRED.
+  string target_snapshot_id = 2;
+
+  // This indicates the zero based starting byte position in the volume
+  // snapshot from which the result should be computed. It is intended
+  // to be used to continue a previously interrupted call.
+  // The plugins may round down this offset to the nearest alignment
+  // boundary based on the BlockMetadataType used.
```
bswartz, line 2041: Sorry to be pedantic about this but the explanation isn't clear enough that someone could write a test to verify this behavior. Is the SP only allowed to round down if the type is FIXED_LENGTH? And is it only allowed to round by the exactly the size of one block? Would an SP that rounds down by more than one block (because internally it operates on groups of blocks) be in violation of the spec? If an SP uses VARIABLE_LENGTH blocks, how much can it round down by?

My worry is that if we leave it up to SPs and COs to figure out who should discard the "repeated" metadata blocks that might result from an interrupted transmission, they might make incompatible assumptions such that in cases of actual interruption, one or more changed blocks would get missed, but only on certain SP+CO combinations, and the bug is unlikely to get caught because it's a rare thing for these connections to get interrupted unless you specifically test for it. This bug would result in very subtle data corruption, which wouldn't even have a chance of breaking anything until a corrupted backup is restored, and even then the corruption might not get noticed, but it would be there.

I want to be crystal clear about what the SP is allowed to do with this field and what assumptions the CO is allowed to make, so we can write automated tests that ensure there is no chance of a misunderstanding by a SP or a CO implementer resulting in data corruption. I realize that the earlier wording of this section (which I had recommended) no longer made sense after the additional changes we agreed to, but we still need more clarity here.

bswartz (changes requested):

csi.proto
```diff
@@ -1912,3 +1930,164 @@ message GetVolumeGroupSnapshotResponse {
...
+  int64 volume_capacity_bytes = 2;
+
+  // This is a list of data range tuples.
+  // If the value of max_results in the GetMetadataAllocatedRequest
+  // message is greater than zero, then the number of entries in this
+  // list MUST be less than or equal to that value.
+  // The byte_offset field of each message MUST be strictly increasing.
+  // The data range of the first BlockMetadata message in the list MUST
+  // include the starting_offset requested in the
+  // GetMetadataAllocatedRequest message.
```
bswartz, line 2015: Oh wait, this is the language I was looking for in my earlier comment. I think this may be clear enough because it's precise. Do we need to add any clarification in the case that this message is streamed and there are multiple responses for a single request? In that case the restriction mentioned only applies to the first response. The second and later responses must follow after the last block of the previous response (which is hopefully obvious).

carlbraganza: Okay. I'll clarify that it is the first tuple in the stream that must have this property.

carlbraganza:
After responding to @bswartz's previous comment, it occurred to me that requiring the `starting_offset` to be present in the first tuple range is not correct, because there may not be allocated/changed blocks in that range. What we want is that the SP to **_consider this starting_offset in its computation_** when generating the data stream.  Accordingly, I changed the text to the following:

>  // The computation of the data range in the first BlockMetadata
>  // message in the stream of GetMetadataAllocatedResponse messages
>  // returned MUST consider the starting_offset requested in the
>  // GetMetadataAllocatedRequest message, though it is not required
>  // that the first data range tuple returned actually include the
>  // starting_offset, as the first allocated block range may start
>  // after this offset.

The language is likewise modified in the description of the GetMetadataDeltaResponse message.

I don't know if this negatively impact's @bswartz's request for precise language that can be tested, but it is more accurate.

csi.proto
carlbraganza: Oops! @bswartz please see the follow up comment below: https://github.com/container-storage-interface/spec/pull/551#issuecomment-1828864769

spec.md
carlbraganza: @jdef Please see this comment below: https://github.com/container-storage-interface/spec/pull/551#issuecomment-1828864769

csi.proto
bswartz: Thanks for pointing out the case about the block containing starting_offset not containing any changes -- I had missed this. I'm convinced we can be more precise -- I'm going to take a swing a recommending some new language which is impossible to misunderstand.

csi.proto
```diff
@@ -1912,3 +1930,172 @@ message GetVolumeGroupSnapshotResponse {
...
+  // This is the identifier of a second snapshot in the same volume,
+  // created after the base snapshot.
+  // This field is REQUIRED.
+  string target_snapshot_id = 2;
+
+  // This indicates the zero based starting byte position in the volume
+  // snapshot from which the result should be computed. It is intended
+  // to be used to continue a previously interrupted call.
+  // The plugins may round down this offset to the nearest alignment
+  // boundary based on the BlockMetadataType used.
```
bswartz, line 2045: // This indicates the zero based starting byte position in the volume
  // snapshot from which the result should be computed. It is intended
  // to be used to continue a previously interrupted call.
  // The CO SHOULD specify this value to the be the byte immediately
  // after the last byte of the last block received, if continuing an
  // interrupted operation, or zero if not.
  // The SP MUST NOT return any blocks for which the block byte_offset
  // plus the block size_bytes is less than starting_offset.
  // THE SP MAY return blocks with byte_offset less than starting_offset if
  // starting_offset is less than byte_offset plus size_bytes for that block.

csi.proto
```diff
@@ -1912,3 +1930,172 @@ message GetVolumeGroupSnapshotResponse {
...
+  // is greater than zero, then the number of entries in this list MUST
+  // be less than or equal to that value.
+  // The byte_offset field of each message MUST be strictly increasing.
+  // The computation of the data range in the first BlockMetadata
+  // message in the stream of GetMetadataDeltaResponse messages
+  // returned MUST consider the starting_offset requested in the
+  // GetMetadataDeltaRequest message, though it is not required
+  // that the first data range tuple returned actually include the
+  // starting_offset, as the first changed block range may start
+  // after this offset.
```
bswartz, line 2093: // The computation of the data range in the first BlockMetadata
  // message in the stream of GetMetadataDeltaResponse messages
  // returned MUST consider the starting_offset requested in the
  // GetMetadataDeltaRequest message.
  // The SP MUST NOT return blocks for which are before starting offset
  // (blocks for which byte_offset plus size_bytes is less than or equal to
  // starting_offset).
  // Changed blocks which overlap with starting_offset or come after
  // starting_offset (blocks for which byte_offset plus size_bytes is greater
  // than starting_offset) MUST be returned by the SP.

bswartz (changes requested):

bswartz:
I'm thinking somewhere in the spec.md file (not the proto file) in the section about change-block tracking we need to mention in plain English how the SP is responsible for ensuring EVERY changed block gets communicated to the CO before reporting the end of stream, and that the CO is responsible for picking the correct `starting_offset` when restarting an interrupted connection, in order to ensure that it doesn't miss any blocks. The stakes are high for not screwing up this metadata transmission, so it's worth recommending caution to implementers not to be careless.

csi.proto
carlbraganza: Cool! I'll generalize this to a "data range" (instead of "block").

carlbraganza:
> I'm thinking somewhere in the spec.md file (not the proto file) in the section about change-block tracking we need to mention in plain English how the SP is responsible for ensuring EVERY changed block gets communicated to the CO before reporting the end of stream, and that the CO is responsible for picking the correct `starting_offset` when restarting an interrupted connection, in order to ensure that it doesn't miss any blocks. The stakes are high for not screwing up this metadata transmission, so it's worth recommending caution to implementers not to be careless.

I think this is a great idea. I'm adding the following section between the description of the metadata format and the RPC definitions:

```
### Special considerations for processing a metadata gRPC stream

The remote procedure calls of this service return snapshot metadata within a gRPC stream of response messages.
There are some important things for the SP and CO to consider here:

- In normal operation an SP MUST terminate the stream only after ALL metadata is transmitted,
  using the language specific idiom for terminating a gRPC stream source.
  This results in the CO receiving its language specific notification idiom for the end of a gRPC stream,
  which provides a **definitive indication that all available metadata has been received**.

- If the SP encounters an error while recovering the metadata it MUST abort transmission of the
  stream with its language specific error idiom.
  This results in the CO receiving the error in its language specific idiom, which
  will be different from the language specific idiom for the end of a gRPC stream.

- It is possible that the gRPC stream gets interrupted for arbitrary reasons beyond
  the control of either the SP or the CO.
  The SP will get an error when writing to the stream and must abort its transmission.
  The CO will receive an error in its language specific idiom, which
  will be different from the language specific idiom for the end of a gRPC stream.

- In all circumstances where the CO receives an error when reading from the gRPC stream
  it may attempt to **continue** the operation by re-sending its request message but with a
  starting_offset adjusted to a byte position beyond that of any metadata already received.
  The SP must always examine the starting_offset and ensure that metadata sent in
  the gRPC stream logically includes the starting_offset but not any byte position prior
  to this offset.
```

spec.md
```diff
@@ -3237,14 +3237,14 @@ There are some important things for the SP and CO to consider here:
...
-  The SP will get an error when writing to the stream and must abort its transmission.
+  The SP will get an error when writing to the stream and MUST abort its transmission.
   The CO will receive an error in its language specific idiom, which
   will be different from the language specific idiom for the end of a gRPC stream.
 
 - In all circumstances where the CO receives an error when reading from the gRPC stream
-  it may attempt to **continue** the operation by re-sending its request message but with a
-  starting_offset adjusted to a byte position beyond that of any metadata already received.
-  The SP must always examine the starting_offset and ensure that metadata sent in
+  it MAY attempt to **continue** the operation by re-sending its request message but with a
```
jdef, line 3245: nit: one sentence per line please (there are several sentences in this section that violate this rule)

spec.md
```diff
@@ -3237,14 +3237,14 @@ There are some important things for the SP and CO to consider here:
...
+  starting_offset adjusted to the NEXT byte position beyond that of any metadata already received.
+  The SP MUST always examine the starting_offset and ensure that metadata sent in
```
jdef, line 3247: nit: can we write `starting_offset` instead (wrapped in ticks) to clearly call out the field name? I realize that there are probably other places in the spec that don't do this right (prior to this PR) - and we should clean those up too, at some later point.

spec.md
```diff
@@ -3237,14 +3237,14 @@ There are some important things for the SP and CO to consider here:
...
   the gRPC stream logically includes the starting_offset but not any byte position prior
```
jdef, line 3248: "but not any byte position prior" seems to contradict the gRPC method docs, which seem to indicate that `starting_offset` can live in the window defined by (offset, offset+size). or am I reading this wrong?

jdef (commented):
playing catch-up here, trying to review individual commits. additional word-smithing may be needed.

spec.md
carlbraganza: Fixing!

spec.md
carlbraganza: Yes, you're correct! I think I got stuck in the previous thread on continuation.  Changing to:

> ... logically includes the `starting_offset` in the first metadata tuple returned.

bswartz (approved):
/lgtm

spec.md
```diff
@@ -3149,6 +3167,274 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+#### Metadata Format
+
+Block volume data ranges are specified by a sequence of `BlockMetadata` tuples.
+Tuples in a sequence are in ascending order of `byte_offset`, with no overlap between adjacent tuples.
+
+```protobuf
+
+// BlockMetadata specifies a data range.
+message BlockMetadata {
+  // This is the 0 based byte position in the volume or snapshot,
```
jdef, line 3186: nit: consistency.. "0 based" or "zero based", let's pick one and stick w/ it throughout. FWIW I think "zero based" reads better, but I don't think we've standardized on a particular editorial style.

spec.md
```diff
@@ -3149,6 +3167,274 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // snapshot from which the result should be computed.
+  // It is intended to be used to continue a previously interrupted
+  // call.
+  // The CO SHOULD specify this value to be the offset of the byte
+  // position immediately after the last byte of the last data range
+  // received, if continuing an interrupted operation, or zero if not.
+  // The SP MUST NOT return any data ranges for which the byte_offset
+  // plus the size_bytes is less than starting_offset.
+  // The SP MAY return data ranges with byte_offset less than
+  // starting_offset if starting_offset is less than byte_offset plus
```
jdef, line 3265: "... byte_offset less than starting_offset if starting_offset is less than byte_offset ..."

sorry, i'm still confused by this phrasing. what am i missing?

spec.md
```diff
@@ -3149,6 +3167,274 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
```
jdef, line 3365: ditto

spec.md
bswartz: This is language I recommended, and what I was trying to achieve is to constrain the SP's behavior in case of restarting an abnormally terminated stream. Basically I want to be very precise about what starting_offset means, so that SPs don't take liberties and start somewhere other than where the CO expects. Unfortunately, the only way to do that is to describe what mathematical relationships must hold between the values in the request and response.

csi.proto
```diff
@@ -1912,3 +1930,190 @@ message GetVolumeGroupSnapshotResponse {
   // This field is REQUIRED
   VolumeGroupSnapshot group_snapshot = 1;
 }
+
+// BlockMetadata specifies a data range.
+message BlockMetadata {
+  // This is the 0 based byte position in the volume or snapshot,
```
PrasadG193, line 1936:
```suggestion
  // This is the zero based byte position in the volume or snapshot,
```

spec.md
PrasadG193, line 3186:
```suggestion
  // This is the zero based byte position in the volume or snapshot,
```

PrasadG193 (commented):
Change occurences of "0 based" -> "zero based" to maintain consistency as suggested by jdef

spec.md
jdef: Given a requested starting_offset S, and the first block_metadata B0 of the first SP response in a stream, the SP MUST ensure that B0.byte_offset <= S < B0.byte_offset + B0.size_bytes.

.. is this what you're getting at? if so, can we represent the mathematical relationship using language like this instead of the two sentences that attempt to model this in English? I see where I went wrong reading it the first time, and I probably would have not had the same trouble w/ the same in equation form.

bswartz: Yeah I'm perfectly fine using language like that. I had worried that introducing new variables like S and B0 might be extra confusing, but this seems like a case where it will be a little confusing either way, and I do like the clarity of mathematical symbols to express what is really a mathematical relationship.

@carlbraganza do you think you can reword using the language that @jdef recommends?

spec.md
```diff
@@ -3149,6 +3167,276 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // call.
+  // The CO SHOULD specify this value to be the offset of the byte
+  // position immediately after the last byte of the last data range
+  // received, if continuing an interrupted operation, or zero if not.
+  // The SP MUST ensure that the returned response stream does not
+  // contain BlockMetadata tuples that start before but not overlap
+  // the requested starting_offset: i.e. if S is the requested
+  // starting_offset, and B0 is block_metadata[0] of the first message
+  // in the response stream, then the following must always be true:
+  // B0.byte_offset > S || S < B0.byte_offset + B0.size_bytes.
```
jdef, line 3267:
```suggestion
  // S < B0.byte_offset or-else S < B0.byte_offset + B0.size_bytes.
```
^ this seems more readable? (for me, anyway); otherwise 👍

carlbraganza: No problems with switching the order of the first test, but "or-else" ???

How about if I change it to
> ...
> in the response stream, then either (S < B0.byte_offset) must
> be true or (S < B0.byte_offset + B0.size_bytes) must be true.

csi.proto
```diff
@@ -1912,3 +1930,192 @@ message GetVolumeGroupSnapshotResponse {
...
+  // GetMetadataAllocatedRequest message.
+  // The SP MUST NOT return data ranges for which byte_offset plus
+  // size_bytes is less than or equal to starting_offset.
+  // Allocated data ranges which overlap the starting_offset
+  // or for which byte_offset plus size_bytes is greater than
+  // starting_offset MUST be returned by the SP.
+  // BlockMetadata messages MUST NOT overlap for a given snapshot, i.e.
+  // for any two BlockMetadata messages, A and B, if
+  // A.byte_offset < B.byte_offset then
+  // A.byte_offset + A.size_bytes <= B.byte_offset MUST be true.
```
bswartz, line 2033: Another simplification of the English since we can lean on mathematical expressions for precision. Also I think that we don't need to repeat the language about starting_offset here. A simple reference to the other section should be enough.

```suggestion
  // This is a list of data range tuples.
  // If the value of max_results in the GetMetadataAllocatedRequest
  // message is greater than zero, then the number of entries in this
  // list MUST be less than or equal to that value.
  // The SP MUST respect the value of starting_offset in the request.
  // The byte_offset field of each BlockMetadata message MUST be
  // strictly increasing and BlockMetadata messages MUST NOT overlap:
  // i.e. for any two BlockMetadata messages, A and B, if A is returned
  // before B, then A.byte_offset + A.size_bytes <= B.byte_offset MUST
  // be true.
```

csi.proto
```diff
@@ -1912,3 +1930,192 @@ message GetVolumeGroupSnapshotResponse {
...
+  // call.
+  // The CO SHOULD specify this value to be the offset of the byte
+  // position immediately after the last byte of the last data range
+  // received, if continuing an interrupted operation, or zero if not.
+  // The SP MUST ensure that the returned response stream does not
+  // contain BlockMetadata tuples that start before but not overlap
+  // the requested starting_offset: i.e. if S is the requested
+  // starting_offset, and B0 is block_metadata[0] of the first message
+  // in the response stream, then either (S < B0.byte_offset) must be
+  // true or else (S < B0.byte_offset + B0.size_bytes) must be true.
```
bswartz, line 2063:
```suggestion
  // The SP MUST ensure that the returned response stream does not
  // contain BlockMetadata tuples that end before the requested
  // starting_offset: i.e. if S is the requested starting_offset, and B0 is
  // block_metadata[0] of the first message in the response stream,
  // then S < B0.byte_offset + B0.size_bytes must be true.
```

csi.proto
```diff
@@ -1912,3 +1930,192 @@ message GetVolumeGroupSnapshotResponse {
...
+  // GetMetadataDeltaRequest message.
+  // The SP MUST NOT return data ranges for which byte_offset plus
+  // size_bytes is less than or equal to starting_offset.
+  // Changed data ranges which overlap the starting_offset
+  // or for which byte_offset plus size_bytes is greater than
+  // starting_offset MUST be returned by the SP.
+  // BlockMetadata messages MUST NOT overlap for a given snapshot, i.e.
+  // for any two BlockMetadata messages, A and B, if
+  // A.byte_offset < B.byte_offset then
+  // A.byte_offset + A.size_bytes <= B.byte_offset MUST be true.
```
bswartz, line 2117:
```suggestion
  // This is a list of data range tuples.
  // If the value of max_results in the GetMetadataAllocatedRequest
  // message is greater than zero, then the number of entries in this
  // list MUST be less than or equal to that value.
  // The SP MUST respect the value of starting_offset in the request.
  // The byte_offset field of each BlockMetadata message MUST be
  // strictly increasing and BlockMetadata messages MUST NOT overlap:
  // i.e. for any two BlockMetadata messages, A and B, if A is returned
  // before B, then A.byte_offset + A.size_bytes <= B.byte_offset MUST
  // be true.
```

bswartz (changes requested):
I found some more typos, and rather than just recommend the minimal fix, I am recommending a significant simplification of the language, because using mathematical expressions allows us to be more precise with less verbiage.

csi.proto
```diff
@@ -1912,3 +1930,192 @@ message GetVolumeGroupSnapshotResponse {
...
+  // call.
+  // The CO SHOULD specify this value to be the offset of the byte
+  // position immediately after the last byte of the last data range
+  // received, if continuing an interrupted operation, or zero if not.
+  // The SP MUST ensure that the returned response stream does not
+  // contain BlockMetadata tuples that start before but not overlap
+  // the requested starting_offset: i.e. if S is the requested
+  // starting_offset, and B0 is block_metadata[0] of the first message
+  // in the response stream, then either (S < B0.byte_offset) must be
+  // true or else (S < B0.byte_offset + B0.size_bytes) must be true.
```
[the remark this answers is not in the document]
carlbraganza: Thanks Ben! This is simpler.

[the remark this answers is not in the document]
carlbraganza: Thanks, Ben. That is simpler.  Will now read as:

>  // The SP MUST ensure that the returned response stream does not
  // contain BlockMetadata tuples that end before the requested
  // starting_offset: i.e. if S is the requested starting_offset, and
  // B0 is block_metadata[0] of the first message in the response
  // stream, then (S < B0.byte_offset + B0.size_bytes) must be true.

csi.proto
carlbraganza: Changed to read:
>  // This is a list of data range tuples.
  // If the value of max_results in the GetMetadataAllocatedRequest
  // message is greater than zero, then the number of entries in this
  // list MUST be less than or equal to that value.
  // The SP MUST respect the value of starting_offset in the request.
  // The byte_offset fields of adjacent BlockMetadata messages
  // MUST be strictly increasing and messages MUST NOT overlap:
  // i.e. for any two BlockMetadata messages, A and B, if A is returned
  // before B, then (A.byte_offset + A.size_bytes <= B.byte_offset)
  // MUST be true.
  // This MUST also be true if A and B are from block_metadata lists in
  // different response messages in the gRPC stream.
  // This field is OPTIONAL.

I added an additional sentence to extend the notion of "adjacent" to include the entire gRPC stream.

csi.proto
carlbraganza: Handled like the previous case.

csi.proto
carlbraganza: Handled like the previous case.

bswartz (approved):
/lgtm

jdef:
skimmed through the changes yesterday, LGTM

only remaining nit is the needless newline whitespaces that pad the new protobuf blocks

xing-yang:
@carlbraganza Can you squash the commits?

bswartz:
@carlbraganza I can't find evidence of a signed CLA. Did that end up happening?

bswartz:
BTW I can squash the commits too, but we should add some more detail to the squashed commit message.

carlbraganza:
Rebased on master and squashed comments.

bswartz:
As soon as we get a signed CLA from the author(s) we can merge this.

saad-ali:
I can confirm we've received CLA from Veeam including Carl Braganza

carlbraganza:
I looked at the build test failure - my guess is that I used the wrong version of protoc.

bswartz:
You probably need to rebase on HEAD and rebuild. We merged a patch that changed the Makefile significantly and also updated the protoc compiler version

PR #552, opened 2023-10-01
update protoc and related

### LINKED ISSUE #471: Build is broken with go 1.16
We don't have to fix this right away but it looks like builds might be broken in 1.16 because of go module changes.

(cd /home/hekumar/goal/src/github.com/golang/protobuf/ && go get -v -d $(go list -f '{{ .ImportPath }}' ./...)) && \
go build -o "protoc-gen-go" github.com/golang/protobuf/protoc-gen-go
no required module provides package github.com/golang/protobuf/protoc-gen-go: working directory is not part of a module
make[1]: *** [Makefile:66: protoc-gen-go] Error 1
make[1]: Leaving directory '/home/hekumar/goal/src/github.com/container-storage-interface/spec/lib/go'
make: *** [Makefile:36: build] Error 2

huww98:
**What type of PR is this?**

/kind cleanup

updated protoc to v25.2
updated protoc-gen-go to v1.32.0
added protoc-gen-go-grpc at v1.3.0

Modified the Makefile to use "go install" to install tools.
Use extended regular expressions in "sed" command.

Fixes build on arm MacOS host
Maybe Fixing #471, not tested

```release-note
none
```


huww98:
cc @jdef

huww98:
@saad-ali Fixed GitHub action. It seems we are not running `protoc` there, also fixed this.
https://github.com/container-storage-interface/spec/blob/80d53107c70981b9da8aaf9cd1c90249562b22f0/Makefile#L46-L48
This rule is not running because the timestamp of `csi.proto` is not updated.

It seems to me that we should remove all the `diff` in `Makefile`, and rely on `git` to check for updated file.

saad-ali:
Can you make sure you've signed the CLA. See https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#licence-and-cla

huww98:
@saad-ali I just sent the CCLA. Please check.

saad-ali:
Confirmed, CCLA received. Thank you @huww98!

xing-yang:
cc @bswartz

jdef:
are we sure there are no breaking API changes here? e.g. the APIs exposed from the generated bindings seem to have changed considerably. i realize that there were no changes to the CSI API specification, but I wonder about potential breakage for folks relying on anything generated by the old compilers that may have changed. not that we want to block this PR because of that - but the potential for downstream breakage should probably be noted somewhere.

also, has anyone attempted to run these newly generated bindings against some sort of CO-maintained test suite? "CSI sanity" (or similar) comes to mind. or, at least, tried to recompile CO/CSI integration layers in conjunction w/ the changes herein?

huww98:
The only breaking I'm aware of is [this](https://github.com/grpc/grpc-go/releases/tag/cmd%2Fprotoc-gen-go-grpc%2Fv1.0.0):
> The Unimplemented<Service>Server struct is now required to be embedded in implementations of the generated <Service>Server interface.

I will try this change on existing CSI driver and Kubernetes, and report back soon.

huww98:
@jdef See [here](https://github.com/huww98/csi-driver-host-path/compare/master...huww98:csi-driver-host-path:csi-next) for how I get csi-driver-host-path to work with this change. I tested this with csi-sanity. Both from [master branch](https://github.com/kubernetes-csi/csi-test/tree/1153a7c6faa8a266d373127cca6a15c8456c8efe) and recompiled with this change (I made some changes to the mock driver used in testing to compile it). The result is:

```
Ran 73 of 84 Specs in 0.630 seconds
SUCCESS! -- 73 Passed | 0 Failed | 1 Pending | 10 Skipped
```

go.mod
```diff
@@ -5,12 +5,12 @@ go 1.18
 require (
 	github.com/golang/protobuf v1.5.3
 	google.golang.org/grpc v1.57.0
+	google.golang.org/protobuf v1.31.0
```
saad-ali, line 8: Per @xing-yang do we need both versions of protobuf here?

lib/go/Makefile
```diff
@@ -21,7 +21,7 @@ export GOPATH
 
 # Only set PROTOC_VER if it has an empty value.
 ifeq (,$(strip $(PROTOC_VER)))
-PROTOC_VER := 3.9.1
+PROTOC_VER := 24.3
```
saad-ali, line 24: Per @bswartz where does this version number come from?

go.mod
xing-yang, line 8: module github.com/golang/protobuf is deprecated.  Since you added "google.golang.org/protobuf", can you remove "github.com/golang/protobuf"?

saad-ali (commented):
It looks like there are a lot of different changes some fixing issues, some refactoring. 

Would you be willing to break the changes in to smaller (more targeted) PRs? Or would you be willing to attend one of the monthly CSI community meetings to walk us through the changes?

lib/go/Makefile
huww98: This is just the latest version from https://github.com/protocolbuffers/protobuf/releases when I open this PR. I will update it again.

go.mod
huww98: removed

huww98:
> Would you be willing to break the changes in to smaller (more targeted) PRs?

It is hard. The old version does not support my arm64 MacBook. I have to upgrade the version first then tweak the Makefile to make it compile.

> Or would you be willing to attend one of the monthly CSI community meetings to walk us through the changes?

OK if that is necessary.

huww98:
I've reviewed the meeting record yesterday. About the GitHub action, I've fixed it in this PRs. Please see https://github.com/huww98/csi-spec/actions/runs/7736057831/job/21092762877 .

In the action, It downloads all dependencies, run protoc, then compile the generated go file to make sure it compiles at least. Plus I have tried to integrate it into csi-driver-host-path and csi-sanity manually. I think this change should be safe enough.

go.mod
```diff
@@ -5,12 +5,12 @@ go 1.18
 require (
 	github.com/golang/protobuf v1.5.3
```
jdef:
```
github.com/golang/protobuf v1.5.3
```
^ this keep creeping in, perhaps because of the go_opt flags pointing to it. i wonder if upgrading those package refs to avoid this repo would help?

huww98: this is already fixed in the 3rd commit of this PR, please check.

jdef:
Will look again, was looking at each commit and didn't make it through all 3

go.mod
```diff
@@ -3,14 +3,14 @@ module github.com/container-storage-interface/spec
 go 1.18
 
 require (
-	github.com/golang/protobuf v1.5.3
 	google.golang.org/grpc v1.57.0
+	google.golang.org/protobuf v1.31.0
```
jdef, line 7: protoc-gen-go 1.32 is compiled below, but here the requirement is 1.31. can we standardize on 1.32?

huww98: OK, fixed. I forgot to update this when I update the PR.

saad-ali:
@huww98 came to the [CSI meeting](https://docs.google.com/document/d/1-oiNg5V_GtS_JBAEViVBhZ3BYVFlbSz70hreyaD7c5Y/edit#heading=h.msboimirl3jq) and walked us through the changes. 

@huww98 has one more commit that may need to be added to this.

bswartz (changes requested):
This doesn't work on my system. The command "go env GOBIN" returns the empty string, and therefore I later get this error:

.protoc/bin/protoc -I../.. --go-grpc_out=csi --go_out=csi \
	--go_opt=paths=source_relative --go-grpc_opt=paths=source_relative \
	"csi.proto"
protoc-gen-go-grpc: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--go-grpc_out: protoc-gen-go-grpc: Plugin failed with status code 1.
make[1]: *** [Makefile:83: csi/csi.pb.go] Error 1
make[1]: Leaving directory '~/src/spec/lib/go'
make: *** [Makefile:20: lib/go/csi/csi_grpc.pb.go] Error 2

huww98:
@bswartz fixed that.

@saad-ali Since the meeting, I've added GOBIN env support. I've also updated the second commit to avoid adding phony target then removing it, although not visible in the final codebase.

bswartz (approved):
/lgtm

kasserater (commented):
one of the dependencies here has a fixed CVE, approved

PR #561, opened 2024-03-11
describe what if topology not match exactly

huww98:
**What type of PR is this?**

Describe what to expect when the accessible_topology of node and volume does not match exactly.

Fixes #

```release-note
none
```


spec.md
```diff
@@ -1050,6 +1050,11 @@ message Volume {
   // COs MAY use this information along with the topology information
   // returned by NodeGetInfo to ensure that a given volume is accessible
   // from a given node when scheduling workloads.
+  // The volume is accessible from all locations in the
+  // specified topology.
```
jdef, line 1054: prematurely line-wrapped?

spec.md
```diff
@@ -2767,6 +2772,10 @@ message NodeGetInfoResponse {
   // COs MAY use this information along with the topology information
   // returned in CreateVolumeResponse to ensure that a given volume is
   // accessible from a given node when scheduling workloads.
+  // The node can access some of the locations in the
+  // specified topology.
```
jdef, line 2776: prematurely line-wrapped?

spec.md
```diff
@@ -1050,6 +1050,11 @@ message Volume {
   // COs MAY use this information along with the topology information
   // returned by NodeGetInfo to ensure that a given volume is accessible
   // from a given node when scheduling workloads.
+  // The volume is accessible from all locations in the
+  // specified topology.
+  // The volume is accessible from a node if the volume's
```
jdef, line 1055: ... from a node if at least one of the volume's accessible topologies is a superset of ...

because a volume can have multiple accessible topologies

spec.md
```diff
@@ -2767,6 +2772,10 @@ message NodeGetInfoResponse {
   // COs MAY use this information along with the topology information
   // returned in CreateVolumeResponse to ensure that a given volume is
   // accessible from a given node when scheduling workloads.
+  // The node can access some of the locations in the
+  // specified topology.
+  // The node can access a volume if the node's accessible_topology
+  // is a subset of the volume's accessible_topology.
```
jdef, line 2778: .. is a subset of at least one of the volume's accessible topologies.

jdef (commented):
I believe that this PR clarifies the original intent of topology matching, thanks!

I'm undecided on the risk factor, and have limited brain cycles to spend here at the moment. Curious if anyone else in the community is concerned about potential breakage this language change might imply.

spec.md
huww98: I've added "union of", since I am already using the term of set theory.

PR #563, opened 2024-04-12
describe ABORTED as error message for CreateVolumeGroupSnapshot

nixpanic:
**What type of PR is this?**

Documentation enhancement for a common CreateVolumeGroupSnapshot error.

The `ABORTED` error code is currently not documented for CreateVolumeGroupSnapshot but is explained for CreateSnapshot.

Fixes #

```release-note
none
```


saad-ali (changes requested):
Per @bswartz --
* "Not clear what expected behavior if GroupSnapshot for volume A, B, C pending and another one for group B, C, D is triggered"
* "Probably ok if there are 2 overlapping group snapshots and SP decided to return ABORTED for 2nd snapshot and let you retry." So it's probably ok not to explicitly mention this case.

Per @xing-yang: "Specified group snapshot, could mean all parameters have to be the same". 

Next step:
* Let's clarify what "specified group snapshot" means.

saad-ali:
Also, @nixpanic, please make sure you've signed the CLA.

nixpanic:
> Per @bswartz --
> 
>     * "Not clear what expected behavior if GroupSnapshot for volume A, B, C pending and another one for group B, C, D is triggered"
> 
>     * "Probably ok if there are 2 overlapping group snapshots and SP decided to return ABORTED for 2nd snapshot and let you retry." So it's probably ok not to explicitly mention this case.

If there is a conflict with the volumes that are part of the group snapshot that should be created, the `FAILED_PRECONDITION` error is more suitable to return. I do not think `ABORTED` needs to cover a case like this (too).

> Per @xing-yang: "Specified group snapshot, could mean all parameters have to be the same".

Yes, that is more suitable. This `ABORTED` error is for the `CreateVolumeGroupSnapshot` call, and is intended to match the similar error that `CreateVolume` can return.

> Next step:
> 
>     * Let's clarify what "specified group snapshot" means.

I've rephrased _specified group snapshot_ to _group snapshot with the exact same parameters_ and similar.


> Also, @nixpanic, please make sure you've signed the CLA.

This was completed with a group approval for IBM employees, sent by Brad Topol in the beginning of the month :partying_face:

bswartz (approved):

nixpanic:
@bswartz , this seems to be pending still. Could you let me know if there is something I need to do?

PR #571, opened 2024-08-23
Empty block contents

ebblake:
**What type of PR is this?**
Doc clarification, plus feature proposal

Clarify that by default, a Plugin MUST ensure that a newly-created volume reads as all zeroes, and additionally offer a way to allow Plugins to create volumes more quickly by skipping this guarantee.

Fixes #

The first commit should be usable whether or not the community agrees with the proposal of the second.

NONE (the new option proposed in the second patch is backwards-compatible)
```release-note
none
```


nixpanic:
Explicitly mentioning that the filesystem on a new volume should be empty (except for possible metadata like `lost+found`) and that initial data on a block-mode volume should be read as zero's sounds good to me.

But I'd not add an option for uninitialized data (from previous usage). Users like things to go faster and enable it, unaware op the side effects.

PR #572, opened 2024-08-26
Regenerate with protoc-gen-go 1.3.5

### LINKED ISSUE #570: 1.10 generated files breaking changes
Update to https://github.com/container-storage-interface/spec/pull/552/ has broken quite some Kubernetes CSI sidecars. 

So far I identified two issues:

1. [gomock](https://github.com/golang/mock) cannot easily compare two messages now. Protobuf now stores some private stuff in `state` field, e.g. `NodeGetInfoResponse`: https://github.com/container-storage-interface/spec/blob/2f340622302a8d26950aba0f73f240ea1132a67b/lib/go/csi/csi.pb.go#L4827-L4830 And this private field has quite random values.

    It is possible to implement a custom gomock matcher that compares just the public fields, or use `proto.Equal`, as suggested in https://protobuf.dev/reference/go/faq/#deepequal.

1. It's not possible to *copy* message content. For example, [this](https://github.com/kubernetes-csi/external-provisioner/blob/6471a8e9c45da5df54c7ae8af2fcfc93f08b72da/cmd/csi-provisioner/csi-provisioner.go#L314-L318) code copies value of NodeGetInfoResponse:

    ```go
    nodeInfo, err := ctrl.GetNodeInfo(grpcClient, *operationTimeout)
    if err != nil {
        klog.Fatalf("Failed to get node info from CSI driver: %v", err)
    }
    nodeDeployment.NodeInfo = *nodeInfo
    ```
    Such a copy will copy also the private fields and one of the fields is [sync.Mutex](https://github.com/protocolbuffers/protobuf-go/blob/21d187aa53794eb29c1b5d94dac44987433e283a/internal/impl/message.go#L40), which should not be copied (`govet` complains hard about it).
    In this particular case, the Kubernetes external-provisioner can use a pointer, it does not need to copy the struct content.

Both issues are solvable using proper protobuf functions ([WIP PR here](https://github.com/kubernetes-csi/external-provisioner/pull/1262)), but they were quite surprising for a minor go package bump.

I am not asking for a revert, just for a better communication of the breaking changes and semantic versioning of go packages. This should not be a minor package bump. There should be probably a separate repo with the generated go code, so we can bump the versions independently on CSI spec. Kubernetes solved this by calling all its go packages 0.x, so no go API stability should be expected, and they break their API from time to time.

jsafrane:
Regenerated csi.pb.go with older protoc-gen-go to restore grpc message structures.

go.mod has github.com/golang/protobuf 1.5.3, which has a protoc-gen-go that generates code that introduces breaking changes, see #570.

Instead of 1.5.3, download github.com/golang/protobuf v1.3.2 separately, build protoc-gen-go from it and use it to generate csi.pb.go. The code is copied from the previous release of CSI spec, only with minor changes.

A future update to 1.5.3 must be done carefully and announce a breaking change.

Fixes #570

```release-note
Restored generated protobuf go message structures compatible with CSI spec v1.9.0.
```


lib/go/Makefile
```diff
@@ -58,15 +56,37 @@ $(PROTOC):
...
+########################################################################
+# Use protoc-gen-go v1.3.5. This ensures backward compatible go structures.
+# TODO: announce a breaking change and replace the whole chunk with
+# protoc-gen-go version from go.mod:
+# go install google.golang.org/protobuf/cmd/protoc-gen-go
+PROTOC_GEN_GO_PKG := github.com/golang/protobuf/protoc-gen-go
+PROTOC_GEN_GO := protoc-gen-go
+HERE := $(shell pwd)
+$(PROTOC_GEN_GO): PROTOBUF_PKG := $(dir $(PROTOC_GEN_GO_PKG))
+$(PROTOC_GEN_GO): PROTOBUF_VERSION := v1.3.2
```
huww98, line 71:
```suggestion
$(PROTOC_GEN_GO): PROTOBUF_VERSION := v1.3.5
```

lib/go/Makefile
```diff
@@ -46,8 +46,6 @@ PROTOC_URL := https://github.com/protocolbuffers/protobuf/releases/download/v$(P
 PROTOC_TMP_DIR := .protoc
 PROTOC := $(PROTOC_TMP_DIR)/bin/protoc
 
-$(GOBIN)/protoc-gen-go: ../../go.mod
-	go install google.golang.org/protobuf/cmd/protoc-gen-go
```
huww98, line 50: Why not just change this line to
```
go install github.com/golang/protobuf/protoc-gen-go@v1.3.5
```
So we don't need any other changes to the Makefile

huww98 (commented):
Should run `go mod tidy` to move `github.com/golang/protobuf` to direct dep.

Sorry for the inconvenience of introducing breaking changes. Could you suggest an alternative way to release the upgrade in the future? Maybe split the golang generated code to another repo (or place the go.mod in `lib/go` subdirectory), and only release v0 for the new module. If someone want old version of protoc-gen-go, he can always generate his own structs.

huww98:
Also note this PR does not revert the breaking change of requiring embedding `UnimplementedIdentityServer`.

lib/go/Makefile
jsafrane: This did not work for me, it wanted to downgrade other packages in go.mod

huww98: That's strange. I tried this and it works for me. `github.com/golang/protobuf` stays at v1.5.3 in go.mod

jsafrane: Your're right, I was installing `google.golang.org/protobuf/cmd/protoc-gen-go@v1.3.5` and not `github.com/golang/protobuf/protoc-gen-go@v1.3.5`.

I updated the PR, now it's a super simple PR.

OKEAMAH (approved):
Z

saad-ali:
Closing per https://github.com/container-storage-interface/spec/issues/570#issuecomment-2360257012

PR #578, opened 2025-01-21
Fix resource exhausted typo

ebblake:
**What type of PR is this?**
Bug fix to the specification.

Inconsistent specs are hard to implement.

Fixes #

```release-note
none
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2030,8 +2030,8 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
 | Snapshot already exists but is incompatible | 6 ALREADY_EXISTS | Indicates that a snapshot corresponding to the specified snapshot `name` already exists but is incompatible with the specified `volume_id`. | Caller MUST fix the arguments or use a different `name` before retrying. |
+| Not enough space to create snapshot | 8 RESOURCE_EXHAUSTED | There is not enough space on the storage system to handle the create snapshot request. | Caller SHOULD fail this request. Future calls to CreateSnapshot MAY succeed if space is freed up. |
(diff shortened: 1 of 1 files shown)


ebblake:
I found this while trying to determine what error to report.   CreateVolume mentions both RESOURCE_EXHAUSTED and OUT_OF_RANGE but without a clear statement of which is better when running out of space; while CreateSnapshot clearly wants RESOURCE_EXHAUSTED but ControllerExpandVolume wants OUT_OF_RANGE.  In my view, OUT_OF_RANGE applies to things like "requested size needs to be rounded larger than requested limit" or "limit is smaller than the source volume size" (something you can't fix no matter how often you retry with the same arguments), while RESOURCE_EXHAUSTED is for errors that may or may not be transient if more storage resources are freed up in the meantime).  I also note that https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1790-recover-resize-failure/README.md#making-allocatedresourcestatus-not-change-unnecessarily-for-every-error-in-131 mentioned OUT_OF_RANGE but not RESOURCE_EXHAUSTED as a plausible reason to back off the retry loop.

jdef:
thanks for the PR. please sign the CLA if you've not done so already.

ebblake:
> thanks for the PR. please sign the CLA if you've not done so already.

I have: https://github.com/kubernetes-sigs/contributor-playground/pull/1452
Or is there a different CLA for this project?

ebblake:
> Diff. See the Ccla file in top dir
Thanks - I'm chasing down whether Red Hat has already signed such a file, and if so, whether it is blanket for all employees or whether I need to be added to a list.

ebblake:
> Ping @ebblake. Do you have a signed CLA yet?

I've asked the Red Hat legal department, but so far no response. :(

jdef:
There's at least one redhat contributor here already, and I'd be surprised
if RH didn't already sign the CLA for all its folks.


James DeFelice

bswartz:
I want to merge this but it looks like there's a merge conflict I can't resolve through github

PR #580, opened 2025-03-02
Remove "optional" for GetVolumeGroupSnapshot to avoid confusion

xing-yang:
**What type of PR is this?**
cleanup

This PR removes "optional" from the following description about `GetVolumeGroupSnapshot`. Technically this sentence is correct as `GetVolumeGroupSnapshot` is optional. However, it adds confusion as the sentence immediately following this says it must be implemented if a plugin has `CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT` capability. For volume group snapshots, there is a single capability for create/delete/get volume group snapshots. If that capability is supported, all there RPCs create/delete/get must be implemented.

```
This optional RPC MAY be called by the CO to fetch current information about a volume group snapshot.

A Controller Plugin MUST implement this `GetVolumeGroupSnapshot` RPC call if it has `CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT` capability.
```

Fixes #

```release-note
none
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3086,7 +3086,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 
 #### `GetVolumeGroupSnapshot`
 
-This optional RPC MAY be called by the CO to fetch current information about a volume group snapshot.
+This RPC MAY be called by the CO to fetch current information about a volume group snapshot.
 
 A Controller Plugin MUST implement this `GetVolumeGroupSnapshot` RPC call if it has `CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT` capability.
 


jdef:
I seem to remember similar language in the spec for other RPCs. Whatever we
decide to do here, we should be consistent everywhere.

xing-yang:
I found similar language for ControllerGetVolume RPC.  The only difference is that GET_VOLUME capability is only for ControllerGetVolume RPC, while CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT capability is for Create/Delete/GetVolumeGroupSnapshot RPCs.

Should we remove "optional" for  ControllerGetVolume as well?

```
This optional RPC MAY be called by the CO to fetch current information about a volume.

A Controller Plugin MUST implement this ControllerGetVolume RPC call if it has GET_VOLUME capability.
```

bswartz:
> Should we remove "optional" for ControllerGetVolume as well?

Yes, we should.

PR #586, opened 2025-04-17
Add GetSnapshot RPC

xing-yang:
**What type of PR is this?**
feature

This PR adds a GetSnapshot RPC

Fixes #

```release-note
Adds GetSnapshot RPC.
```


spec.md
```diff
@@ -2147,6 +2156,50 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
 
+#### `GetSnapshot`
+
+**ALPHA FEATURE**
+
+This optional RPC MAY be called by the CO to fetch current information about a snapshot.
+
+A Controller Plugin MUST implement this `GetSnapshot` RPC call if it has `GET_SNAPSHOT` capability.
+
+SP supporting Group Controller Service and the CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT capability SHOULD implement the `GetSnapshot` RPC in order to retrieve information for individual snapshots that are part of a group snapshot.
```
bswartz, line 2167: An SP that supports...

spec.md
```diff
@@ -2147,6 +2156,50 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+`GetSnapshotResponse` should contain current information of a snapshot if it exists.
+If the snapshot does not exist any more, `GetSnapshot` should return gRPC error code `NOT_FOUND`.
```
bswartz, line 2170: Should we also say that snapshots which were recently created and not yet ready_to_use must be retrievable with this RPC? I think it's kind of obvious, but it would be good to change the current mechanism of polling for ready_to_use to rely on GetSnapshot instead of CreateSnapshot for SPs that support it in the future. Having some language here would make us feel safer about making that change.

bswartz (changes requested):

spec.md
xing-yang: Done

spec.md
xing-yang: Done

bswartz (approved):
/lgtm

nixpanic (commented):
Looks good to me, thanks! You probably want to squash the commits into one before it gets merged?

xing-yang:
Squashed the commits.

nixpanic (approved):
Thanks!

It has obviously been an omission, and I can imaging COs would want to use this. Can you leave a link in the comments where this new procedure would be used?

bswartz:
> Thanks!
> 
> It has obviously been an omission, and I can imaging COs would want to use this. Can you leave a link in the comments where this new procedure would be used?

It's expected that statically provisioned group snapshots would use this API to populate information about the individual members of the group snapshot. Today we don't have a good way of getting the necessary information in that case. One hack is to rely on ListSnapshots with a snapshot ID passed in, but many drivers don't support that, due to other challenges related to snapshot listing.

humblec (approved):

jdef:
Is there a KEP that describes the use case? Unclear why the list API
doesn't work here.


James DeFelice

xing-yang:
Here is the KEP: https://github.com/kubernetes/enhancements/pull/5013/files
ListSnapshots API works as well, however, many storage vendors could not support list snapshots due to the performance overhead.
Usually storage vendors do not have a problem supporting GetSnapshot if they want to support volume snapshot or volume group snapshot.

jdef:
Thanks for explaining

James DeFelice

jdef:
Lgtm

PR #591, opened 2025-06-24
Documented FAILED_PRECONDITION as a possible error for GetMetadataDelta

carlbraganza:
If changed block tracking has not been enabled in the underlying storage subsystem then GetMetadataDelta will fail.
This PR documents the gRPC error code that should be returned when this happens.

**What type of PR is this?**
This is a documentation change only.

This provides information to CSI driver developers on the precise error code to return for this error, and
guidance to the backup application developer on how to respond.

Fixes #

This change was discussed in the Data Protection Working Group meeting.

```release-note
NONE
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3343,7 +3343,7 @@ The following conditions are well defined:
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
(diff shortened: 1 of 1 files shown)


jdef:
s/operation would work/operation would proceed/

bswartz (approved):
/lgtm

akankshapanse:
@carlbraganza will the similar change be required for GetMetadataAllocated() API as well, mainly for cases like CBT is not enabled on underlying disk/data copy ?

PR #597, opened 2025-07-24
Collapse the INVALID_ARGUMENTS error rows and clarify

bswartz:
This clarifies that the INVALID_ARGUMENT error code for ControllerModifyVolume is safe to not retry (i.e. "infeasible")

```release-note
none
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1711,8 +1711,7 @@ message ControllerModifyVolumeResponse {
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
-| Parameters not supported | 3 INVALID_ARGUMENT | Indicates that the CO has specified mutable parameters not supported by the volume. | Caller MAY verify mutable parameters. |
-| Exceeds capabilities | 3 INVALID_ARGUMENT | Indicates that the CO has specified capabilities not supported by the volume. | Caller MAY verify volume capabilities by calling ValidateVolumeCapabilities and retry with matching capabilities. |
+| Parameters not supported | 3 INVALID_ARGUMENT | Indicates that the CO has specified invalid mutable parameter keys or values, or the specified volume cannot support the specified parameters. The SP MUST NOT have applied any modification to the volume as part of this specific call. | A caller may verify volume capabilities by calling ValidateVolumeCapabilities and then retry the request with valid mutable parameters. |
 | Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified volume_id does not exist. | Caller MUST verify that the volume_id is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
 
 #### `GetCapacity`


huww98:
This could be a breaking change to SP. e.g. AWS does not comply with this now.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/2ba0d6c473fffe0e45eb99a973628884123313b5/pkg/driver/controller_modify_volume.go#L153

gnufied:
>This could be a breaking change to SP. e.g. AWS does not comply with this now.

In CSI this feature is alpha, this is why it is important we tighten the wording before this feature goes GA in CSI.

huww98:
I'm afraid this may be too much to ask for SP. In the previous AWS case, whether the parameter is invalid is determined by server, and cannot be determined solely in CSI. If combined with multi-stage modification, we may not able to verify all parameters of all stage before making any changes. I don't have such specific use-case though, but I expect we will have such case in the future.

So if this change is applied, we will not able to return "invalid argument" for the second stage. We could replace them with Internal error. But IMO this is not a good design, because SP could easily implementing this wrong.

xing-yang (approved):
lgtm

huww98:
Take AWS EBS as an example. It needs to modify both disk and tag. Obviously parameters to both stage can contains invalid values.

The current implementation is modify tag first. After that succeeded, modify the disk. So if the CSI restarts after tag has been modified, but before disk. Then it find invalid argument when modifying disk in the new CSI process. How can it rollback the changes to tags? It does not know the tag value before modification.

Maybe AWS guys can swap the order, modify the disk first, then tag. In that case, if the tags contains invalid argument, CSI even cannot rollback the changes, because we need to wait for 6 hours to modify the disk again. Maybe CSI can validate the tags locally before doing anything. But we cannot be 100% sure the local validation logic is in sync with server.

> Obviously this requires a round-trip to the actual server, but I can't think of a reason that the server couldn't validate all of the parameters as a unit.

Not one round-trip, but multiple. 3 for AWS, 4 for Alibaba Cloud currently. Server only receives some of the parameters in any of the round-trip.

> Isn't this how transactional changes work?

That is ideal. But do you expect CSI to implement an undo log system for this, as database usually do? It is way too complex. And CSI usually does not have access to persistent storage itself. It is generally very hard to make a workflow in complex distributed system transactional. The best we can promise is idempotency, so that the workflow can be retried.

spec.md
```diff
@@ -1711,8 +1711,7 @@ message ControllerModifyVolumeResponse {
 
 | Condition | gRPC Code | Description | Recovery Behavior |
 |-----------|-----------|-------------|-------------------|
-| Parameters not supported | 3 INVALID_ARGUMENT | Indicates that the CO has specified mutable parameters not supported by the volume. | Caller MAY verify mutable parameters. |
-| Exceeds capabilities | 3 INVALID_ARGUMENT | Indicates that the CO has specified capabilities not supported by the volume. | Caller MAY verify volume capabilities by calling ValidateVolumeCapabilities and retry with matching capabilities. |
+| Parameters not supported | 3 INVALID_ARGUMENT | Indicates that the CO has specified invalid mutable parameter keys or values, or the specified volume cannot support the specified parameters. | Caller MAY verify volume capabilities by calling ValidateVolumeCapabilities and retry with matching capabilities. Indicates that no changes were made by the SP. |
```
huww98, line 1714: Yes, I agree that if we accept this PR, we could replace INVALID_ARGUMENT in the half-way with INTERNAL. It is viable. But do note that INVALID_ARGUMENT also have the special meaning that external-resizer will retry at lower frequency. We will loss that if we use INTERNAL.

```suggestion
| Parameters not supported | 3 INVALID_ARGUMENT | Indicates that the CO has specified invalid mutable parameter keys or values, or the specified volume cannot support the specified parameters. | Caller MAY verify volume capabilities by calling ValidateVolumeCapabilities and retry with matching capabilities. Indicates that no changes were made by the SP in this call and any previous calls with the same arguments. |
```
I think we should add this to make it safe for CO to abandon the operation after many retries.

huww98:
But why we need to abandon the operation in the first place? If modify from A to B failed, and user request A to CO again, CO then request A again to SP, it is simple and easy. And we don't need to this PR, or asking AWS to make the change.

spec.md
bswartz: The "slow retry" behavior is not required by CSI, it's just what Kubernetes decides to do when the error is infeasible, because Kubernetes is a declarative system.

If you return INTERNAL, you will get a normal (faster) retry, which is presumably desirable if the operation was merely interrupted and can proceed with another attempt.

For situations where the SP is truly stuck and there was no way to detect that it would get stuck in advance, the only remedy is feedback to the user/operator and someone making a change to the request, or going to the underlying system and fixing some underlying problem that's invisible to the CO/SP.

In the case of VAC, the possible fixes are:
* The user changes the PVC's VAC back to the old one, or some other one and expects it to reconcile successfully
* An admin creates a new VAC with correct parameters (fixing whatever the problem was with the given VAC) and sets the PVC to that one, and expects it to reconcile successfully
* An operator goes straight to the underlying storage system and fixes whatever is preventing the original mutation

bswartz: In the CSI spec we try to be very hands-off on how the SPs achieve their goals. All we care about is the contract between them and the CO. This is frustrating to SP authors, but it's genuinely challenging to write a spec that is both as flexible as CSI and also reliable as the Kubernetes-side implementation of the spec.

Consider that this feature was originally envisioned as a way to set QoS policies on volumes and change them at a later time. I think it's pretty obvious how one would validate QoS policies and then mutate them idempotently. Other features that follow the same pattern are a good fit for this feature.

If there are use cases where the simple QoS policy pattern doesn't fit, we need to discuss those concrete use cases, otherwise, we assume that SPs can do whatever they want and they own all the problems. The specific issue here is that the CO has a problem and the only way to resolve it is to be more prescriptive in the spec, which would be a breaking change if it were not alpha, so we're trying to do it now.

sunnylovestiramisu: Also how about we use some stronger wording:
 	
``` 
Indicates that one or more arguments in the request is invalid. The SP MUST NOT have applied any modification to the volume as part of this specific call. A caller may verify volume capabilities by calling ValidateVolumeCapabilities and then retry the request with valid arguments.
```

And about the case:

1. Modify from A to B: partially applied, INTERNAL is returned by SP
2. Modify again to C: INVALID_ARGUMENT is returned by SP

The initial non-atomic failure has put the volume into an undefined and potentially unrecoverable state. When the modification from A to B failed and was only partially applied, the SP violated the spec's requirement for atomicity. The volume is now in a state (A') that was never intended to exist. This A' state is effectively "corrupt" from the perspective of state machine. It doesn't match the last known successful configuration (A), nor does it match the target configuration (B).

For these cases I think the inescapable conclusion is "**Manual Intervention is Required**". Do we want to write it into the Spec though? Beause Secs should focus on defining correct behavior and prevention, not cure.

bswartz:
Okay @gnufied explained to me the point of the target VAC and the need for rollback support, because of quotas and concerns about exploitation of the quota system in case of failures. Based on that understanding I still think this CSI spec change makes sense, and non-conforming drivers will need to update.

spec.md
huww98: > the SP violated the spec's requirement for atomicity.

I don't think CSI requires atomicity. And it cannot require such, because at least AWS and Alibaba Cloud cannot promise atomicity.

> For these cases I think the inescapable conclusion is "Manual Intervention is Required"

OK for require more complex "Manual Intervention". But not OK to leave volume at unintended state undetectable by user, because that can violates quota system, and can be utilized by malicious users.

> If an SP does perfect validation up front, and all the later errors are recoverable through retries, then recovery/rollback will never be needed. It's a good reason to improve validation.

That is too idea IMO. I agree we should improve validation. But as CSI spec, we should also design for edge cases. obviously not all errors are recoverable through retries. In other word, when the volume is half modified, we require SP to:
* not return INVALID_ARGUMENT: fine, we can replace them with INTERNAL.
* not return final error: unacceptable, could leave volume at broken state forever.

huww98: > Also how about we use some stronger wording

Why this is stronger? I'm considering such SP implementation:

The modification has 2 part
1. part 1 is applied, part 2 has invalid argument, but since the volume is already changed, SP returns INTERNAL
2. CO retries the request, since part 1 is already applied, so no changes to the volume in this specific call, SP may return INVALID_ARGUMENT then.

This SP is also complying to the proposed CSI spec, but it is still not safe for CO to abandon the operation.

huww98:
> it's valid to change the desired state to a different value before the actual state is updated.

Yes, it is totally valid. I want CO to never abandon an operation without notifying SP. I want the CO to think the actual state is unknown when it starts the first `ControllerModifyVolume`. So if modify from A to B failed for any reason, the actual state is neither A nor B, CO should call `ControllerModifyVolume` again to go forward or backward. I've implemented this in https://github.com/kubernetes-csi/external-resizer/issues/513

But if we really want to proceed this PR, we should be careful about cases mentioned in https://github.com/container-storage-interface/spec/pull/597#discussion_r2232375030 and https://github.com/container-storage-interface/spec/pull/597#discussion_r2231368477

AndrewSirenko (commented):
Lgtm

The EBS CSI Driver returning infeasible despite modifying tags is a bug that we put on our backlog to fix.

spec.md
bswartz: > I don't think CSI requires atomicity. And it cannot require such, because at least AWS and Alibaba Cloud cannot promise atomicity.

I think this was a either a typo or a there is confusion. The CSI does *does not* require atomicity, as you say. What it does require is idempotency. I think Sunny was suggesting that the error checking portion of the SP implementation must be "read only" such that -- from the CO's point of view, an ILLEGAL_ARUGMENT failure can be handled atomically.

> I agree we should improve validation. But as CSI spec, we should also design for edge cases. obviously not all errors are recoverable through retries. In other word, when the volume is half modified, we require SP to:

All we're trying to communicate here is that it's safe for the CO to give up after receiving an ILLEGAL_ARUGMENT error. The CO has to consider issues like quota management, and wants to be able to revert quota consumption if the volume modification never happened.

Any error where the SP made some change and then failed should report a different error, because the CO needs to retry in that case, and not release quota consumption, to avoid a situation where users are able to game the k8s quota system by intentionally applying VACs that they know will fail.

bswartz: > > Also how about we use some stronger wording
> 
> Why this is stronger? I'm considering such SP implementation:
> 
> The modification has 2 part
> 
>     1. part 1 is applied, part 2 has invalid argument, but since the volume is already changed, SP returns INTERNAL

We are saying, if there are two parts, you have to validate both parts before applying part 1. We're discouraging SPs doing a sequence of validate, apply, validate, apply, etc. Do all the validation up front, and only proceed if you expect all the changes to be successful. Any failures in that case are unexpected, and deserve a retry.

bswartz: > Yes, I agree that if we accept this PR, we could replace INVALID_ARGUMENT in the half-way with INTERNAL. It is viable. But do note that INVALID_ARGUMENT also have the special meaning that external-resizer will retry at lower frequency. We will loss that if we use INTERNAL.
> 
> I think we should add this to make it safe for CO to abandon the operation after many retries.
> 
> Even if we add the words suggested above. It is still not perfectly safe to abandon the operation. For example as I mentioned in the slack thread:
> 
>     1. Modify from A to B: partially applied, INTERNAL is returned by SP
> 
>     2. Modify again to C: INVALID_ARGUMENT is returned by SP
> 
> 
> Now, although the SP is completely complying to the CSI spec suggested here, we cannot safely abandon the operation because the volume is in half A half B state.

If some some reason A -> B passes validation, but still fails for unexpected reasons (i.e. INTERNAL error), my understanding is that the CO will continue retrying the modification to B, but if the user changes the VAC to C, then the CO will accept that if C contains all the mutable parameter keys that B did. If C doesn't contain all the keys that B did, it will keep trying to modify to B (this is for safety within the quota system).

If the SP rejects the modification to C with INVALID_ARGUMENT, then I suspect it gives up and goes back to trying to modify to B, because the CO interprets that error as if the VAC itself is invalid.

The point is that INVALID_ARGUMENT is a signal that the VAC itself is not usable, at least for that volume, and INTERNAL is a signal that the VAC was fine, and the underlying system has some problem, in which the normal kubernetes approach is to keep trying.

huww98: > We are saying, if there are two parts, you have to validate both parts before applying part 1. We're discouraging SPs doing a sequence of validate, apply, validate, apply, etc.

I understand this. I'm trying to say, CSI spec needs to be precise, we need to formalize this into the spec. Just discouraging is not enough IMO.

> the CO will accept that if C contains all the mutable parameter keys that B did

This is a solution, but is not a good one IMO. CSI spec said "COs will treat these (mutable_parameters) as opaque", which I think conflicts with this rule. In practice, consider this case:
1. User modify to VAC A with `{"type": "cloud_essd", "performanceLevel": "PL3"}` and failed with INTERNAL error (out of stock, for example).
2. User then try to recover by modify again to VAC B with `{"type": "cloud_auto"}`, where `cloud_auto` does not have `performanceLevel` parameter.

With your plan, CO should not accept B. But this is valid and safe from SP's PoV.

> If the SP rejects the modification to C with INVALID_ARGUMENT, then I suspect it gives up and goes back to trying to modify to B, because the CO interprets that error as if the VAC itself is invalid.

This can be viable. But I suspect it can be hard for CO to implement this. Consider this:
1. User modify to VAC A, failed with INTERNAL, because the credential used to communicate with storage backend is not correctly configured.
2. User then modify to VAC B, C, D then E, in sequence, all failed with INTERNAL.
3. Cluster admin correct the credential.
4. CO retries E, it failed with INVALID_ARGUMENT.
5. CO give up E then retries D, it also failed with INVALID_ARGUMENT.
6. CO retries C, B, A in sequence, all failed with INVALID_ARGUMENT.
7. now CO is safe to abandon this operation.

As you can see, in order to implement this, CO will need to persistent a potentially unbounded stack of VACs it has tried. We don't have a place to store this in Kubernetes PVC.status now.

gnufied: Currently CO (k8s) doesn't look into mutable_parameters at all. I was proposing that as a theoretical option in case, if we were to guard against partial application of VACs (like a bulletproof option at the cost of flexibility). 

Basically at this point we concede that, recovering to a mutable_parameter that is a subset of proposed transition is lossy and it it is possible that volume will be left in a state where one or more of parameters were applied.  The only way CO can solve this is by preventing recovery to subset of `mutable_parameters`

>1. User modify to VAC A, failed with INTERNAL, because the credential used to communicate with storage backend is not correctly configured.
> 2. User then modify to VAC B, C, D then E, in sequence, all failed with INTERNAL.
> 3. Cluster admin correct the credential.
> 4. CO retries E, it failed with INVALID_ARGUMENT.
> 5. CO give up E then retries D, it also failed with INVALID_ARGUMENT.

CO will not automatically retry recovery to any of the failed states, unless user takes some action to recover. I am not sure where is this coming from. 

> My intuition is that it would be bad SP behavior to respond with INTERNAL first, and then INVALID_ARGUMENT later to the same volume/mutable parameters pair, 

In general not great, but CO does tolerate this.

huww98:
I think the spirit of this change is good. But even if it is merged, it will not resolve any issue we are facing.

If I understand correctly, this PR is intended to make it safe for CO to abandon a modify operation without notifying the SP.

Consider the following circumstances when modifying from state A, and user want to return to state A:
1. A -> B (INVALID_ARGUMENTS)
2. A -> B (TIMEOUT) -> B (INVALID_ARGUMENTS)
3. A -> B (INTERNAL) -> B (INVALID_ARGUMENTS)
4. A -> B (INTERNAL) -> C (INVALID_ARGUMENTS)
5. A -> B (INTERNAL) -> C (INTERNAL) -> D (INTERNAL) -> E (INVALID_ARGUMENTS)

With this change, we only able to safely abandon the operation in case 1, which is the simplest. But if we do implement this feature in CO, we really want to also support case 2 and 3, because we don't want a random component restart or network error to break this feature. Case 4 and 5 are better-to-have though, CO may just not supporting canceling in this case.

So CO really cannot take any advantage of this new promise from SP. IMO we should either make stronger requirement to SP, or don't make this change and leave more freedom to SP.

---

While I've asked this before, it is still unanswered. Why would CO want to abandon the operation in the first place? Previously, we call ControllerUnpublishVolume to cancel a failed ControllerPublishVolume; we call ControllerExpandVolume again with different size to cancel a failed ControllerExpandVolume. All works well. Why can't we just do the same, call ControllerModifyVolume again with different parameter to cancel a failed ControllerModifyVolume? Do you want to avoid an extra gRPC call and save 100ms?

> In the case of expand, it's valid to change the desired state to a different value before the actual state is updated. I think the same should be true of VAC changes.

@bswartz this is your previous reply. I'm not sure but are you supporting my opinion that we never need to abandon the operation? If CO failed to reconcile to the desired state B and user set new desired state back to A. Can we just reconcile to state A again, instead of assuming the volume is not modified from state A and abandon the operation?

spec.md
huww98: > Basically at this point we concede that, recovering to a mutable_parameter that is a subset of proposed transition is lossy and it it is possible that volume will be left in a state where one or more of parameters were applied. The only way CO can solve this is by preventing recovery to subset of mutable_parameters

We can also ask cluster admin to take responsibility of this. We can write this in the doc: if you want your quota to work as expected, you should fully specify the volume state in each VAC.

> CO will not automatically retry recovery to any of the failed states, unless user takes some action to recover. I am not sure where is this coming from.

@gnufied You proposed "I suspect it gives up and goes back to trying to modify to B", and I'm trying to generalize that "goes back" to "recursively goes back" when there are even more VACs involved.

> My intuition is that it would be bad SP behavior to respond with INTERNAL first, and then INVALID_ARGUMENT later to the same volume/mutable parameters pair

@bswartz for example, if SP got "network connection timeout" while trying to verify the arguments, I think the best it can do is returning INTERNAL.

gnufied:
As per our discussion in CSI implementation meeting.  Error handling for each of the cases is now documented in - https://github.com/kubernetes/enhancements/pull/5462 

This change to CSI spec should be IMO merged, because while it is not perfect it is better than any other alternative.

bswartz:
I have squashed the commits so this is read to merge

sunnylovestiramisu:
+1 it is good to merge

/lgtm

saad-ali (approved):
/lgtm

Thanks for the helping improve this.

PR #603, opened 2026-03-21
Add ControllerGetNodeInfo RPC (alpha)

huww98:
**What type of PR is this?**

/kind feature

See KEP-6011: CSI ControllerGetNodeInfo for details

Fixes #

```release-note
Add ControllerGetNodeInfo RPC and corresponding new field in NodeGetInfo (alpha)
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -453,11 +453,6 @@ service Node {
 
   rpc NodeGetInfo (NodeGetInfoRequest)
     returns (NodeGetInfoResponse) {}
-
-  rpc NodeGetID (NodeGetIDRequest)
-    returns (NodeGetIDResponse) {
-        option (alpha_method) = true;
-    }
 }
 ```
 
@@ -1731,7 +1726,7 @@ This is useful when the node side plugin cannot or should not access cloud APIs 
 
 If the SP also supports `PUBLISH_UNPUBLISH_VOLUME` controller capability, the CO MAY call this RPC to dynamically update `max_volumes_per_node` when volume attachment fails with `RESOURCE_EXHAUSTED`.
 
-The CO SHOULD call this RPC after obtaining the node ID via `NodeGetID`.
+The CO SHOULD call this RPC after obtaining the node ID via `NodeGetInfo`.
 
 The SP returns the maximum number of volumes that can be published to the node.
 The SP calculates this limit based on the instance type's attachment limit, accounting for non-volume resources that consume attachment slots (e.g., network interfaces on some instance types).
@@ -1751,8 +1746,7 @@ The CO MAY call this RPC multiple times for the same node.
 message ControllerGetNodeInfoRequest {
   // The identifier of the node as understood by the SP.
   // This field is REQUIRED.
-  // This field MUST match the node_id returned by `NodeGetInfo` or
-  // `NodeGetID`.
+  // This field MUST match the node_id returned by `NodeGetInfo`.
   // This field overrides the general CSI size limit.
   // The size of this field SHALL NOT exceed 256 bytes.
   string node_id = 1;
@@ -2877,14 +2871,15 @@ message NodeServiceCapability {
       // or node publish RPC calls.
       VOLUME_MOUNT_GROUP = 6;
 
-      // Indicates the SP supports the NodeGetID RPC.
-      // This enables COs to fetch only the node identifier from
-      // the node side without requiring cloud API credentials.
-      // The topology and capacity information can then be fetched
-      // via ControllerGetNodeInfo.
-      // If the SP supports GET_ID, it MUST also support
-      // GET_NODE_INFO controller capability.
-      GET_ID = 7 [(alpha_enum_value) = true];
+      // Indicates the SP supports the controller_get_node_info field
+      // in NodeGetInfoRequest. When the CO sets that field, the SP MAY
+      // omit accessible_topology and max_volumes_per_node from
+      // NodeGetInfoResponse and return only node_id, which the node
+      // side can obtain without cloud API credentials; the CO then
+      // fetches topology and capacity via ControllerGetNodeInfo.
+      // If the SP supports NODE_INFO_FROM_CONTROLLER, it MUST also
+      // support the GET_NODE_INFO controller capability.
+      NODE_INFO_FROM_CONTROLLER = 7 [(alpha_enum_value) = true];
     }
 
     Type type = 1;
@@ -2910,8 +2905,21 @@ The CO MAY call this RPC more than once for a given node.
 The SP SHALL NOT expect the CO to call this RPC more than once.
 The result of this call will be used by CO in `ControllerPublishVolume`.
 
+If the SP has the `NODE_INFO_FROM_CONTROLLER` node capability, the CO MAY set the `controller_get_node_info` field in the request.
+When set, the SP MAY omit `accessible_topology` and `max_volumes_per_node` from the response and return only `node_id`, which the node side can obtain without cloud API credentials (e.g., from local instance metadata).
+The CO obtains topology and capacity via `ControllerGetNodeInfo` instead, and MUST NOT consume `accessible_topology` or `max_volumes_per_node` from the response.
+
 ```protobuf
 message NodeGetInfoRequest {
+  // When true, the CO will obtain accessible_topology and
+  // max_volumes_per_node from ControllerGetNodeInfo. The SP MAY omit
+  // those two fields from NodeGetInfoResponse and return only node_id.
+  // The CO MUST NOT consume accessible_topology or max_volumes_per_node
+  // from a response to a request with this field set.
+  // The CO MUST NOT set this field to true unless the SP has the
+  // NODE_INFO_FROM_CONTROLLER node capability.
+  // This field is OPTIONAL.
+  bool controller_get_node_info = 1 [(alpha_field) = true];
 }
 
 message NodeGetInfoResponse {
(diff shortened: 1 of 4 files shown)


jdef:
There's a lot of k8s stuff in this PR desc, probably better suited for a
KEP that this PR might reference. This (CSI spec) is not a subproject of
k8s, it's CO agnostic, or at least it tries to be. Appreciate the context
though.

Reading this through it's not immediately clear to me why this isn't
solveable some other way- what alternatives were considered and why were
they not sufficient?

James DeFelice

huww98:
> There's a lot of k8s stuff in this PR desc, probably better suited for a
KEP that this PR might reference.

Yes, I will propose this to Kubernetes if it looks fine on the CSI side.
I provide them here for additional context.

> what alternatives were considered and why were
they not sufficient?

There are not many alternatives actually.

The instance metadata service (e.g., http://169.254.169.254 or http://100.100.100.200) typically only provides basic info like instance-id and zone. It does not provide disk attachment limits or supported disk categories, which require cloud API calls.

The information required by NodeGetInfo is not fetchable from node with the security requirement. We have considered adding a new private CRD and a new controller in Kubernetes. The controller watches for new node, fetches the required info and store them in a CR, then node watches for the CR to get the info. But this still have major drawbacks:
* The permission to invoke Cloud API is replaced with a service account with CR read permission, which does not fundamentally solve the security issue.
* Requires more network roundtrips
   * This proposal: Cloud -> csi-controller -> external-attacher -> APIServer(CSINode) -> scheduler
   * With CRD: Cloud -> new controller -> APIServer(CR) -> csi-node -> kubelet -> APIServer(CSINode) -> scheduler
* Even harder considering the new capability to fetch max_volumes_per_node immediately after publish failure. Will need some way for node to tell controller to do an update immediately, which requires write permission, making things even worse.
* This is Kubernetes specific, not resolving the security requirement for other CO.

The core insight is: `max_volumes_per_node` and `accessible_topology` is consumed by scheduler, not node. So this piece of infomation should not appear on node for optimal security, hence this proposal.

bswartz:
I don't see any point to the `NodeGetID` RPC. `NodeGetInfo` already returns a superset of the information.

huww98:
I design `NodeGetID` as a per-node switch for this feature, so that we can upgrade controller first, then upgrade nodes gradually: if node reports GET_ID cap, we go for the new workflow, otherwise we go for the old flow. We can also safely install new CSI driver (with new sidecars) in old Kubernetes cluster this way: old kubelet don't call NodeGetID, then new sidecar should not call ControllerGetNodeInfo, even if both sidecar and CSI driver support this.

As just discussed in the meeting, an operator may choose to remove credential from node side, so that NodeGetInfo can no longer retrieve topology info. I think it would be better for SP to fail NodeGetInfo instead of returning incomplete info.

Alternative: Adding a new field in NodeGetInfoResponse (say `incomplete`) as the switch stated above. But for old CO (kubelet) that don't understand this field, it may silently accept the incomplete topology or capacity, leading to confusing scheduling result. Also we may need to consider what to do if node and controller returns conflicting topology.

For the periodical update scalability issue, I take the number from https://kubernetes.io/docs/setup/best-practices/cluster-large/:
> No more than 150,000 total pods
> No more than 300,000 total containers

if every container have a volume, we can have 300,000 volumes in the cluster (we didn't tested this large scale). Our DescribeDisks API call can return at most 500 disks per-call. The default rate limit is 1000/60s in a region. So we can actually finish the query of the whole cluster in less than 1 min. So I think this is fine.

I can see that it could be hard for SPs that want to determine topology at controller side but capacity from node side. But I haven't come up with a better idea though.

torredil:
@huww98 This has a merge conflict now, please rebase when you get a chance. Feel free to ping when its updated and reviewable.

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
+  // calculating available slots.
+  //
+  // The CO MUST NOT use these volume IDs for any purpose other than
+  // slot accounting. In particular, the CO MUST NOT attempt to
+  // unpublish or otherwise operate on volumes it did not publish.
+  repeated string published_volume_ids = 3;
+}
+```
+
+##### ControllerGetNodeInfo Errors
```
torredil: Should we have an error for invalid argument? perhaps somehow when the node id mismatch similar to GetVolumeGroupSnapshot Errors.

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
+#### `ControllerGetNodeInfo`
+
+A Controller Plugin MUST implement this RPC call if it has `GET_NODE_INFO` controller capability.
+
+This RPC allows the CO to fetch node information (topology, published volumes, and maximum attachable volumes) from the controller side.
+This is useful when the node side plugin cannot or should not access cloud APIs to retrieve this information (e.g., for security reasons where cloud API credentials should not be distributed to nodes).
+
+If the SP also supports `PUBLISH_UNPUBLISH_VOLUME` controller capability, the CO MAY call this RPC to dynamically update `max_volumes_per_node` when volume attachment fails with `RESOURCE_EXHAUSTED`.
+
+The CO SHOULD call this RPC after obtaining the node ID via `NodeGetInfo`.
```
torredil: (the wording here is too relaxed). I think it would be clearer to explicitly state that when the controllergetnodeinfo capability is advertised,  the CO MUST call ControllerGetNodeInfo before using the node information for scheduling or publishing.

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
+
+The SP returns the maximum number of volumes that can be published to the node.
+The SP calculates this limit based on the instance type's attachment limit, accounting for non-volume resources that consume attachment slots (e.g., network interfaces on some instance types).
+
+The SP MAY also return the list of volume IDs that are currently published to the node according to the cloud provider's API. These volumes may not be published by CO, but still occupy slots reported by `max_volumes_per_node`.
```
torredil, line 1734: np: there should only be one sentence per line (same for line 1738 below): https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#markdown-style.

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
+The CO SHALL combine these volumes with its own records (deduplicating as needed) when calculating available slots.
+The CO MUST NOT use these volume IDs for any purpose other than slot accounting. In particular, the CO MUST NOT attempt to unpublish or otherwise operate on volumes it did not publish.
+
+If the SP cannot or chooses not to return this list, the SP SHOULD account for non-CSI volumes when calculating `max_volumes_per_node`.
+
+This operation MUST be idempotent.
+The CO MAY call this RPC multiple times for the same node.
+
+```protobuf
+message ControllerGetNodeInfoRequest {
```
torredil, line 1746: This RPC is alpha, its missing the `option (alpha_message) = true;` label:

```suggestion
message ControllerGetNodeInfoRequest {
  option (alpha_message) = true;

```

(same thing for `ControllerGetNodeInfoResponse`).

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
 | Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified volume_id does not exist. | Caller MUST verify that the volume_id is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
 
+#### `ControllerGetNodeInfo`
+
+A Controller Plugin MUST implement this RPC call if it has `GET_NODE_INFO` controller capability.
+
+This RPC allows the CO to fetch node information (topology, published volumes, and maximum attachable volumes) from the controller side.
+This is useful when the node side plugin cannot or should not access cloud APIs to retrieve this information (e.g., for security reasons where cloud API credentials should not be distributed to nodes).
+
+If the SP also supports `PUBLISH_UNPUBLISH_VOLUME` controller capability, the CO MAY call this RPC to dynamically update `max_volumes_per_node` when volume attachment fails with `RESOURCE_EXHAUSTED`.
```
torredil: if the CO didn't set `controller_get_node_info` for this node it already has `max_volumes_per_node` from `NodeGetInfo`, so this gives it a second source with nothing saying which one wins. I think this should be limited to nodes where the flag was set:

```suggestion
The CO MUST NOT call this RPC for a node unless it set `controller_get_node_info` in the `NodeGetInfoRequest` for that node.
If the SP also supports `PUBLISH_UNPUBLISH_VOLUME` controller capability, the CO MAY call this RPC to dynamically update `max_volumes_per_node` when volume attachment fails with `RESOURCE_EXHAUSTED`.
```

spec.md
```diff
@@ -1712,6 +1717,92 @@ message ControllerModifyVolumeResponse {
...
+  // no topological constraints declared for V.
+  //
+  // Example 1:
+  //   accessible_topology =
+  //     {"region": "R1", "zone": "Z2"}
+  // Indicates the node exists within the "region" "R1" and the "zone"
+  // "Z2".
+  Topology accessible_topology = 2;
+
+  // The volume IDs that are currently published to this node.
```
torredil, line 1782: If these aren't the same `volume_id` the CO used in `ControllerPublishVolume` it won't be able to tell which of these volumes are its own and will treat all of them as non-CSI:

```suggestion
  // The volume IDs that are currently published to this node, as of
  // the time of this call. For volumes published by the CO this MUST
  // be the same volume_id used in ControllerPublishVolume.
```

torredil (commented):
Largely lgtm, will take another look after its rebased.

PR #604, opened 2026-05-21
[ALPHA] Add volume health proposal

gnufied:
Implements spec as proposed in https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1432-volume-health-monitor/README.md

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -1057,8 +1057,10 @@ enum VolumeHealthErrorType {
   // being accessible from that node.
   INACCESSIBLE = 2;
 
-  // Permanent data loss is known or strongly suspected on underlying 
-  // volume.
+  // Permanent data loss is known or strongly suspected on the
+  // underlying volume. The CO MAY use this signal to inform
+  // users to trigger recovery workflows such as restoring from
+  // a snapshot or replica.
   DATA_LOSS = 3;
 }
 
@@ -1334,8 +1336,8 @@ message ControllerServiceCapability {
       // Indicates the SP supports the ControllerListVolumeHealth RPC.
       // This enables COs to fetch volume health information from
       // the Controller Plugin's perspective.
-      // A SP that supports ControllerListVolumeHealth MUST also
-      // support ControllerGetVolumeHealth.
+      // A plugin which supports the LIST_VOLUME_HEALTH capability
+      // MUST also support the GET_VOLUME_HEALTH capability.
       LIST_VOLUME_HEALTH = 17 [(alpha_enum_value) = true];
     }
 
--- a/csi.proto
+++ b/csi.proto
@@ -1057,7 +1057,8 @@ enum VolumeHealthErrorType {
   // being accessible from that node.
   INACCESSIBLE = 2;
 
-  // Data loss is known or strongly suspected on underlying volume.
+  // Permanent data loss is known or strongly suspected on underlying 
+  // volume.
   DATA_LOSS = 3;
 }
 
@@ -1097,7 +1098,7 @@ message VolumeHealth {
   // ones, but MUST NOT remove an older error entry from
   // health_statuses until that condition is no longer present.
   //
-  // COs MAY ignore unknown VolumeHealthErrorType values, that they 
+  // COs MUST ignore unknown VolumeHealthErrorType values, that they 
   // don't know about.
   //
   // For example:
@@ -1325,17 +1326,17 @@ message ControllerServiceCapability {
       // This enables COs to fetch an existing snapshot.
       GET_SNAPSHOT = 15 [(alpha_enum_value) = true];
 
+      // Indicates the SP supports the ControllerGetVolumeHealth RPC.
+      // This enables COs to fetch health information for a single
+      // volume from the Controller Plugin's perspective.
+      GET_VOLUME_HEALTH = 16 [(alpha_enum_value) = true];
+	  
       // Indicates the SP supports the ControllerListVolumeHealth RPC.
       // This enables COs to fetch volume health information from
       // the Controller Plugin's perspective.
       // A SP that supports ControllerListVolumeHealth MUST also
       // support ControllerGetVolumeHealth.
-      LIST_VOLUME_HEALTH = 16 [(alpha_enum_value) = true];
-
-      // Indicates the SP supports the ControllerGetVolumeHealth RPC.
-      // This enables COs to fetch health information for a single
-      // volume from the Controller Plugin's perspective.
-      GET_VOLUME_HEALTH = 17 [(alpha_enum_value) = true];
+      LIST_VOLUME_HEALTH = 17 [(alpha_enum_value) = true];
     }
 
     Type type = 1;
@@ -1759,8 +1760,8 @@ message NodeGetVolumeHealthRequest {
   // This field is REQUIRED.
   string volume_id = 1;
 
-  // The path where the volume is or was expected to be 
-  // published on the node. If not empty, it MUST be an absolute path
+  // The path where the volume was published or CO attempted to 
+  // publish on the node. If not empty, it MUST be an absolute path
   // in the root filesystem of the process serving this request.
   // This field is OPTIONAL.
   // This field overrides the general CSI size limit.
@@ -1769,7 +1770,8 @@ message NodeGetVolumeHealthRequest {
   // length of at least 128 bytes.
   string volume_publish_path = 2;
   
-  // The path where the volume is or was staged, if the plugin has the
+  // The path where the volume was staged or CO attempted to
+  // stage the volume, if the plugin has the
   // STAGE_UNSTAGE_VOLUME capability, otherwise empty.
   // If not empty, it MUST be an absolute path in the root
   // filesystem of the process serving this request.
@@ -1830,14 +1832,14 @@ message NodeGetStorageHealthResponse {
 enum StorageHealthErrorType {
   UNKNOWN_STORAGE_HEALTH_ERROR_TYPE = 0;
 
-  // The storage backend is completely unreachable from this node.
-  // Volumes using this backend are expected to be unavailable.
-  STORAGE_UNREACHABLE = 1;
-
   // The storage backend is operating in a degraded
   // state (e.g. reduced path count, high latency).
   // Volumes using this backend may experience reduced performance.
-  STORAGE_DEGRADED = 2;
+  STORAGE_DEGRADED = 1;
+  
+  // The storage backend is completely unreachable from this node.
+  // Volumes using this backend are expected to be unavailable.
+  STORAGE_UNREACHABLE = 2;
 }
 message NodeGetCapabilitiesRequest {
   // Intentionally empty.
--- a/csi.proto
+++ b/csi.proto
@@ -977,6 +977,7 @@ message ListVolumesResponse {
     // reported by the SP. The CO MUST be resilient to that.
     repeated string published_node_ids = 1;
 
+    // Field 2 was VolumeCondition, an alpha API that has been removed.
     reserved 2;
   }
 
@@ -1158,6 +1159,7 @@ message ControllerGetVolumeResponse {
     // reported by the SP. The CO MUST be resilient to that.
     repeated string published_node_ids = 1;
 
+    // Field 2 was VolumeCondition, an alpha API that has been removed.
     reserved 2;
   }
 
@@ -1303,6 +1305,8 @@ message ControllerServiceCapability {
       // The SP MUST also support PUBLISH_UNPUBLISH_VOLUME.
       LIST_VOLUMES_PUBLISHED_NODES = 10;
 
+      // Value 11 was VOLUME_CONDITION, an alpha API that has been
+      // removed.
       reserved 11;
 
       // Indicates the SP supports the ControllerGetVolume RPC.
@@ -1730,6 +1734,7 @@ message NodeGetVolumeStatsResponse {
   // This field is OPTIONAL.
   repeated VolumeUsage usage = 1;
 
+  // Field 2 was VolumeCondition, an alpha API that has been removed.
   reserved 2;
 }
 
@@ -1866,6 +1871,8 @@ message NodeServiceCapability {
       // See VolumeExpansion for details.
       EXPAND_VOLUME = 3;
 
+      // Value 4 was VOLUME_CONDITION, an alpha API that has been
+      // removed.
       reserved 4;
 
       // Indicates the SP supports the SINGLE_NODE_SINGLE_WRITER and/or
--- a/csi.proto
+++ b/csi.proto
@@ -79,6 +79,16 @@ service Controller {
   rpc ListVolumes (ListVolumesRequest)
     returns (ListVolumesResponse) {}
 
+  rpc ControllerListVolumeHealth (ControllerListVolumeHealthRequest)
+    returns (ControllerListVolumeHealthResponse) {
+        option (alpha_method) = true;
+    }
+
+  rpc ControllerGetVolumeHealth (ControllerGetVolumeHealthRequest)
+    returns (ControllerGetVolumeHealthResponse) {
+    	option (alpha_method) = true;
+    }
+
   rpc GetCapacity (GetCapacityRequest)
     returns (GetCapacityResponse) {}
 
@@ -157,6 +167,15 @@ service Node {
   rpc NodeGetVolumeStats (NodeGetVolumeStatsRequest)
     returns (NodeGetVolumeStatsResponse) {}
 
+  rpc NodeGetVolumeHealth (NodeGetVolumeHealthRequest)
+    returns (NodeGetVolumeHealthResponse) {
+        option (alpha_method) = true;
+    }
+
+  rpc NodeGetStorageHealth (NodeGetStorageHealthRequest)
+    returns (NodeGetStorageHealthResponse) {
+        option (alpha_method) = true;
+    }
 
   rpc NodeExpandVolume(NodeExpandVolumeRequest)
     returns (NodeExpandVolumeResponse) {}
@@ -958,11 +977,7 @@ message ListVolumesResponse {
     // reported by the SP. The CO MUST be resilient to that.
     repeated string published_node_ids = 1;
 
-    // Information about the current condition of the volume.
-    // This field is OPTIONAL.
-    // This field MUST be specified if the
-    // VOLUME_CONDITION controller capability is supported.
-    VolumeCondition volume_condition = 2 [(alpha_field) = true];
+    reserved 2;
   }
 
   message Entry {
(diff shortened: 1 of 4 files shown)


spec.md
```diff
@@ -1557,6 +1571,8 @@ message ListVolumesRequest {
 
 message ListVolumesResponse {
   message VolumeStatus{
+    reserved 2;
+
```
nixpanic: I wondered if removing the `volume_condition` field (and others) causes compatibility issues. It seems that using the `reserved` notation addresses my concerns.

👍

spec.md
```diff
@@ -2717,6 +2840,118 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // Secrets required by plugin to complete the request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 1 [(csi_secret) = true];
+}
+
+message NodeGetStorageHealthResponse {
+  option (alpha_message) = true;
+
+  message StorageBackendHealth {
```
nixpanic, line 3019: This is actually something that could be provided by the controller too. Network storage clusters (like Ceph) can return their health with sufficient parameters and secrets. If the backend storage is unhealthy, there probably is little reason to attempt to use it for new volumes, or expect (fully) functional usage by the nodeplugin.

gnufied: I was thinking about this use case, but where would we report it? In `CSIDriver` object? I do not think, CSI driver object was designed for such use case, but idea is sound in theory.

I am intentionally keeping some of the health reporting as future enhancement, so as we can ship something rather than being bogged down by designing everything upfront. I am adding a section in https://github.com/kubernetes/enhancements/pull/6138 to keep track of future enhancements. I will update

nixpanic:
>I was thinking about this use case, but where would we report it? In `CSIDriver` object? I do not think, CSI driver object was designed for such use case, but idea is sound in theory.
>
>I am intentionally keeping some of the health reporting as future enhancement, so as we can ship something rather than being bogged down by designing everything upfront. I am adding a section in https://github.com/kubernetes/enhancements/pull/6138 to keep track of future enhancements. I will update 

Thanks for the consideration. There is no need to include such a feature with this PR, but it would be nice if we can add it at a later time.

The (currently) only location where this makes sense to report, is at the StorageClass level. Depending on the parameters/secrets, a part of a storage backend could be unhealthy (like a pool of disks, HDD vs NVMe, or network connection type, or the like).

nixpanic (approved):
The current proposal looks good to me, and it should be workable with the CSI-drivers for Ceph.

spec.md
gnufied: So I have restored this field for now and marked it as deprecated. It turned out, removing the field did break CSI drivers which were setting this field.

nixpanic: I am not sure I share this sentiment. CSI drivers that opted in to support an alpha feature, can not assume that the alpha feature stays (the same) in a next spec update. When a vendor rebases their CSI driver to the next spec update, they may need to adjust the implementation for alpha features. To me, that is an understandable _risk_ when adoption alpha features.

Deprecation is very friendly to do, but I would not see it as a required step before dropping an alpha feature.

nixpanic (approved):

spec.md
```diff
@@ -1608,6 +1627,205 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+If the plugin is unable to complete the ControllerListVolumeHealth call
+successfully, it MUST return a non-ok gRPC code in the gRPC status.
+If the conditions defined below are encountered, the plugin MUST return
+the specified gRPC error code. The CO MUST implement the specified error
+recovery behavior when it encounters the gRPC error code.
+
+| Condition | gRPC Code | Description | Recovery Behavior |
+|-----------|-----------|-------------|-------------------|
+| Invalid `starting_token` | 10 ABORTED | Indicates that `starting_token` is not valid. | Caller SHOULD start the `ControllerListVolumeHealth` operation again with an empty `starting_token`. |
+| Volume does not exist | 5 NOT_FOUND | Indicates that a volume corresponding to the specified `volume_id` does not exist. | Caller MUST verify that the `volume_id` is correct and that the volume is accessible and has not been deleted before retrying with exponential back off. |
```
xing-yang: This makes sense for ControllerGetVolumeHealth.  For ControllerListVolumeHealth, however, volume_id is not an input so we should not get this error code.

spec.md
```diff
@@ -2773,6 +3122,16 @@ message NodeServiceCapability {
...
+
+      // Indicates the SP supports the NodeGetVolumeHealth RPC.
+      // This enables COs to fetch per-volume health information from
+      // the Node Plugin's perspective.
+      GET_VOLUME_HEALTH = 7 [(alpha_enum_value) = true];
+
+      // Indicates the SP supports the NodeGetStorageHealth RPC.
+      // This enables COs to fetch node-local storage health
+      // information that can affect one or more volumes.
+      STORAGE_HEALTH = 8 [(alpha_enum_value) = true];
```
xing-yang: Should we call this GET_STORAGE_HEALTH to be consistent with GET_VOLUME_HEALTH?

spec.md
gnufied: fixed.

spec.md
gnufied: Okay, lets discuss this in today's CSI meeting and take a call about this.

spec.md
```diff
@@ -1608,6 +1627,204 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // When a volume is reported as INACCESSIBLE in NodeGetVolumeHealth 
+  // RPC call the CO MAY interpret the result as volume not 
+  // being accessible from that node.
+  INACCESSIBLE = 1;
+
+  // Data loss is known or strongly suspected on underlying volume.
+  DATA_LOSS = 2;
+
+  // The volume is usable but is not operating optimally.
+  DEGRADED = 3;
```
gnufied: Move this to first one. from @bswartz

spec.md
```diff
@@ -2717,6 +2935,144 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  // length of at least 128 bytes.
+  string staging_target_path = 3;
+}
+
+message NodeGetVolumeHealthResponse {
+  option (alpha_message) = true;
+
+  // The health information for the requested volume.
+  // This field is REQUIRED.
+  VolumeHealth volume_health = 1;
```
gnufied, line 2978: This should be an array too

spec.md
```diff
@@ -2717,6 +2935,144 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+  UNKNOWN_STORAGE_HEALTH_ERROR_TYPE = 0;
+
+  // The storage backend is completely unreachable from this node.
+  // Volumes using this backend are expected to be unavailable.
+  STORAGE_UNREACHABLE = 1;
+
+  // The storage backend is operating in a degraded
+  // state (e.g. reduced path count, high latency).
+  // Volumes using this backend may experience reduced performance.
+  STORAGE_DEGRADED = 2;
```
gnufied: Can we reuse the enum we defined for `VolumeHealth` here ?

spec.md
```diff
@@ -2717,6 +2935,144 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+node capability.
+
+```protobuf
+message NodeGetStorageHealthRequest {
+  option (alpha_message) = true;
+
+  // Secrets required by the plugin to complete the request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 1 [(csi_secret) = true];
```
gnufied, line 3013: Where will this be coming from? Where will this be stored in k8s?

spec.md
```diff
@@ -1608,6 +1627,204 @@ The CO MUST implement the specified error recovery behavior when it encounters t
...
+
+The Plugin SHOULD omit volumes with no known adverse health condition when
+listing all volume health. 
+
+If volumes are created, deleted, or change health while the CO is
+concurrently paging through `ControllerListVolumeHealth` results then it
+is possible that the CO MAY either witness duplicate volumes in the
+list, not witness existing volumes, or both. The CO SHALL NOT expect a
+consistent "view" of all volume health when paging through the volume
+health list via multiple calls to `ControllerListVolumeHealth`.
```
gnufied, line 1651: Can we have that a driver must implement `Get` call if it has `List` call.

spec.md
gnufied: Wait, sorry I was wrong. `VolumeHealth` itself has repeated `VolumeHealthEntry` and hence this field does not need to be repeated.

spec.md
gnufied: In general I think the agreement in the CSI community sync was to simply drop the older fields rather than deprecate them. Lets do that.

spec.md
gnufied: @bswartz I have been thinking long and hard about this and IMO keeping these two enum separate makes most sense in the long run. If we chose one enum, we will have to overload the semantics, the expected CO behaviour for node vs single volume could be different.

spec.md
gnufied: We discussed this some more in CSI implementation meeting and we are keeping these secret fields, but in current iteration of k8s implementation, secrets will be unsupported for volume health. The exact details for where the secret for these health RPCs will be coming in k8s will be decided later.

spec.md
gnufied: Updated the spec for this.

spec.md
gnufied: moved.

spec.md
gnufied: removed.

torredil (approved):

csi.proto
```diff
@@ -985,6 +1000,140 @@ message ListVolumesResponse {
...
+  // ControllerListVolumeHealth or ControllerGetVolumeHealth 
+  // RPC calls, the CO MAY interpret the result as 
+  // volume not being accessible from all nodes in the cluster.
+  //
+  // When a volume is reported as INACCESSIBLE in NodeGetVolumeHealth 
+  // RPC call the CO MAY interpret the result as volume not 
+  // being accessible from that node.
+  INACCESSIBLE = 2;
+
+  // Data loss is known or strongly suspected on underlying volume.
```
bswartz: Permanent data loss is known or suspected.....

This is to distinguish from situations where data was lost but self-healing will eventually repair the problem.

We may also want to add text here indicating that we recommend discarding this volume and rebuilding from other (higher level) replicas if possible, because this is supposed to be one of the actionable health states and we don't want there to be any confusion about what its intended use is.

csi.proto
```diff
@@ -985,6 +1000,140 @@ message ListVolumesResponse {
...
+  //
+  // The SP MAY report one or more health issues with the volume at the
+  // same time. A future version of the CSI spec MAY introduce
+  // additional VolumeHealthErrorType values. A Plugin that adopts a
+  // newer spec version MAY report newer error types alongside existing
+  // ones, but MUST NOT remove an older error entry from
+  // health_statuses until that condition is no longer present.
+  //
+  // COs MAY ignore unknown VolumeHealthErrorType values, that they 
+  // don't know about.
```
bswartz: I would say they must ignore? What are you thinking of leaving the door open for?

My fear is that if we leave this door open, and some CO starts to react to health states it doesn't understand, that may prevent us from safely adding new health states where very particular handling is required when we know that olders COs will do something by default which happens to be wrong for that case.

csi.proto
```diff
@@ -1190,6 +1324,18 @@ message ControllerServiceCapability {
...
+      // This enables COs to fetch volume health information from
+      // the Controller Plugin's perspective.
+      // A SP that supports ControllerListVolumeHealth MUST also
+      // support ControllerGetVolumeHealth.
+      LIST_VOLUME_HEALTH = 16 [(alpha_enum_value) = true];
+
+      // Indicates the SP supports the ControllerGetVolumeHealth RPC.
+      // This enables COs to fetch health information for a single
+      // volume from the Controller Plugin's perspective.
+      GET_VOLUME_HEALTH = 17 [(alpha_enum_value) = true];
```
bswartz: I though we agreed one capability would govern both RPCs, so it's only possible to have both or none. Please combine these.

csi.proto
```diff
@@ -1609,18 +1752,92 @@ message VolumeUsage {
...
 
-  // Normal volumes are available for use and operating optimally.
-  // An abnormal volume does not meet these criteria.
+  // The ID of the volume.
   // This field is REQUIRED.
-  bool abnormal = 1;
+  string volume_id = 1;
+
+  // The path where the volume is or was expected to be 
+  // published on the node. If not empty, it MUST be an absolute path
```
bswartz: "was expected to be" is a little too vague. AIUI, the only case when this might be called on an unpublished volume is AFTER a failed call to PublishVolume. If that's the case, we should say so, otherwise SPs will pull their hair out trying to figure out what to do with this path in the case it's called on a volume where a publish was never attempted, and thus the SP never saw the publish context/volume context.

csi.proto
```diff
@@ -1609,18 +1752,92 @@ message VolumeUsage {
...
+  UNKNOWN_STORAGE_HEALTH_ERROR_TYPE = 0;
+
+  // The storage backend is completely unreachable from this node.
+  // Volumes using this backend are expected to be unavailable.
+  STORAGE_UNREACHABLE = 1;
+
+  // The storage backend is operating in a degraded
+  // state (e.g. reduced path count, high latency).
+  // Volumes using this backend may experience reduced performance.
+  STORAGE_DEGRADED = 2;
```
bswartz: Please reorder these so that degraded comes first, assuming it's a similar wildcard to the volume degraded state.

csi.proto
```diff
@@ -1672,6 +1879,16 @@ message NodeServiceCapability {
...
+
+      // Indicates the SP supports the NodeGetVolumeHealth RPC.
+      // This enables COs to fetch per-volume health information from
+      // the Node Plugin's perspective.
+      GET_VOLUME_HEALTH = 7 [(alpha_enum_value) = true];
+
+      // Indicates the SP supports the NodeGetStorageHealth RPC.
+      // This enables COs to fetch health of entire storage subsystem
+	  // from node's perspective. 
+      GET_STORAGE_HEALTH = 8 [(alpha_enum_value) = true];
```
bswartz, line 1902: Why split these? Having 2 capabilities means the CO will need to deal with all 4 SP cases (has neither, has both, has GET_VOLUME_HEALTH only, and has GET_STORAGE_HEALTH only) which is extra code on the CO side. Do we really expect some SPs to implement one or the other? Is it a big burden to make them implement both, or to just return always healthy for the one they don't want to implement?

bswartz (changes requested):

csi.proto
gnufied: I think we kinda did. LIST_VOLUME_HEALTH implies GET_VOLUME_HEALTH, but GET_VOLUME_HEALTH does not imply LIST_VOLUME_HEALTH. 

Do we want to force providers who only can support getting health of individual volumes to also support LIST_VOLUME_HEALTH?

bswartz: Oh that's a fair point. I see the utility in allowing just Get but not List. In that case we should make GET come before LIST so that the dependency points backwards instead of forewards.

csi.proto
gnufied: >Permanent data loss is known or suspected.....
+1

>We may also want to add text here indicating that we recommend discarding this volume and rebuilding from other (higher level) replicas if possible, because this is supposed to be one of the actionable health states and we don't want there to be any confusion about what its intended use is.

So far we have avoided being too prescriptive on remediation (notice inaccessible doesn't say do not schedule workload that uses this volume on this node). I think that returned health condition is basically a suggestion to the CO and the workload (managed via some kind of operator), at the end of day - it is impossible to enforce this one way or the other from CSI perspective, so being very prescriptive IMO does not add much value.

csi.proto
gnufied: >Do we really expect some SPs to implement one or the other.

Yes it could be easily that a lot of cloudproviders based SP will not have `GET_STORAGE_HEALTH` capability but will definitely want `GET_VOLUME_HEALTH` on the node. On the other hand - providers like TopoLVM(https://github.com/topolvm/topolvm) will have both capabilities. 

>Having 2 capabilities means the CO will need to deal with all 4 SP cases (has neither, has both, has GET_VOLUME_HEALTH only, and has GET_STORAGE_HEALTH only) which is extra code on the CO side. 

So these 2 have very different lifecycle from CO perspective. Volume Health of a individual volume is reported in the pod and after scheduling the pod to the node, whereas storage susbsytem health from the node is reported in `CSINode` and regardless of pod scheduling concerns. IMO - it makes sense to keep them separate.

csi.proto
gnufied: updated the wording.

gnufied:
@bswartz addressed the review comments PTAL

csi.proto
bswartz: I agree we want to leave a little wiggle room, but too much wiggle room makes it impossible for SPs to implement the feature. An SP will worry that reporting a health condition could result in unknown COs doing unknown inappropriate things in response to that health condition, and could choose to just never report anything other than degraded for fear of some unintented outcome. If we constrain what the CO can/should do in response to any given condition, it will give the SP authors confidence.

csi.proto
bswartz: I think the normal way we express dependencies is to mention the other capability (since that capability explains what must be implemented. How about:

// Indicates the SP supports the ControllerListVolumeHealth RPC.
// This enables COs to fetch volume health information from
// the Controller Plugin's perspective.
// A plugin which supports the LIST_VOLUME_HEALTH capability
// MUST also support the GET_VOLUME_HEALTH capability.

csi.proto
bswartz: Okay I buy this

csi.proto
gnufied: Updated.

csi.proto
gnufied: Okay I have updated the wording to following:

>  // Permanent data loss is known or strongly suspected on the
  // underlying volume. The CO MAY use this signal to inform
  // users to trigger recovery workflows such as restoring from
  // a snapshot or replica.

bswartz (approved):
LGTM Thanks for addressing all my comments @gnufied

xing-yang:
lgtm

spec.md
```diff
@@ -1643,11 +1863,7 @@ message ControllerGetVolumeResponse {
     // reported by the SP. The CO MUST be resilient to that.
     repeated string published_node_ids = 1;
 
-    // Information about the current condition of the volume.
-    // This field is OPTIONAL.
-    // This field MUST be specified if the
-    // VOLUME_CONDITION controller capability is supported.
-    VolumeCondition volume_condition = 2;
+    reserved 2;
```
xing-yang, line 1868: Nit: Can you add a note on why this is reserved

spec.md
```diff
@@ -1841,22 +2057,11 @@ message ControllerServiceCapability {
...
-      // Plugin, only the Node Plugin, or both.
-      // If `VolumeCondition` is implemented in both the Controller and
-      // Node Plugins, it SHALL report from different perspectives.
-      // If for some reason Controller and Node Plugins report
-      // misaligned volume conditions, CO SHALL assume the worst case
-      // is the truth.
-      // Note that, for alpha, `VolumeCondition` is intended be
-      // informative for humans only, not for automation.
-      VOLUME_CONDITION = 11 [(alpha_enum_value) = true];
+      reserved 11;
```
xing-yang, line 2064: Same here

spec.md
```diff
@@ -2662,11 +2879,8 @@ message NodeGetVolumeStatsRequest {
 message NodeGetVolumeStatsResponse {
   // This field is OPTIONAL.
   repeated VolumeUsage usage = 1;
-  // Information about the current condition of the volume.
-  // This field is OPTIONAL.
-  // This field MUST be specified if the VOLUME_CONDITION node
-  // capability is supported.
-  VolumeCondition volume_condition = 2 [(alpha_field) = true];
+
+  reserved 2;
```
xing-yang, line 2888: Same

spec.md
```diff
@@ -1569,11 +1588,7 @@ message ListVolumesResponse {
     // reported by the SP. The CO MUST be resilient to that.
     repeated string published_node_ids = 1;
 
-    // Information about the current condition of the volume.
-    // This field is OPTIONAL.
-    // This field MUST be specified if the
-    // VOLUME_CONDITION controller capability is supported.
-    VolumeCondition volume_condition = 2 [(alpha_field) = true];
+    reserved 2;
```
xing-yang, line 1592: Nit: Add a note why it is reserved.

gnufied: fixed. and other occurences too.

saad-ali:
/lgtm
/approve

Merging to catch next release.

PR #605, opened 2026-06-08
Fix RESOURCE_EXHAUSTED error code from 13 to 8

nixpanic:
**What type of PR is this?**

/bugfix

Inconsistent/incorrect error-code can cause confusion on what is intended to happen.

This should not cause problems for CSI-driver implementers. The `RESOURCE_EXHAUSTED` error description is clear enough. Code 13 would be `INTERNAL`, which does not make much sense to return when the description is read.

```release-note
The numeric error-code for CreateSnapshot and CreateGroupSnapshot is corrected for the RESOURCE_EXHAUSTED value from 13 (INTERNAL) to 8.
```


nixpanic:
Not sure why the build test fails... The protobuf file should not have been modified by the change in the text, there is no reason to rebuild anything.

gnufied:
I think the build is failing because - you need to checkout the `spec` repo in `GOPATH` and run `make` command. I noticed if repo is not checked out in `GOPATH` it can generate different `pb.go` files.

nixpanic:
> I think the build is failing because - you need to checkout the `spec` repo in `GOPATH` and run `make` command. I noticed if repo is not checked out in `GOPATH` it can generate different `pb.go` files.

Thanks for the hint. I don't think the changes affect the `csi.proto` file, so didn't expect to need to update the generated protobuf stubs either. Oh well :man_shrugging:

PR #606, opened 2026-06-11
KEP-5943: Introducing Snapshot Topology

mdzraf:
**What type of PR is this?**

Introducing Snapshot Topology to CSI Spec.

This PR is part of the changes necessary for https://github.com/kubernetes/enhancements/pull/5944 

```release-note
NONE
```

### DIFF
--- a/csi.proto
+++ b/csi.proto
@@ -256,6 +256,17 @@ message PluginCapability {
       // The presence of this capability determines whether the CO will
       // attempt to invoke the OPTIONAL SnapshotMetadata service RPCs.
       SNAPSHOT_METADATA_SERVICE = 4 [(alpha_enum_value) = true];
+
+      // SNAPSHOT_ACCESSIBILITY_CONSTRAINTS indicates that the snapshots
+      // for this plugin MAY NOT be equally usable from all
+      // topologies in the cluster. A snapshot is usable from a location
+      // if volumes created from that snapshot are guaranteed to be
+      // accessible from that location. The CO MUST use the topology
+      // information returned in the CreateSnapshotResponse to ensure
+      // that a desired volume can be provisioned from a given snapshot
+      // when scheduling workloads.
+      SNAPSHOT_ACCESSIBILITY_CONSTRAINTS = 5
+        [(alpha_enum_value) = true];
     }
     Type type = 1;
   }
@@ -1382,6 +1393,22 @@ message CreateSnapshotRequest {
   // - Specify primary or secondary for replication systems that
   //   support snapshotting only on primary.
   map<string, string> parameters = 4;
+
+  // Specifies where (regions, zones, racks, etc.) the provisioned
+  // snapshot MUST be usable from. A snapshot is usable from a location
+  // if volumes created from that snapshot are guaranteed to be
+  // accessible from that location.
+  // An SP SHALL advertise the requirements for topological
+  // accessibility information in documentation. COs SHALL only specify
+  // topological accessibility information supported by the SP.
+  // This field is OPTIONAL.
+  // This field SHALL NOT be specified unless the SP has the
+  // SNAPSHOT_ACCESSIBILITY_CONSTRAINTS plugin capability.
+  // If this field is not specified and the SP has the
+  // SNAPSHOT_ACCESSIBILITY_CONSTRAINTS plugin capability, the SP MAY
+  // choose where the provisioned snapshot is usable from.
+  TopologyRequirement accessibility_requirements = 5
+    [(alpha_field) = true];
 }
 
 message CreateSnapshotResponse {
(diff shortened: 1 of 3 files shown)


xing-yang:
Do we need to add some constraints for TopologyRequirement in CreateVolumeRequest if VolumeContentSource is a Snapshot?

csi.proto
```diff
@@ -1229,6 +1237,19 @@ message CreateSnapshotRequest {
...
+  // An SP SHALL advertise the requirements for topological
+  // accessibility information in documentation. COs SHALL only specify
+  // topological accessibility information supported by the SP.
+  // This field is OPTIONAL.
+  // This field SHALL NOT be specified unless the SP has the
+  // SNAPSHOT_ACCESSIBILITY_CONSTRAINTS plugin capability.
+  // If this field is not specified and the SP has the
+  // SNAPSHOT_ACCESSIBILITY_CONSTRAINTS plugin capability, the SP MAY
+  // choose where the provisioned snapshot is accessible from.
+  TopologyRequirement accessibility_requirements = 5 [(alpha_field) = true];
```
gnufied: Why can't this be determined by SP from volume?

To be clear - I accept that we probably need the topology in `CreateSnapshot` response, but I am trying to think why we need this in the request.

mdzraf: This was one of the approaches that was considered but decided against because it would be good to support SPs that allow for creating Snapshots with topologies that differ from the source volume amongst other reasons. There is a more detailed explanation in the [Alternatives Section of the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/5943-topology-for-volume-snapshot/README.md#inherit-topology-from-the-source-persistentvolume)

mdzraf:
> Do we need to add some constraints for TopologyRequirement in CreateVolumeRequest if VolumeContentSource is a Snapshot?

We can discuss this but we will already be enforcing it at provisioning time via the external-provisioner and the custom scheduler plugin.

spec.md
```diff
@@ -2037,7 +2070,7 @@ The CO MUST implement the specified error recovery behavior when it encounters t
 |-----------|-----------|-------------|-------------------|
 | Snapshot already exists but is incompatible | 6 ALREADY_EXISTS | Indicates that a snapshot corresponding to the specified snapshot `name` already exists but is incompatible with the specified `volume_id`. | Caller MUST fix the arguments or use a different `name` before retrying. |
```
mdzraf, line 2259: 13 Is not the right number for `RESOURCE_EXHAUSTED` this is fixed in #605, hence why the number change here as well instead of a separate entry.

mdzraf:
cc: @bswartz & @saad-ali for reviews/approvals as discussed in CSI Community Sync Meeting.

spec.md
bswartz: A reference to `accessibility_requirements` is needed here, as in `CreateVolume`

csi.proto
```diff
@@ -1229,6 +1237,19 @@ message CreateSnapshotRequest {
   // - Specify primary or secondary for replication systems that
   //   support snapshotting only on primary.
   map<string, string> parameters = 4;
+
+  // Specifies where (regions, zones, racks, etc.) the provisioned
+  // snapshot MUST be accessible from.
```
bswartz: I'm uncomfortable with the word "accessible" here. The word has a specific meaning for volumes which doesn't apply to snapshots. What we are trying to communicate is not accessibility but whether the snapshot can be used to create volumes in a given topology. We don't really have a word that means that anywhere in the spec. I could imagine "located" or "usable from". We could use the awkward phrasing:

Specifies where (regions, zones, racks, etc.) volumes created from the provisioned snapshot MUST be accessible from.

This communicates the actual thing that CSI cares about, which is whether you can create a volume that is accessible in the topology in question, but I'm not sure it's exactly what we want to say.

If we introduced a new word:

Specifies where (regions, zones, racks, etc.) the provisioned snapshot MUST be usable from. A snapshot is usable from a location if volumes created from that snapshot are guaranteed to be accessible.

This is a little bit more clear. What do you think?

bswartz (changes requested):

csi.proto
mdzraf: I like that wording better as well. Addressed in new rev.

spec.md
mdzraf: Addressed in new rev, thank you!

bswartz (approved):
LGTM

xing-yang:
lgtm

saad-ali:
@mdzraf can you please 

1. Make sure you've signed the CSI CLA - https://github.com/container-storage-interface/spec/blob/master/CONTRIBUTING.md#licence-and-cla
2. Please rebase your changes.

mdzraf:
Hi @saad-ali , I'ver signed the CLA and sent the email to the CSI Approvers list and CC'd you on it, and PR is rebased. 

Thank you!

saad-ali:
Confirmed CLA, thanks!

Can you check why the build might be failing here?

mdzraf:
Hi @saad-ali , I've fixed it now, should pass.

saad-ali:
Thank you.

/lgtm
/approve

